PR #8, opened 2024-07-30
Provide two variant of ERC4626Fees with different models

Amxx:
@cairoeth lets do the `maxXxxx` changes here.


ernestognw (dismissed):
I like the way this is exposed, it's clear both options are different and the description is good too. LGTM

PR #19, opened 2024-11-07
Fix `prepare` when using as git repository dependency

cairoeth:
Implements https://github.com/OpenZeppelin/openzeppelin-contracts/commit/4764ea50750d8bda9096e833706beba86918b163


cairoeth:
merging given that it blocks wizard PR and it was already validated in the main repo

PR #25, opened 2024-11-26
Add ERC-4337 and ERC-7702 account implementations

ernestognw:


Amxx (commented):
There is a lot of duplicated code between AccountECDSA, AccountP256 and AccountRSA:
- ERC7739Signer
- ERC721Holder
- ERC1155HolderLean

maybe we should bundle all that in some way that all variants can get easily.

contracts/account/draft-AccountECDSA.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * @dev Returns the ERC-191 signed `userOpHash` hashed with keccak256 using `personal_sign`.
+     */
+    function _userOpSignedHash(
+        PackedUserOperation calldata /* userOp */,
+        bytes32 userOpHash
+    ) internal view virtual override returns (bytes32) {
+        return userOpHash.toEthSignedMessageHash();
+    }
```
Amxx, line 59: Any reason to not use EIP-712?

contracts/utils/cryptography/draft-ERC7739Signer.sol
```diff
@@ -108,9 +115,16 @@ abstract contract ERC7739Signer is EIP712, IERC1271 {
...
+     * To ensure there's no way to inherit from this ERC-7739 signer and still be able to sign raw messages,
+     * this function provides a nested EIP-712 hash. Developers must implement only this function to ensure
+     * no raw message signing is possible.
+     *
      * WARNING: Implementing a signature validation algorithm is a security-sensitive operation as it involves
      * cryptographic verification. It is important to review and test thoroughly before deployment. Consider
      * using one of the signature verification libraries ({ECDSA}, {P256} or {RSA}).
      */
-    function _validateSignature(bytes32 hash, bytes calldata signature) internal view virtual returns (bool);
+    function _validateNestedEIP712Signature(
```
Amxx, line 126: What is the motivation for this change of name. I personnally prefer the simplicity and clarity of `_validateSignature`, and I don't see why we need to expose details like "NestedEIP712" in the naming.

contracts/utils/cryptography/draft-ERC7739Signer.sol
```diff
@@ -55,21 +55,28 @@ abstract contract ERC7739Signer is EIP712, IERC1271 {
...
      * @dev Nested EIP-712 typed data verification.
+     *
+     * NOTE: Instead of overriding this function, try with {_validateNestedEIP712Signature}. It encapsulates
+     * nested EIP-712 hashes.
      */
     function _isValidNestedTypedDataSignature(
         bytes32 hash,
         bytes calldata encodedSignature
-    ) internal view virtual returns (bool) {
+    ) internal view returns (bool) {
```
Amxx, line 79: Our policity is to make every internal function virtual, unless we have a very strong reason against it. is there any serious issues/danger in having this function virtual ?

contracts/token/ERC1155/utils/ERC1155HolderLean.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+// OpenZeppelin Contracts (last updated v5.0.0) (token/ERC1155/utils/ERC1155Holder.sol)
+
+pragma solidity ^0.8.20;
+
+import {IERC1155Receiver} from "@openzeppelin/contracts/interfaces/IERC1155Receiver.sol";
+
+/**
+ * @dev Version of {ERC1155Holder} that doesn't include {IERC165} detection.
```
Amxx, line 9: Do we want that? It goes against [the ERC-1155 specifications](https://eips.ethereum.org/EIPS/eip-1155#erc-1155-token-receiver)

> Smart contracts MUST implement the ERC-165 `supportsInterface` function and signify support for the `ERC1155TokenReceiver` interface to accept transfers. See “ERC1155TokenReceiver ERC-165 rules” for further detail.

contracts/mocks/docs/account/MyFactoryAccountECDSA.sol
```diff
@@ -0,0 +1,37 @@
...
+    address private immutable _impl = address(new MyAccountECDSA());
+
+    /// @dev Predict the address of the account
+    function predictAddress(bytes32 salt) public view returns (address) {
+        return _impl.predictDeterministicAddress(salt, address(this));
+    }
+
+    /// @dev Create clone accounts on demand
+    function cloneAndInitialize(bytes32 salt, address signer) public returns (address) {
+        return _cloneAndInitialize(salt, signer);
```
Amxx, line 25: Should signer be hashed into the hash. Right now this function can be front-run with the same hash and a different signer.

contracts/mocks/docs/account/MyAccountECDSA.sol
```diff
@@ -0,0 +1,22 @@
...
+    /**
+     * NOTE: EIP-712 domain is set at construction because each account clone
+     * will recalculate its domain separator based on their own address.
+     */
+    constructor() EIP712("MyAccountECDSA", "1") {
+        _disableInitializers();
+    }
+
+    function initializeSigner(address signerAddr) public virtual initializer {
+        _initializeSigner(signerAddr);
```
Amxx, line 20: Not sure about other usecases that justify this being an `_initializeSigner`. But looking at these 3 contracts (MyAccountECDSA, MyAccountP256, MyAccountRSA), it feels like it should be a constructor.

contracts/mocks/docs/account/MyAccountCustom.sol
```diff
@@ -0,0 +1,44 @@
...
+
+pragma solidity ^0.8.20;
+
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
+import {ERC4337Utils, PackedUserOperation} from "@openzeppelin/contracts/account/utils/draft-ERC4337Utils.sol";
+import {Initializable} from "@openzeppelin/contracts/proxy/utils/Initializable.sol";
+import {AccountBase} from "../../../account/draft-AccountBase.sol";
+import {ERC7739Signer} from "../../../utils/cryptography/draft-ERC7739Signer.sol";
+
+contract MyAccountCustom is ERC7739Signer, AccountBase, Initializable {
```
Amxx, line 12: This mixes vanilla contracts (EIP712) with an upgradeable design (the initializer).

I'm wondering what will happen when we get an upgradeable version of the community repo. Like `AccountBaseUpgradeable`?

contracts/mocks/account/AccountRSAMock.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {PackedUserOperation} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
+import {ERC4337Utils} from "@openzeppelin/contracts/account/utils/draft-ERC4337Utils.sol";
```
Amxx, line 6:
```suggestion
```

contracts/mocks/account/AccountECDSAMock.sol
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {PackedUserOperation} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
+import {ERC4337Utils} from "@openzeppelin/contracts/account/utils/draft-ERC4337Utils.sol";
```
Amxx, line 6:
```suggestion
```

contracts/mocks/ERC1155Mock.sol
Amxx, line 1: Ideally, we would just import what we need, and the exposed plugin would take care of it.

Amxx: I was thinking we could whitelist some contracts we want to expose directly in the hardhat-exposed configuration. Something like
```solidity
  exposed: {
    include: [
      'contracts/**/*',
      '@openzeppelin/contracts/token/ERC721/ERC721.sol',
      '@openzeppelin/contracts/token/ERC1155/ERC1155.sol'
    ],
  },
```

but the plugin don't support that unless we change the `config.path.sources` to include "@openzeppelin/contracts/".

@frangio do you have any idea what we could do ?

test/utils/cryptography/draft-ERC7739Signer.test.js
Amxx, line 1: 7739 is only about the "wrapping" logic, and doesn't actually include anything that is specific the the underlying key/curve (that is abstracted in _verifySignature). IMO testing P256/RSA here is just making the tests more complexe and slower to run for no real advantage.

test/helpers/signers.js
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  get EOA() {
+    return new ethers.Wallet(ethers.hexlify(this._privateKey));
+  }
+
+  _ensureLowerOrderS({ s, recovery, ...rest }) {
+    return ensureLowerOrderS(this.N, { s, recovery, ...rest });
+  }
+}
```
Amxx, line 52: As I said in the past, I think having this bastraction for ECDSA is a mistake. `ethers.Wallet` (or equivalent) should be all that we need. Having both coexist may just create confusion.

Additionally, there is a risk that we mess up somewhere in our implementation. It happened in the past !

I'd be favorable to implementing ethers.Signer compatible objects that supports signature using P256 and RSA ... But I would not touch ECDSA for which we basically have everything we need already

contracts/mocks/ERC1155Mock.sol
frangio: The simplest reason why the plugin doesn't support this is that it creates `contracts-exposed/**/*` files for matching `contracts/**/*` files.

What would be the location of the exposed files created for something under `@openzeppelin/`?

I'm not opposed to something like this but it requires some design around this.

Amxx: I'd do
- `contracts-exposed/contracts/**/*`
- `contracts-exposed/@openzeppelin/`

Amxx: feels to me like this PR would mostly take care of that https://github.com/frangio/hardhat-exposed/pull/32

contracts/utils/cryptography/draft-ERC7739Signer.sol
ernestognw: You're right, I'm reverting this change. I intended to convey that the hash passed into this function is already "safe" to clarify why the AccountBase has an `_userOpSignedHash` function.

We can rediscuss that afterwards

contracts/utils/cryptography/draft-ERC7739Signer.sol
ernestognw: There's no issue/danger, and I wanted to experiment with the idea of isolating the implementation, I think those two functions shouldn't customized and instead developers should either override `_validateSignature` or `_isValidSignature`.

What do you think of making both `_isValidNestedTypedDataSignature` and `_isValidNestedPersonalSignSignature` private instead?

I'm reverting for now

test/utils/cryptography/draft-ERC7739Signer.test.js
ernestognw: Yeah, this is right, especially now that we also use the `shouldBehaveLikeERC7739Signer` in each account tests. I wouldn't mind removing RSA and P256 but I don't think it hurts having those cases after your refactor

contracts/mocks/ERC721Mock.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {ERC721} from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
+
+contract ERC721Mock is ERC721 {
+    constructor(string memory _name, string memory _symbol) ERC721(_name, _symbol) {}
+
+    function $mint(address owner, uint256 tokenId) external virtual {
```
frangio, line 10: Is this necessary or is `$_mint` already available?

contracts/utils/cryptography/draft-ERC7739Signer.sol
Amxx: making them private is prossibly the right move.

ernestognw:
@frangio and I consider that the Base Account should not inherit from ERC7339Signer since ERC-7339 is a workaround with important limitations.

frangio:
>important limitations

I was thinking of composability. This made me think it's not good enough to be a default.

On the other hand, 1271 without 7739 is unsafe... possibly with bigger impact than the lack of composability.

So let's discuss it a bit more.

ernestognw:
I updated moving the EIP 712 user op typed hash logic to AccountBase.

One thing is that AccountBase will need EIP-712 to sign the typed user op hash, anyway. In that case, I'm reconsidering having ERC-7739 by default.

> On the other hand, 1271 without 7739 is unsafe... possibly with bigger impact than the lack of composability.

Right. I agree we should be opinionated on the ERC-1271 implementation. iiuc, implementing [ERC-7803](https://eips.ethereum.org/EIPS/eip-7803) would allow for better composability and works better as a default, is that correct?

frangio:
ERC-7803 solves composability but it's not usable today, it needs to be accepted and deployed across wallets, there is no guarantee that will happen at this point.

Amxx (approved):
IMO this is good to merge.

There is still work to do, but that can be done in parallele in separate PR, such that the review works remains acceptable (many small PR are easier to process)

Upcomming work (some of which as aready started)
- Documentation
- Testing (particularly EIP7702)
- Extensions (7579, ...)
- Signer constructor/initialization
- UserOperation -> hash for signature.

ernestognw:
I reviewed the changes and saw they required a couple of documentation (NatSpec) updates so I committed them myself. Basically, I fixed multiple broken links in the documentation (due to #30). I think it's fine to hardcode the OpenZeppelin docs for now so that the correct source is linked.

I'm merging this since there are no controversial changes, so I can focus on ERC7702 testing, documentation, an ERC7579 implementation.

> Signer constructor/initialization

What are your thoughts here?

PR #22, opened 2024-11-20
Implement audit recommendations for crosschain gateway implementations

Amxx:

PR #26, opened 2024-11-28
Simplify remappings and use master for OpenZeppelin dependencies

ernestognw:
### Motivation

Given the recent work in cross-chain and account abstraction, we've found that we would benefit from pointing to OpenZeppelin Contracts at `master`. The reasoning is that we can easily deliver small changes in libraries over there that are later used in this repository.

Otherwise, we would need to wait until an actual release before finishing the implementation in this side. This PR aims to simplify the remappings and remove npm dependencies to depend only in git submodules

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -20,4 +20,5 @@ runs:
       with:
         version: nightly
     - name: Install submodule dependencies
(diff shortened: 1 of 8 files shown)


ernestognw:
I'm merging to speed up #22

PR #32, opened 2024-12-05
Add cross chain audit report

Amxx:


contracts/crosschain/audits/2024-12-04 OpenZeppelin Cross-Chain Messaging Audit.pdf
ernestognw, line 1: I believe the name of the file shouldn't include spaces. I'd rename to `2024-12-04`

ernestognw (approved):

PR #34, opened 2024-12-06
Vendor ERC4337 Contracts

ernestognw:
## Description

Previously, we've agreed to vendor these contracts into OpenZeppelin Contracts. These were removed when the account abstraction work was decided to get migrated to this repository instead.


ernestognw:
All passing, merging since it's required for #25

PR #37, opened 2024-12-09
Add inheritance ordering checks

ernestognw:


ernestognw:
Merging after discussing internally. We think this covers enough compatibility with OpenZeppelin Contracts since it already imports the relevant contracts.

Amxx (approved):
We'll need to improve it to add all of the main repo into the ordering, but that is a good start

Amxx:
Here the relevant are governor. It is currently not imported :/

ernestognw:
Yes. I added this to our 2025 wishlist

PR #44, opened 2024-12-18
Add tests for AccountERC7702Mock

Amxx:


test/helpers/erc4337.js
```diff
@@ -80,6 +88,42 @@ class SmartAccount extends ethers.BaseContract {
...
+      params.paymasterVerificationGasLimit ??= 100_000n;
+      params.paymasterPostOpGasLimit ??= 100_000n;
+      params.paymasterAndData = ethers.solidityPacked(
+        ['address', 'uint128', 'uint128'],
+        [params.paymaster, params.paymasterVerificationGasLimit, params.paymasterPostOpGasLimit],
+      );
+    }
+
+    return new UserOperationWithContext(params);
+  }
```
ernestognw, line 124: I don't like this logic is duplicated with `SmartAccount` but I see it'd be too much to abstract away only this function.

Works for me

ernestognw (approved):

test/helpers/erc4337.js
Amxx: Added a fillUserOp in the helper. Let me know what you think

ernestognw: I like it, thanks!

ernestognw (approved):

PR #45, opened 2024-12-19
Remove entrypoint address for EIP-712 signature of UserOperations

Amxx:


ernestognw (approved):

ernestognw:
For the record, the rationale is that the account already defines their entrypoint.

The signature shouldn't be replayable considering the entrypoint already has a nonce and the sender address is included in the userOpHash.

PR #46, opened 2024-12-20
Add ERC-7579 account extension

Amxx:
Based on https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5250


Amxx:
We should discuss [this](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5250/files#diff-52e99bab868addbfbac354587d8e48634881e5ef6060dc03a2a749968d6fc487R98-R103).

The current behavior of AccountBase does not match what we had in this PR. We should double check what ERC 4337 says regarding user op execution.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    modifier onlyModule(uint256 moduleType) {
+        _checkModule(moduleType, msg.sender);
+        _;
+    }
+
+    /// @dev fallback handler for ERC-7579 fallback modules
+    fallback() external payable virtual {
+        _fallback();
+    }
```
ernestognw, line 29: I'd put this at the end

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+    function _validateUserOp(
+        PackedUserOperation calldata userOp,
+        bytes32 userOpHash
+    ) internal virtual override returns (uint256) {
+        address module = address(bytes20(userOp.signature[0:20]));
+        return
+            isModuleInstalled(MODULE_TYPE_VALIDATOR, module, _emptyCalldataBytes())
+                ? IERC7579Validator(module).validateUserOp(userOp, userOpHash)
+                : super._validateUserOp(userOp, userOpHash);
+    }
```
ernestognw, line 119: AccountCore already passes a "signable" version of `userOpHash` to `_rawSignatureValidation`, which implementation is missing here.

I'd say the right move is to remove this override to `_validateUserOp` and implement `_rawSignatureValidation` by calling a module

```solidity
    bytes32 hash,
    bytes calldata signature
    address module = address(bytes20(signature[0:20]));
    if (!isModuleInstalled(MODULE_TYPE_VALIDATOR, module, msg.data)) return false;
    return IERC7579Validator(module).isValidSignatureWithSender(msg.sender, hash, signature) == bytes4(0x77390001);
}
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+
+    /// @inheritdoc IERC7579AccountConfig
+    function accountId() public view virtual returns (string memory accountImplementationId) {
+        //vendorname.accountname.semver
+        return "@openzeppelin/contracts.erc7579account.v0-beta";
+    }
+
+    /// @inheritdoc IERC7579AccountConfig
+    function supportsExecutionMode(bytes32 encodedMode) public pure returns (bool) {
```
ernestognw, line 38:
```suggestion
    function supportsExecutionMode(bytes32 encodedMode) public view virtual returns (bool) {
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+        (CallType callType, , , ) = Mode.wrap(encodedMode).decodeMode();
+        return
+            callType == ERC7579Utils.CALLTYPE_SINGLE ||
+            callType == ERC7579Utils.CALLTYPE_BATCH ||
+            callType == ERC7579Utils.CALLTYPE_DELEGATECALL;
+    }
+
+    /// @inheritdoc IERC7579AccountConfig
+    // TODO: add hook support
+    function supportsModule(uint256 moduleTypeId) public pure returns (bool) {
```
ernestognw, line 48:
```suggestion
    function supportsModule(uint256 moduleTypeId) public view virtual returns (bool) {
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+    ) public virtual onlyEntryPointOrSelf {
+        _uninstallModule(moduleTypeId, module, deInitData);
+    }
+
+    /// @inheritdoc IERC7579ModuleConfig
+    function isModuleInstalled(
+        uint256 moduleTypeId,
+        address module,
+        bytes calldata additionalContext
+    ) public view returns (bool) {
```
ernestognw, line 78:
```suggestion
    ) public view virtual returns (bool) {
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+        require(supportsModule(moduleTypeId), ERC7579Utils.ERC7579UnsupportedModuleType(moduleTypeId));
+        require(
+            IERC7579Module(module).isModuleType(moduleTypeId),
+            ERC7579Utils.ERC7579MismatchedModuleTypeId(moduleTypeId, module)
+        );
+
+        if (moduleTypeId == MODULE_TYPE_VALIDATOR) {
+            require(_validators.add(module), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
+        } else if (moduleTypeId == MODULE_TYPE_EXECUTOR) {
+            require(_executors.add(module), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
```
ernestognw, line 138: I'd prefer this syntax for the first 2:

```solidity
require(
    moduleTypeId != MODULE_TYPE_VALIDATOR || _validators.add(module),
    ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module)
);
require(
    moduleTypeId != MODULE_TYPE_EXECUTOR || _executors.add(module),
    ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module)
);
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+        IERC7579Module(module).onInstall(initData);
+        emit ModuleInstalled(moduleTypeId, module);
+    }
+
+    function _uninstallModule(uint256 moduleTypeId, address module, bytes memory deInitData) internal virtual {
+        if (moduleTypeId == MODULE_TYPE_VALIDATOR) {
+            require(_validators.remove(module), ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module));
+        } else if (moduleTypeId == MODULE_TYPE_EXECUTOR) {
+            require(_executors.remove(module), ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module));
+        } else if (moduleTypeId == MODULE_TYPE_FALLBACK) {
```
ernestognw, line 157: Same here:

```solidity
require(
    (moduleTypeId != MODULE_TYPE_VALIDATOR || _validators.remove(module)) ||
        (moduleTypeId != MODULE_TYPE_EXECUTOR || _executors.remove(module)),
    ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module)
);
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+
+        IERC7579Module(module).onUninstall(deInitData);
+        emit ModuleUninstalled(moduleTypeId, module);
+    }
+
+    function _installFallback(address module, bytes4 selector) internal virtual returns (bool) {
+        if (_fallbackHandler(selector) == address(0)) {
+            _fallbacks[selector] = module;
+            return true;
+        } else return false;
```
ernestognw, line 171:
```suggestion
       if (_fallbacks[selector] != address(0)) return false;
        _fallbacks[selector] = module;
        return true;
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    function _uninstallFallback(address module, bytes4 selector) internal virtual returns (bool) {
+        if (_fallbackHandler(selector) == module && module != address(0)) {
+            delete _fallbacks[selector];
+            return true;
+        } else return false;
```
ernestognw, line 178:
```suggestion
        address handler = _fallbackHandler(selector);
        if (handler == address(0) || handler != module) return false;
        delete _fallbacks[selector];
        return true;
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    function _fallbackHandler(bytes4 selector) internal view virtual returns (address) {
+        return _fallbacks[selector];
+    }
+
+    function _fallback() internal virtual {
+        address handler = _fallbackHandler(msg.sig);
+        if (handler == address(0)) revert ERC7579MissingFallbackHandler(msg.sig);
```
ernestognw, line 187: Why wouldn't just `return` here? I don't find the specification to be clear about this but I'd say it's better to treat it as a noop and return right away.

ernestognw:
This is ready for review but needs resolution from #54, docs can be added in a different PR

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * This contract does not implement validation logic for user operations since these functionality
+ * is often delegated to self-contained validation modules. Developers must install a validator module
+ * upon initialization (or any other mechanism to enable execution from the account):
+ *
+ * ```solidity
+ * contract MyAccountERC7579 is AccountERC7579, Initializable {
+ *     constructor() EIP712("MyAccountRSA", "1") {}
+ *
+ *   function initializeAccount(address validator, bytes calldata validatorData) public initializer {
```
frangio, line 29: It's worth noting that this is vulnerable to frontrunning if used as a 7702 account.

I wonder if it might be worth defining a special-purpose `accountInitializer` modifier that is "safe for accounts" whether used as 7702 templates or standalone contracts.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,356 @@
...
+        if (callType == ERC7579Utils.CALLTYPE_BATCH) return executionCalldata.execBatch(execType);
+        if (callType == ERC7579Utils.CALLTYPE_DELEGATECALL) return executionCalldata.execDelegateCall(execType);
+        revert ERC7579Utils.ERC7579UnsupportedCallType(callType);
+    }
+
+    /**
+     * @dev Lowest-level signature validation function. See {ERC7739Signer-_rawSignatureValidation}.
+     *
+     * This function delegates the signature validation to a validation module if the first 20 bytes of the
+     * signature correspond to an installed validator module.
```
frangio, line 195: I asked before but is this standard? Or is it at least conventional across existing projects?

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,356 @@
...
+    function _installModule(uint256 moduleTypeId, address module, bytes memory initData) internal virtual {
+        require(supportsModule(moduleTypeId), ERC7579Utils.ERC7579UnsupportedModuleType(moduleTypeId));
+        require(
+            IERC7579Module(module).isModuleType(moduleTypeId),
+            ERC7579Utils.ERC7579MismatchedModuleTypeId(moduleTypeId, module)
+        );
+
+        require(
+            (moduleTypeId != MODULE_TYPE_VALIDATOR || _validators.add(module)) &&
+                (moduleTypeId != MODULE_TYPE_EXECUTOR || _executors.add(module)),
```
frangio, line 247: Side effects in a require condition is not a good pattern IMO, I would refactor this.

contracts/account/extensions/AccountERC7579.sol
frangio: ERC-7579 explicitly states that "This standard does not dictate how validator selection is implemented." It seems wrong to implement a specific selection mechanism that is not standardized.

It might be just as or more common to have a single validator module installed and to use an unprefixed signature.

ernestognw: This is based on the [Reference Implementation](https://github.com/erc7579/erc7579-implementation/blob/main/src/MSAAdvanced.sol#L289), which naturally inspired Rhinestone's implementations like the [Safe adapter](https://github.com/rhinestonewtf/safe7579/blob/bb29e8b1a66658790c4169e72608e27d220f79be/src/Safe7579.sol#L350) ([also uses the same mechanism for nonces](https://github.com/rhinestonewtf/safe7579/blob/bb29e8b1a66658790c4169e72608e27d220f79be/src/Safe7579.sol#L265-L267)). It is also present [in Biconomy's Nexus](https://github.com/bcnmy/nexus/blob/54f4e19baaff96081a8843672977caf712ef19f4/contracts/Nexus.sol#L241) (same for [nonces](https://github.com/bcnmy/nexus/blob/54f4e19baaff96081a8843672977caf712ef19f4/contracts/lib/NonceLib.sol#L15-L19)) and [Etherspot's Prime Account](https://github.com/etherspot/etherspot-prime-contracts/blob/cfcdb48c4172cea0d66038324c0bae3288aa8caa/src/modular-etherspot-wallet/wallet/ModularEtherspotWallet.sol#L258) (same for [nonces](https://github.com/etherspot/etherspot-prime-contracts/blob/cfcdb48c4172cea0d66038324c0bae3288aa8caa/src/modular-etherspot-wallet/wallet/ModularEtherspotWallet.sol#L225-L228))

However, other providers like Kernel uses a more complex scheme (see [types](https://github.com/zerodevapp/kernel/blob/cfedcb9b047838f72491baaadef8fe7a95b2c837/src/types/Types.sol) and how [these are packed in signature and nonces](https://github.com/zerodevapp/kernel/blob/cfedcb9b047838f72491baaadef8fe7a95b2c837/src/utils/ValidationTypeLib.sol#L61-L102).

I'll create an `_extractNonceValidator` and `_extractSignatureValidator` internal functions for customization. The reason why it can't be a single version for both is that `validateUserOp` has access to the user operation while `isValidSignature` doesn't.

ernestognw: Solved in [c2612d0](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/46/commits/c2612d05a07ef29acba2484fc84c590a783befb4)

contracts/account/extensions/AccountERC7579.sol
ernestognw: Done at [e2c56df](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/46/commits/e2c56df5a5426aa7a1678728614ca49d0fb0bd5c)

contracts/account/extensions/AccountERC7579.sol
ernestognw: We can warn about this in concrete, I'm not sure if ERC-7579 will be default to enable 7702 accounts

ernestognw (commented):
One last thing we need to decide is whether to add the `withHook` modifier to the public `installModule` and `uninstallModule` functions. It is `RECOMMENDED` in ERC-7579 and I would say it's a great guarantee to have a default way to restrict bypassing the account logic by just uninstalling/installing a module.

Recently we spoke with the ERC-6900 team and they do recommend using hooks for all "native functions" (i.e. functions not implemented by modules). See [Alchemy's implementation.](https://github.com/alchemyplatform/modular-account/blob/develop/src/account/ModularAccountBase.sol#L127C14-L133)

We're generally not too strict on adding functionality since it's easier to add than remove, but I think this is one of those case where we wouldn't want to be opinionated

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+import {Packing} from "@openzeppelin/contracts/utils/Packing.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {ERC7739Signer} from "../../utils/cryptography/ERC7739Signer.sol";
+import {AccountCore} from "../AccountCore.sol";
+
+/**
+ * @dev Extension of {AccountCore} that implements support for ERC-7579 modules.
+ *
+ * To comply with the ERC-1271 support requirement, this contract implements {ERC7739Signer} as an
+ * opinionated layer to avoid signature replayability across accounts controlled by the same key.
```
Amxx, line 19: Do we want that here. It feels to me like 7579 and 7739 are independant. Sure more devs may want to use both, but I don't hink we should enforce 7739 here.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+ *     _installModule(MODULE_TYPE_VALIDATOR, validator, validatorData);
+ *   }
+ * }
+ * ```
+ *
+ * NOTE: Hook support is not included. See {AccountERC7579Hooked} for a version that hooks to execution.
+ */
+abstract contract AccountERC7579 is
+    AccountCore,
+    ERC7739Signer,
```
Amxx, line 39: I'd remove that inheritance
```suggestion
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+     * @dev Returns whether the account supports the given execution mode.
+     *
+     * Supported call types:
+     * * Single (`0x00`): A single transaction execution.
+     * * Batch (`0x01`): A batch of transactions execution.
+     * * Delegate (`0xff`): A delegate call execution.
+     *
+     * Supported exec types:
+     * * Default (`0x00`): Default execution type (revert on failure).
+     * * Try (`0x01`): Try execution type (emits ERC7579TryExecuteFail on failure).
```
Amxx, line 78: The exec type is not checked. Should we check it ?

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+        if (moduleTypeId == MODULE_TYPE_VALIDATOR) return _validators.contains(module);
+        if (moduleTypeId == MODULE_TYPE_EXECUTOR) return _executors.contains(module);
+        if (moduleTypeId == MODULE_TYPE_FALLBACK) return _fallbacks[bytes4(additionalContext[0:4])] == module;
+        return false;
+    }
+
+    /// @inheritdoc IAccountExecute
+    function executeUserOp(PackedUserOperation calldata userOp, bytes32 /*userOpHash*/) public virtual onlyEntryPoint {
+        Address.functionDelegateCall(address(this), userOp.callData[4:]);
+    }
```
Amxx, line 138: Do we need that ? We removed it from the core. I don't think we should ahve it here (this means removing the IAccountExecute inheritance)

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        if (signature.length < 20) return false;
+        address module = _extractSignatureValidator(signature);
+        return
+            isModuleInstalled(MODULE_TYPE_VALIDATOR, module, msg.data) &&
+            IERC7579Validator(module).isValidSignatureWithSender(address(this), hash, signature[20:]) ==
+            IERC1271.isValidSignature.selector;
+    }
```
Amxx, line 205: The lack of super call here prevents combining this with (for example) an ERC-7702 signer.

There may be cases where someone wants a 7702 account that also has 7579 modularity for compatibility with third party modules. In that case we'll want something that fallback from one option to another.

It may be as simple as doing
```suggestion
    function _rawSignatureValidation(
        bytes32 hash,
        bytes calldata signature
    ) internal view virtual override returns (bool) {
        if (super._rawSignatureValidation(hash, signature)) {
            return true;
        } else {
            if (signature.length < 20) return false;
            address module = _extractSignatureValidator(signature);
            return
                isModuleInstalled(MODULE_TYPE_VALIDATOR, module, msg.data) &&
                IERC7579Validator(module).isValidSignatureWithSender(address(this), hash, signature[20:]) ==
                IERC1271.isValidSignature.selector;
        }
    }
```

Amxx: note that inheritance ordering may be an issue. We may have to change the Signer's implementation to also do that fallback

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+            IERC7579Module(module).isModuleType(moduleTypeId),
+            ERC7579Utils.ERC7579MismatchedModuleTypeId(moduleTypeId, module)
+        );
+
+        bool exists;
+        if (moduleTypeId == MODULE_TYPE_VALIDATOR) exists = !_validators.add(module);
+        if (moduleTypeId == MODULE_TYPE_EXECUTOR) exists = !_executors.add(module);
+        require(!exists, ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
+
+        if (moduleTypeId == MODULE_TYPE_FALLBACK) {
```
Amxx, line 241: IMO this would be clearer (also it avoid the double negation)

```suggestion
        if (moduleTypeId == MODULE_TYPE_VALIDATOR) {
            require(_validators.add(module), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module);
        } else if (moduleTypeId == MODULE_TYPE_EXECUTOR) {
            require(_executors.add(module) , ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module);
        } else if (moduleTypeId == MODULE_TYPE_FALLBACK) {
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+            (moduleTypeId != MODULE_TYPE_VALIDATOR || _validators.remove(module)) &&
+                (moduleTypeId != MODULE_TYPE_EXECUTOR || _executors.remove(module)),
+            ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module)
+        );
+
+        if (moduleTypeId == MODULE_TYPE_FALLBACK) {
+            bytes4 selector;
+            (selector, deInitData) = abi.decode(deInitData, (bytes4, bytes));
+            require(_uninstallFallback(module, selector), ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module));
+        }
```
Amxx, line 275: For consistency with install

```suggestion
        if (moduleTypeId == MODULE_TYPE_VALIDATOR) {
            require(_validators.remove(module), ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module);
        } else if (moduleTypeId == MODULE_TYPE_EXECUTOR) {
            require(_executors.remove(module) , ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module);
        } else if (moduleTypeId == MODULE_TYPE_FALLBACK) {
            bytes4 selector;
            (selector, deInitData) = abi.decode(deInitData, (bytes4, bytes));
            require(_uninstallFallback(module, selector), ERC7579Utils.ERC7579UninstalledModule(moduleTypeId, module));
        }
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+
+        IERC7579Module(module).onUninstall(deInitData);
+        emit ModuleUninstalled(moduleTypeId, module);
+    }
+
+    /**
+     * @dev Installs a fallback handler for the given selector. Returns true if the handler was installed,
+     */
+    function _installFallback(address module, bytes4 selector) internal virtual returns (bool) {
```
Amxx, line 284: _installModule is already internal, and can be overriden. I'd make that one private so that fallbacks are not modified without an event (and the onInstall/onUninstall call being performed)

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+        if (_fallbacks[selector] != address(0)) return false;
+        _fallbacks[selector] = module;
+        return true;
+    }
+
+    /**
+     * @dev Uninstalls a fallback handler for the given selector. Returns true if the handler was uninstalled,
+     */
+    function _uninstallFallback(address module, bytes4 selector) internal virtual returns (bool) {
```
Amxx, line 293:
```suggestion
    function _uninstallFallback(address module, bytes4 selector) private returns (bool) {
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+    ///
+    /// To construct a signature, set the first 20 bytes as the module address and the remaining bytes as the
+    /// signature data:
+    ///
+    /// ```
+    /// <module address (20 bytes)> | <signature data>
+    /// ```
+    function _extractSignatureValidator(bytes calldata signature) internal pure virtual returns (address) {
+        return address(bytes20(signature[0:20]));
+    }
```
Amxx, line 326: I think having two different mechanisms for extracting the validator based on the usecase is really confusing. IMO we should only have one.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,370 @@
...
+     */
+    function _fallback() internal virtual {
+        address handler = _fallbackHandler(msg.sig);
+        require(handler != address(0), ERC7579MissingFallbackHandler(msg.sig));
+
+        // From https://eips.ethereum.org/EIPS/eip-7579#fallback[ERC-7579 specifications]:
+        // - MUST utilize ERC-2771 to add the original msg.sender to the calldata sent to the fallback handler
+        // - MUST use call to invoke the fallback handler
+        (bool success, bytes memory returndata) = handler.call{value: msg.value}(
+            abi.encodePacked(msg.data, msg.sender)
```
Amxx, line 344: shoud we use Context here ?

I know context is going to go away, and I understand that its unlikelly this will be used with an ERC2771Context ... but it could happen and I wouldn't create inconsistencies in our usage of Context in 5.x

```suggestion
            abi.encodePacked(_msgData(), _msgSender())
```

contracts/account/extensions/AccountERC7579.sol
Amxx: I just realised that removing the `_validateUserOp` override and relying on `AccountCore._validateUserOp` calling `_rawSignatureValidation` could work, but it would use `IERC7579Validator.isValidSignatureWithSender` instead of `IERC7579Validator.validateUserOp`... what a mess

using the nonce is a nice trick because it avoid copying the entier userOp to memory to rewrite the signature before sending is again to the module ... but still it feels very hacky (and inconsistent in the way signatures are generated)

Amxx: If there was a standard encoding of the module address in the signature, and if the module knew about it, it would be cleaner.

Not sure if we can assume the module know it will have to drop the first 20 bytes of the userOp signature that are used for module selection. Maybe we should try to make that "standard validation module behavior"

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,75 @@
...
+    /// @dev Hooked version of {AccountERC7579-_execute}.
+    function _execute(
+        Mode mode,
+        bytes calldata executionCalldata
+    ) internal virtual override withHook returns (bytes[] memory) {
+        return super._execute(mode, executionCalldata);
+    }
+
+    /// @dev Installs a module with support for hook modules. See {AccountERC7579-_installModule}
+    function _installModule(uint256 moduleTypeId, address module, bytes memory initData) internal virtual override {
```
Amxx, line 54: should this throw an `ERC7579AlreadyInstalledModule` if there is already a _hook ?

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,75 @@
...
+        if (moduleTypeId == MODULE_TYPE_HOOK) _hook = module;
+        super._installModule(moduleTypeId, module, initData);
+    }
+
+    /// @dev Uninstalls a module with support for hook modules. See {AccountERC7579-_uninstallModule}
+    function _uninstallModule(uint256 moduleTypeId, address module, bytes memory deInitData) internal virtual override {
+        if (moduleTypeId == MODULE_TYPE_HOOK) _hook = address(0);
```
Amxx, line 61: should this throw an `ERC7579AlreadyInstalledModule` if there is already a _hook ?

```suggestion
        if (moduleTypeId == MODULE_TYPE_HOOK) {
            require(_hook == address(0), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
            _hook = address(0);
        }
```

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,75 @@
...
+    function _execute(
+        Mode mode,
+        bytes calldata executionCalldata
+    ) internal virtual override withHook returns (bytes[] memory) {
+        return super._execute(mode, executionCalldata);
+    }
+
+    /// @dev Installs a module with support for hook modules. See {AccountERC7579-_installModule}
+    function _installModule(uint256 moduleTypeId, address module, bytes memory initData) internal virtual override {
+        if (moduleTypeId == MODULE_TYPE_HOOK) _hook = module;
```
Amxx, line 55:
```suggestion
        if (moduleTypeId == MODULE_TYPE_HOOK) {
            require(_hook, ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
            _hook = module;
        }
```

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,74 @@
...
+        super._uninstallModule(moduleTypeId, module, deInitData);
+    }
+
+    /// @dev Supports hook modules. See {AccountERC7579-supportsModule}
+    function supportsModule(uint256 moduleTypeId) public view virtual override returns (bool) {
+        return moduleTypeId == MODULE_TYPE_HOOK || super.supportsModule(moduleTypeId);
+    }
+
+    /// @dev Returns whether a hook module is installed.
+    function _isHookInstalled(address module) internal view virtual returns (bool) {
```
Amxx, line 71: Do we really need that ? I would remove it and inline the logic in `isModuleInstalled`

contracts/account/modules/ERC7579ECDSAValidator.sol
```diff
@@ -0,0 +1,53 @@
...
+pragma solidity ^0.8.20;
+
+import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import {ERC7579Validator} from "./ERC7579Validator.sol";
+
+/**
+ * @dev {ERC7579Validator} for {ECDSA} signatures.
+ */
+abstract contract ERC7579ECDSAValidator is ERC7579Validator {
+    mapping(address sender => address signer) private _associatedSigners;
```
Amxx, line 12: would it make sens that a sender supports multiple signers ?
```solidity 
mapping(address sender => mapping(address signer => bool isOperator))
```

We may even want to make that a N out of M multisig 

Beyond the "choice of cryptographic primitive", there are a lot of relevant validator that we may want to implement in terms of behavior. I'd leave that to a future PR (this one is already big with just the Account part)

contracts/account/modules/ERC7579P256Validator.sol
```diff
@@ -0,0 +1,62 @@
...
+
+import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
+import {ERC7579Validator} from "./ERC7579Validator.sol";
+
+/**
+ * @dev {ERC7579Validator} for {P256} signatures.
+ */
+abstract contract ERC7579P256Validator is ERC7579Validator {
+    mapping(address sender => bytes32) private _associatedQx;
+    mapping(address sender => bytes32) private _associatedQy;
```
Amxx, line 13: same logic as ERC7579ECDSA, I think we should work on the validator in a different PR.

But still here: 
```suggestion
    mapping(address sender => bytes32[2]) private _associatedQxQy;
```

test/helpers/signers.js
```diff
@@ -150,10 +150,6 @@ class RSASigningKey {
 }
 
 class RSASHA256SigningKey extends RSASigningKey {
-  static random(modulusLength = 2048) {
-    return new RSASHA256SigningKey(generateKeyPairSync('rsa', { modulusLength }));
-  }
-
```
Amxx, line 156: why remove that part ?

ernestognw: Because it's repeated code from the `RSASHA256SigningKey`, it's still available now that we're instantiating the signers with `this`

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,74 @@
...
+        return _hook;
+    }
+
+    /// @inheritdoc AccountERC7579
+    function isModuleInstalled(
+        uint256 moduleTypeId,
+        address module,
+        bytes calldata data
+    ) public view virtual override returns (bool) {
+        return _isHookInstalled(module) || super.isModuleInstalled(moduleTypeId, module, data);
```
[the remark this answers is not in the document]
ernestognw: There's only 1 hook module per account, can later get customized with a hook [multiplexer](https://docs.rhinestone.wtf/module-sdk/modules/hook-multi-plexer). If there's just 1 hook module, I think this is just fine. Alternatively we can pass the `moduleTypeId` and `data` to `_isHookInstalled` for completeness but it'd be useless in this implementation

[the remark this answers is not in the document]
Amxx: Sure there is only one hook module, but if someone calls `isModuleInstalled(MODULE_TYPE_VALIDATOR, <address of the hook module, 0x)`, that should return false IMO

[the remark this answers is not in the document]
Amxx: My proposal would be 
```suggestion
        return (moduleTypeId == MODULE_TYPE_HOOK && module == hook()) 
            || super.isModuleInstalled(moduleTypeId, module, data);
```

contracts/account/modules/ERC7579ECDSAValidator.sol
Amxx: Would it be an acceptable middle ground to remove the `contracts/accounts/modules` from this PR (or move the necessary one to the mocks) and open PRs for these after we merge this one (with "just" the `contracts/accounts/extensions` ?)

contracts/account/extensions/AccountERC7579Hooked.sol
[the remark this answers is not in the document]
ernestognw: Currently, it's not possible to install a module that's not of the specified `moduleTypeId`, so we can expect that if a hook is installed, it should be a hook so this check may be redundant.

Except on upgradeable contracts... thoughts?

contracts/account/extensions/AccountERC7579.sol
Amxx: I missed the part of the ERC that says
> The smart account MUST implement the ERC-1271 interface

resolving

contracts/account/extensions/AccountERC7579.sol
ernestognw: The ERC say it's RECOMMENDED to implement this function in this way

> If an account chooses to implement executeUserOp, this method SHOULD ensure the account executes userOp.calldata except 4 most significant bytes, which are reserved for executeUserOp.selector as per ERC-4337. Thus the userOp.callData[4:] should represent the calldata for a valid call to the account. It is RECOMMENDED that the account executes a delegatecall in order to preserve the original msg.sender to the account.

I would agree since it is a way to ensure hooks are consistently executed. I would agree with moving it to the `AccountERC7579Hooked` instead

contracts/account/extensions/AccountERC7579Hooked.sol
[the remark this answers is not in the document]
Amxx: That is a public function. We should also consider lookups that are not done internally during the installation of modules.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -0,0 +1,340 @@
...
+            ERC7579Utils.ERC7579MismatchedModuleTypeId(moduleTypeId, module)
+        );
+
+        if (moduleTypeId == MODULE_TYPE_VALIDATOR) {
+            require(_validators.add(module), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
+        } else if (moduleTypeId == MODULE_TYPE_EXECUTOR) {
+            require(_executors.add(module), ERC7579Utils.ERC7579AlreadyInstalledModule(moduleTypeId, module));
+        } else if (moduleTypeId == MODULE_TYPE_FALLBACK) {
+            bytes4 selector;
+            (selector, initData) = abi.decode(initData, (bytes4, bytes));
```
Amxx, line 237: This decoding forces a calldata format that is quite long with many zeros.

Reference implementation does is differently: https://github.com/erc7579/erc7579-implementation/blob/main/src/core/ModuleManager.sol#L231-L241. 

Should we decode in a similar way ?

Reference implementation also supports forcing the call fallback mechanism to be static. Do we want that ?

contracts/account/extensions/AccountERC7579.sol
Amxx: see https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/e26066dbcb0f160307ffacc9969389673710ab0b/contracts/mocks/account/AccountERC7702WithModulesMock.sol

contracts/account/extensions/AccountERC7579.sol
Amxx: Just noticed that Biconomy's approach is similar (use the nonce), but it doesn't use the same part of the nonce (incompatible)

package.json
```diff
@@ -23,7 +23,8 @@
     "lint:sol:fix": "prettier --log-level warn --ignore-path .gitignore '{contracts,test}/**/*.sol' --write",
     "coverage": "scripts/checks/coverage.sh",
     "test": "hardhat test",
-    "test:inheritance": "scripts/checks/inheritance-ordering.js artifacts/build-info/*"
+    "test:inheritance": "scripts/checks/inheritance-ordering.js artifacts/build-info/*",
+    "test:pragma": "scripts/checks/pragma-consistency.js artifacts/build-info/*"
```
ernestognw, line 27: Wouldn't be better if we use the `lib/openzeppelin` script?

```suggestion
    "test:pragma": ".lib/openzeppelin/scripts/checks/pragma-consistency.js artifacts/build-info/*"
```

scripts/checks/pragma-consistency.js
ernestognw, line 1: We can create a symlink for this too

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,92 @@
...
+        address module,
+        bytes calldata data
+    ) public view virtual override returns (bool) {
+        return
+            (moduleTypeId == MODULE_TYPE_HOOK && module == hook()) ||
+            super.isModuleInstalled(moduleTypeId, module, data);
+    }
+
+    /// @dev Installs a module with support for hook modules. See {AccountERC7579-_installModule}
+    /// TODO: withHook? based on what value?
```
ernestognw, line 53: What's this comment for? I understand the hook must decide what to do depending on the value.

A way to support multiple hooks I think we should recommend is using a Hook Multiplexer
https://github.com/rhinestonewtf/core-modules/blob/e123a910ba6e5e81878a1aea7265c2fa7815b90b/src/HookMultiPlexer/HookMultiPlexer.sol#L4

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -0,0 +1,92 @@
...
+import {IERC7579Hook, MODULE_TYPE_HOOK} from "@openzeppelin/contracts/interfaces/draft-IERC7579.sol";
+import {ERC7579Utils, Mode} from "@openzeppelin/contracts/account/utils/draft-ERC7579Utils.sol";
+import {AccountERC7579} from "./AccountERC7579.sol";
+
+/**
+ * @dev Extension of {AccountERC7579} with support for a single hook module (type 4).
+ *
+ * If installed, this extension will call the hook module's {IERC7579Hook-preCheck} before
+ * executing any operation with {_execute} (including {execute} and {executeFromExecutor} by
+ * default) and {IERC7579Hook-postCheck} thereafter.
```
ernestognw, line 14: My concern is that someone may implement `executeUserOp` without hooks. I know it's not that likely and perhaps I'm worrying too much, but I want to make extremely clear how hooks must be used. Especially since these are recommended for access management and permissions

```suggestion
 * default) and {IERC7579Hook-postCheck} thereafter.
 *
 * NOTE: The Account does not implement {IAccountExecute-executeUserOp} since ERC-7579 already
 * provides with an execution mechanism through {execute} and {executeFromModule}. Consider ensuring
 * hooks are executed too when implementing this or any other function that allows arbitrary calls from the account.
```

contracts/mocks/account/AccountERC7702WithModulesMock.sol
```diff
@@ -0,0 +1,45 @@
...
+        // Force resolution to AccountERC7579 (discard ERC7821 implementation)
+        AccountERC7579.execute(mode, executionData);
+    }
+
+    function supportsExecutionMode(
+        bytes32 mode
+    ) public view virtual override(AccountERC7579, ERC7821) returns (bool result) {
+        // Force resolution to AccountERC7579 (discard ERC7821 implementation)
+        return AccountERC7579.supportsExecutionMode(mode);
+    }
```
ernestognw, line 27: Thinking more about this. I'm leaning towards removing ERC-7821 from Account by default. I envision the Wizard starts with Account.sol and requires the user to select an execution mechanism:

- [ ] Minimal (7821)
- [ ] Modular (7579)

package.json
Amxx: The script does 
```javascipt
const { output: solcOutput } = require(path.resolve(__dirname, '../..', artifact));
```

Your approach would use end up looking in tghe wrong place 

```
Error: Cannot find module '/home/amxx/Work/OZ/openzeppelin-community-contracts/lib/@openzeppelin-contracts/artifacts/build-info/2387dcef554d84c915c13f59cc3d0841.json'
```

scripts/checks/pragma-consistency.js
Amxx: lets do that

Amxx: that causes the same error, because of dirname.

It feels to me like reusing script is going to take some effort (and cahnging the scripts in the main repo so that they use a config that can be changed here

contracts/account/extensions/AccountERC7579Hooked.sol
Amxx: this is old. cleaning up

scripts/checks/pragma-consistency.js
ernestognw: Yes I agree, let's keep it for now

contracts/account/extensions/AccountERC7579Hooked.sol
Amxx: My understanding is that the execution can be done in 2 ways:

- Case 1: the calldata does not start with the selector for `executeUserOp`.

In that case, the entrypoint passes the calldata to the smart account. This could be a call to `execute`, `installModule`, or basically anything else (`changeMultisigThreshold`, ...) that the SCA implements. If the call targets a function that includes the `withHook()` modifier, then the hooks are called. If the function does not include the modifier, the hooks are not called. Simple

- Case 2: the calldata start with the `executeUserOp` selector. 

In that case, the entrypoint takes the entire userOperation and passes it to the SCA. The purpose of this is that unlike case 1, the SCA has access to everything (nonce, paymasterdata, ...). The point of that is to implement (during execution) checks that would break validation rules (unauthorized storage read). Once these checks are done, the common practice is to do a delegate call to the SCA (`address(this)`) with the innercall (`userOp.callData[4 :]`).

At that point you may be calling `execute`, or `installModule` (which will trigger a hook) or you may execute something else (`changeMultisigThreshold`?) that may not include a hook. In any case, you are NOT doing an external call, and you are still calling the hooks for the function you target (if applicable). If the `executeUserOp` does the hook call, then you may be doing it twice (once with the full userOp and once with the innercall).

That is why I'm thinking:
- usecase for `executeUserOp` is not really execution but rather "execute-time validation".
- having hooks modifier in `executeUserOp` is usefull if you want a hook to take care of that validation (not actually a bad idea, its quite modular)
- forgetting to put hooks in `executeUserOp` would not cause any issues regarding "avoiding" hooks. All functions that `executeUserOp` would be able to trigger without hooks can already be called without triggering the hooks (see case1). If would just disable the modular "execute-time validation".

When I started writting all that my point was to argue that `executeUserOp` is not an execution mechanism, and that the comment was partly inacurate. But now that I finished writing it, I think we should implement `executeUserOp` here.

ernestognw: Now I understand you don't see an issue in bypassing hooks if the developers wants to, which would be the case 1.

In the second case, I do agree the common practice is to do a delegate call to `address(this)`. I think it's also suggested in the standard. So to keep consistency, it's worth implementing it at least in the case of `AccountERC7579`.

My concern is that for the second case, someone would implement a function that **does not delegate call to itself**, which kinda complicates the things in regards to when to expect hook execution or not. I do not want that to be confusing and users can always override if they really need to implement it differently.

I would agree with implementing `executeUserOp` here.

ernestognw: After discussing internally we agreed on not implementing `executeUserOp`. My main worry is that someone would implement it as:

```
function executeUserOp(PackedUserOperation calldata userOp, bytes32 /* userOpHash */) public virtual { Address.functionCall(...); }
```

But generally, that's a bad practice. So it should be generally fine.

ernestognw (approved):

contracts/account/extensions/AccountERC7579Hooked.sol
Amxx: Erratum: I though ERC7579 was recommanding that `executeUserOp` would be hooked. IMO that would make some sens for modular validation. That is not the case.

So if `executeUserOp` is not with hooked, the fact that it cannot be used for hook escaping (if implemented according to the recommandation) makes me think we should not implement it. Anyone that needs it for its inteended usecase, and does the delegatecall as recomanded, would be safe.

PR #49, opened 2024-12-20
Add ERC7821 to Account.sol

### LINKED ISSUE #52: Misuse of  `IAccountExecute.executeUserOp`
Currently, `AccountCore.sol` exposes the (optional) `IAccountExecute` interface, and implements `executeUserOp` in a way that perfoms an arbitray external call. It appears that this is a missuse of this function.

### What is `executeUserOp` for?

`executeUserOp` is handled differently from other functions by the entrypoint. When calldata directly to the Account, `executeUserOp` sends the entier UserOperation. The usecase for that appears to be  so that checks can be performed during the execution phase. Some tests that break the storage access rules cannot be performed during the validation phase, but could be performed in `executeUserOp`. The downside of puting tests here is that if they fail, user (or paymaster) still have to pay the gas.

Therefore the most canonical implementation of this function is
```solidity
function executeUserOp(
    bytes32 /*userOpHash*/
) public virtual override onlyEntryPointOrSelf {
    // perform additional checks
    Address.functionDelegateCall(address(this), userOp.callData[4:]);
}
```

We could chose to implement it in a way were users can add execution checks by overriding it
```solidity 
function executeUserOp(
    bytes32 userOpHash
) public virtual override {
    // perform additional checks
    super.executeUserOp(userOp, userOpHash);
}
```

or we could chose to not implement it... but it feels wrong to use it for doing external calls that don't need to access the entier user operation.

### What do we replace it with?

Accounts needs a function to perform external calls. This function should be protected (`onlyEntryPoint`). It doesn't need to get the entire user operation, just some target, value and calldata.

ERC-7579 and ERC-7821 propose standard "execute" functions that would be a better choice here. Both option support batching.

- `IERC7579Account` contains both `execute`, `executeFromExecutor`. 
  - If we chose that option, only the first one would be part of `Account` (or `AccountCore`). We would have to chose if we also want to implement `IERC7579AccountConfig.supportsExecutionMode`. 
  - We would also have to decide if we support `CALLTYPE_DELEGATECALL` that not all wallets may want for security reasons.
  
- `IERC7821` appears to be that exact subset from ERC-7579 (`execute` and `supportsExecutionMode`). It still contains some complexity.

### Where to put it?

Putting it in `AccountCore.sol` feels opinnionated, but putting it in `Account.sol` (or another module that is inherited by `Account.sol`) would leave `AccountCore.sol` without an external call mechanism. IMO that is ok, because devs may want to use a different one (not documented here) ...but it would need proper documentation.

Amxx:
Fixes #52 by implementing ERC7821 (without the optional opData) for batch tx processing

### DIFF
--- a/contracts/account/Account.sol
+++ b/contracts/account/Account.sol
@@ -5,6 +5,8 @@ pragma solidity ^0.8.20;
 import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol";
 import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
 import {ERC7739Signer} from "../utils/cryptography/ERC7739Signer.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {Multicall} from "@openzeppelin/contracts/utils/Multicall.sol";
 import {AccountCore} from "./AccountCore.sol";
 
 /**
@@ -16,4 +18,10 @@ import {AccountCore} from "./AccountCore.sol";
  * NOTE: To use this contract, the {ERC7739Signer-_rawSignatureValidation} function must be
  * implemented using a specific signature verification algorithm. See {SignerECDSA}, {SignerP256} or {SignerRSA}.
  */
-abstract contract Account is AccountCore, ERC721Holder, ERC1155Holder, ERC7739Signer {}
+abstract contract Account is AccountCore, Multicall, ERC721Holder, ERC1155Holder, ERC7739Signer {
+    function execute(address target, uint256 value, bytes calldata data) public virtual onlyEntryPointOrSelf {
+        // we cannot use `Address.functionCallWithValue` here as it would revert on EOA targets
+        (bool success, bytes memory returndata) = target.call{value: value}(data);
+        Address.verifyCallResult(success, returndata);
+    }
+}
--- a/contracts/account/Account.sol
+++ b/contracts/account/Account.sol
@@ -11,6 +11,7 @@ import {AccountERC7821} from "./extensions/AccountERC7821.sol";
 /**
  * @dev Extension of {AccountCore} with recommended feature that most account abstraction implementation will want:
  *
+ * * {AccountERC7821} for performing external calls in batches.
  * * {ERC721Holder} and {ERC1155Holder} to accept ERC-712 and ERC-1155 token transfers transfers.
  * * {ERC7739Signer} for ERC-1271 signature support with ERC-7739 replay protection
  *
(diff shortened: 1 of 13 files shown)


contracts/account/Account.sol
```diff
@@ -5,6 +5,8 @@ pragma solidity ^0.8.20;
 import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Holder.sol";
 import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
 import {ERC7739Signer} from "../utils/cryptography/ERC7739Signer.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {Multicall} from "@openzeppelin/contracts/utils/Multicall.sol";
```
ernestognw, line 9: I feel that adding multicall here is a long discussion. @frangio and I already had mixed feelings regarding the `delegateCall` in `executeUserOp` so I think Multicall may be too much.

contracts/account/Account.sol
```diff
@@ -16,4 +18,10 @@ import {AccountCore} from "./AccountCore.sol";
  * NOTE: To use this contract, the {ERC7739Signer-_rawSignatureValidation} function must be
  * implemented using a specific signature verification algorithm. See {SignerECDSA}, {SignerP256} or {SignerRSA}.
  */
-abstract contract Account is AccountCore, ERC721Holder, ERC1155Holder, ERC7739Signer {}
+abstract contract Account is AccountCore, Multicall, ERC721Holder, ERC1155Holder, ERC7739Signer {
+    function execute(address target, uint256 value, bytes calldata data) public virtual onlyEntryPointOrSelf {
```
ernestognw, line 22: What's the rationale of having another `execute` function if the Account already has `executeUserOp`?

Amxx: `executeUserOp` takes A LOT of calldata, most of it is ignored/useless when we are not actually executing an ERC-4337 useroperation. When doing a `multicall`, you want a simpler function to call.

Amxx:
Put on hold. Batching needs more discussion. Maybe using ERC-7821

The fix and tests part are moved to #51

contracts/account/Account.sol
frangio: I don't think this is related to `executeUserOp`, but I'm against Multicall on the account, batches should be handled by some "executeBatch" function.

contracts/account/Account.sol
Amxx: Closed: using ERC-7821 instead

contracts/account/Account.sol
Amxx: Closed: using ERC-7821 instead

ernestognw (approved):
I removed IAccountExecute. The current implementation does not add any specific functionality. Although it may be used to check userOp fields, this is only a specific use case we've seen and not a behavior that we must consider "default".

ernestognw (approved):

ernestognw (approved):

ernestognw (approved):

PR #53, opened 2024-12-20
Remove mention of deleted `unfreeze` function in ERC20Custodian docs

arr00:


ernestognw (approved):
Nice catch too, thanks!

PR #58, opened 2025-01-06
Move EIP712 dependency from AccountCore to Account

Amxx:
As discussed during todays call, some account implementation may "vertically integrate" with the signer in a way were they raw sign the userOpHash (or similar).

Optional: we could remove the "default implementation" of `_signableUserOpHash` from `AccountCore`.


contracts/account/extensions/AccountSignerERC7702.sol
```diff
@@ -3,12 +3,12 @@
 pragma solidity ^0.8.20;
 
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
-import {AccountCore} from "../AccountCore.sol";
+import {AbstractSigner} from "../../utils/cryptography/AbstractSigner.sol";
 
 /**
  * @dev {Account} implementation whose low-level signature validation is done by an EOA.
  */
-abstract contract AccountSignerERC7702 is AccountCore {
```
arr00, line 11: This is no longer an extension of account?

contracts/account/extensions/AccountSignerERC7702.sol
```diff
@@ -3,12 +3,12 @@
 pragma solidity ^0.8.20;
 
 import {ECDSA} from "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
-import {AccountCore} from "../AccountCore.sol";
+import {AbstractSigner} from "../../utils/cryptography/AbstractSigner.sol";
 
 /**
  * @dev {Account} implementation whose low-level signature validation is done by an EOA.
```
Amxx, line 9:
```suggestion
 * @dev Implementation of {AbstractSigner} for implementation of ERC-7702 accounts whose low-level signature 
 * validation is done by an EOA.
```

contracts/account/AccountCore.sol
```diff
@@ -93,29 +83,12 @@ abstract contract AccountCore is AbstractSigner, EIP712, IAccount {
      *
      * Given the `userOpHash` calculation is defined by ERC-4337, offchain signers
      * may need to sign again this hash by rehashing it with other schemes (e.g. ERC-191).
-     *
-     * Returns a typehash following EIP-712 typed data hashing for readability.
      */
     function _signableUserOpHash(
-        PackedUserOperation calldata userOp,
-        bytes32 /* userOpHash */
+        PackedUserOperation calldata /*userOp*/,
```
ernestognw, line 88:
```suggestion
        PackedUserOperation calldata /* userOp */,
```

contracts/account/extensions/AccountSignerERC7702.sol
ernestognw: Yeah, I would agree with this change. The reasoning is that with ERC-7702 some users may want to rely on AccountSignerERC7702 to use with a separate version of ERC-7821 (see [this PR](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/61))

contracts/account/extensions/AccountSignerERC7702.sol
```diff
@@ -3,12 +3,13 @@
...
-import {AccountCore} from "../AccountCore.sol";
+import {AbstractSigner} from "../../utils/cryptography/AbstractSigner.sol";
 
 /**
- * @dev {Account} implementation whose low-level signature validation is done by an EOA.
+ * @dev Implementation of {AbstractSigner} for implementation of ERC-7702 accounts whose low-level signature 
+ * validation is done by an EOA.
  */
-abstract contract AccountSignerERC7702 is AccountCore {
+abstract contract AccountSignerERC7702 is AbstractSigner {
```
ernestognw, line 12: Perhaps we might just name this `SignerERC7702`

```suggestion
abstract contract SignerERC7702 is AbstractSigner {
```

And move it to the `cryptography` folder (?)

Amxx: Yes this is a signer, but unlike the other ones, I don't think this one makes sens outside of an account. That is why I kept the name and location. But I you think it makes more sens to rename and move it I wouldn't oppose it.

ernestognw: I'm trying to thing about the developer experience of this. Most would understand that ERC-7702 enables signature for EOAs and these use ECDSA, so I'd be worried they'd start from a SignerECDSA instead where is kinda weird that you'll need to initialize by calling `_initializeSigner`.

I'm leaning towards moving it so I'll push a commit

contracts/account/extensions/AccountSignerERC7702.sol
ernestognw: Moved

ernestognw:
> we could remove the "default implementation" of _signableUserOpHash from AccountCore.

I'm not sure since this is in the middle of `validateUserOp`, which makes it difficult to replace without overriding the whole validation

ernestognw (dismissed):
lfg! 🚀

Amxx:
>> we could remove the "default implementation" of _signableUserOpHash from AccountCore.
>
> I'm not sure since this is in the middle of validateUserOp, which makes it difficult to replace without overriding the whole validation

I did not mean to remove the function definition. I meant to keep the function definition, but without an implementation.

```solidity
    function _signableUserOpHash(
        bytes32 userOpHash
    ) internal view virtual returns (bytes32);
```

ernestognw (approved):

PR #61, opened 2025-01-07
Split ERC-7821 executor into its own contract

ernestognw:
## Motivation

ERC-7821 might be used for EOAs with EIP-7702 to achieve batched execution without any ERC-4337 capability

### DIFF
--- a/contracts/account/Account.sol
+++ b/contracts/account/Account.sol
@@ -6,16 +6,21 @@ import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Hol
 import {ERC1155Holder} from "@openzeppelin/contracts/token/ERC1155/utils/ERC1155Holder.sol";
 import {ERC7739Signer} from "../utils/cryptography/ERC7739Signer.sol";
 import {AccountCore} from "./AccountCore.sol";
-import {AccountERC7821} from "./extensions/AccountERC7821.sol";
+import {ERC7821} from "./extensions/ERC7821.sol";
 
 /**
  * @dev Extension of {AccountCore} with recommended feature that most account abstraction implementation will want:
  *
- * * {AccountERC7821} for performing external calls in batches.
+ * * {ERC7821} for performing external calls in batches.
  * * {ERC721Holder} and {ERC1155Holder} to accept ERC-712 and ERC-1155 token transfers transfers.
  * * {ERC7739Signer} for ERC-1271 signature support with ERC-7739 replay protection
  *
  * NOTE: To use this contract, the {ERC7739Signer-_rawSignatureValidation} function must be
  * implemented using a specific signature verification algorithm. See {SignerECDSA}, {SignerP256} or {SignerRSA}.
  */
-abstract contract Account is AccountCore, AccountERC7821, ERC721Holder, ERC1155Holder, ERC7739Signer {}
+abstract contract Account is AccountCore, ERC7821, ERC721Holder, ERC1155Holder, ERC7739Signer {
+    /// @inheritdoc ERC7821
+    function execute(bytes32 mode, bytes calldata executionData) public payable virtual override onlyEntryPointOrSelf {
+        super.execute(mode, executionData);
+    }
+}
--- a/contracts/account/extensions/ERC7821.sol
+++ b/contracts/account/extensions/ERC7821.sol
@@ -4,18 +4,25 @@ pragma solidity ^0.8.20;
 
 import {ERC7579Utils, Mode, CallType, ExecType, ModeSelector} from "@openzeppelin/contracts/account/utils/draft-ERC7579Utils.sol";
 import {IERC7821} from "../../interfaces/IERC7821.sol";
-import {AccountCore} from "../AccountCore.sol";
 
 /**
- * @dev Minimal batch executor following ERC7821. Only supports basic mode (no optional "opData").
+ * @dev Minimal batch executor following ERC-7821. Only supports basic mode (no optional "opData").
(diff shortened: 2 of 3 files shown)


contracts/account/extensions/ERC7821.sol
```diff
@@ -4,18 +4,25 @@ pragma solidity ^0.8.20;
...
+    /**
+     * @dev Executes the calls in `executionData` with no optional `opData` support.
+     *
+     * NOTE: When using with an {Account}, make sure to use wrap this function with
+     * an authorization check that ensures the caller is the account itself or
+     * the entrypoint.
+     *
+     * Reverts and bubbles up error if any call fails.
+     */
+    function execute(bytes32 mode, bytes calldata executionData) public payable virtual {
```
Amxx, line 25: I'm woried that it would be a very easy (and critical) mistake to forget to add some auth check in an overide. Also, there might be a risk that the authorisation check is lost when resolving inheritance conflict is super is not called properly.

I'd prefer using something like `UUPSUpgradeable`

```solidity
function execute(bytes32 mode, bytes calldata executionData) public payable virtual {
    if (!_erc7821restriction()) revert SomeError();
    if (!supportsExecutionMode(mode)) revert UnsupportedExecutionMode();
    executionData.execBatch(ERC7579Utils.EXECTYPE_DEFAULT);
}

function _erc7821restriction() internal virtual returns(bool) {
    return msg.sender == address(this);
}
```

That can be overriden by doing 
```
function _erc7821restriction() internal virtual override returns(bool) {
    return super._erc7821restriction() || msg.sender == entrypoint;
}
```

That way we always start with more security, and it can be relaxed, instead of doing the oposite.

Amxx: implemented in e6b96b640a82ed6f2a1aec2c4c00438711bcca8c

ernestognw: Nice, I was thinking about this but didn't want to add extra functions.

It looks good, but I'd pass `mode` and `executionCalldata` to `erc7821restriction()`. The reason is that in case `opData` is available, it should be authenticated there.

Amxx: that is fair

contracts/account/Account.sol
```diff
@@ -6,16 +6,21 @@ import {ERC721Holder} from "@openzeppelin/contracts/token/ERC721/utils/ERC721Hol
...
- * * {AccountERC7821} for performing external calls in batches.
+ * * {ERC7821} for performing external calls in batches.
  * * {ERC721Holder} and {ERC1155Holder} to accept ERC-712 and ERC-1155 token transfers transfers.
  * * {ERC7739Signer} for ERC-1271 signature support with ERC-7739 replay protection
  *
  * NOTE: To use this contract, the {ERC7739Signer-_rawSignatureValidation} function must be
  * implemented using a specific signature verification algorithm. See {SignerECDSA}, {SignerP256} or {SignerRSA}.
  */
-abstract contract Account is AccountCore, AccountERC7821, ERC721Holder, ERC1155Holder, ERC7739Signer {}
+abstract contract Account is AccountCore, ERC7821, ERC721Holder, ERC1155Holder, ERC7739Signer {
```
Amxx, line 21:
```suggestion
abstract contract Account is AccountCore, ERC721Holder, ERC1155Holder, ERC7739Signer, ERC7821 {
```

contracts/account/extensions/ERC7821.sol
```diff
@@ -0,0 +1,57 @@
...
+            modeSelector == ModeSelector.wrap(0x00000000);
+    }
+
+    /**
+     * @dev Access control mechanism for the {execute} function.
+     */
+    function _erc7821AuthorizedExecutor(
+        bytes32 /* mode */,
+        bytes calldata /* executionData */
+    ) internal view virtual returns (bool) {
```
Amxx, line 46: any reason to to remove `address caller` from the parameter, and using `msg.sender` implicitelly. I have the feeling its nicer if this internal function can be called to check the permission of someone else.

contracts/account/extensions/ERC7821.sol
```diff
@@ -0,0 +1,57 @@
...
+        return msg.sender == address(this);
+    }
+
+    // slither-disable-next-line write-after-write
+    function _emptyCalldataBytes() private pure returns (bytes calldata result) {
```
Amxx, line 51: We'll need an helper for that at some point. (but not in this PR)

ernestognw: haha yeah I thought the same

contracts/account/extensions/ERC7821.sol
ernestognw: I wanted to make this function composable with `onlyOwner`, `onlyRole` or `restricted` modifiers in the same way `authorizeUpgrade` is in UUPSUpgradeable. I start to think that the `AccessManaged` contract must expose a variant of the `restricted` modifier that allow passing in a custom address. Also note those modifiers generally revert and not return a boolean

ernestognw: Worth noting that if we want this function to be composable with `AccessManaged.restricted`, then it can't have `view` visibility :(

ernestognw:
Agree with the changes @Amxx, I'm ok merging this if you approve but please check [this](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/61#discussion_r1909446017)

Amxx (approved):

Amxx (approved):

PR #64, opened 2025-01-09
Remove Ownable dependency in AxelarGatewayBase

ernestognw:
## Motivation

This allow to setup custom access control easily (e.g. through Wizard)

### DIFF
--- a/contracts/crosschain/axelar/AxelarGatewayBase.sol
+++ b/contracts/crosschain/axelar/AxelarGatewayBase.sol
@@ -2,7 +2,6 @@
 
 pragma solidity ^0.8.27;
 
-import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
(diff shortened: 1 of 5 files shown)


Amxx:
Was that requested ?

In general, I can see the point of having flexibility in the access control mechanism ... but I'm not sure if this contract really needs it. You mention this could be setup through the wizard, but the wizard is for contracts that app developpers will deploy, and for which we need many instances with different characteristics. In this case I only expect one instance per chain to exist. IMO it should be deployed, and owned, by axelar. The ownership would be an axelar multisig, or governor, and that is it.

ernestognw:
Not that it was requested but it's just too strict. Even if it's owned by Axelar it's too restrictive that we're only allowing Ownable.,

Also note that the gateways are adapters, it'd make sense if it was the canonical gateway. I think this change is not really controversial and just allows for better flexibility, which is what we want for a library in the end

PR #65, opened 2025-01-10
Run `forge update` and use Calldata library

ernestognw:


ernestognw:
Merging as a follow up of https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5422

PR #71, opened 2025-01-22
Add PaymasterCore and PaymasterSigner

ernestognw:


contracts/account/paymaster/PaymasterCore.sol
```diff
@@ -0,0 +1,133 @@
...
+    ) public virtual onlyEntryPoint {
+        _postOp(mode, context, actualGasCost, actualUserOpFeePerGas);
+    }
+
+    /**
+     * @dev Handles post user operation execution logic. The caller must be the entry point.
+     *
+     * It receives the `context` returned by `_validatePaymasterUserOp`. Reverts by default
+     * since the function is not called if no context is returned by {validatePaymasterUserOp}.
+     * Overriding without calling `super._postOp(...)` is advised.
```
Amxx, line 76: I'd like to inderstand why reverting is necessary. My initial idea was that this would just be a no-op, ignoring any unexpected context.

Are there any security concern with doing nothing here? I'm concerned that advising to not call super would cause Paymaster extensions to become incompatible.

contracts/account/paymaster/PaymasterCore.sol
```diff
@@ -0,0 +1,133 @@
...
+
+    /// @dev Calls {IEntryPointStake-addStake}.
+    function _addStake(uint256 value, uint32 unstakeDelaySec) internal virtual {
+        entryPoint().addStake{value: value}(unstakeDelaySec);
+    }
+
+    /// @dev Variant of {_addStake} that uses `msg.value` as the stake amount.
+    function _addStake(uint32 unstakeDelaySec) internal virtual {
+        _addStake(msg.value, unstakeDelaySec);
+    }
```
Amxx, line 114: IMO, we shouldn't have that variant. calling the other version with `msg.value` is super easy, and it is clearer/more explicit.

contracts/account/paymaster/PaymasterSigner.sol
```diff
@@ -0,0 +1,92 @@
...
+     * override {_postOp} to process the context passed along.
+     */
+    function _validatePaymasterUserOp(
+        PackedUserOperation calldata userOp,
+        bytes32 /* userOpHash */,
+        uint256 /* maxCost */
+    ) internal virtual override returns (bytes memory context, uint256 validationData) {
+        (uint48 validUntil, uint48 validAfter, bytes calldata signature) = decodePaymasterUserOp(userOp);
+        return (
+            context,
```
Amxx, line 77: your returning the return variable. At no point this is really defined, other than the implicit initialization. IMO that is not good practice
```suggestion
            bytes(""),
```

contracts/mocks/account/paymaster/PaymasterSignerECDSAMock.sol
```diff
@@ -0,0 +1,26 @@
...
+
+    // WARNING: No access control
+    function deposit() external payable {
+        _deposit();
+    }
+
+    // WARNING: No access control
+    function addStake(uint32 unstakeDelaySec) external payable {
+        _addStake(unstakeDelaySec);
+    }
```
Amxx, line 25: In a mock this is not necessary, the internal function are already available thanks to the hardhat-exposed plugin.

contracts/account/paymaster/PaymasterSigner.sol
```diff
@@ -0,0 +1,92 @@
...
+ *     }
+ * }
+ * ```
+ */
+abstract contract PaymasterSigner is PaymasterCore, AbstractSigner, EIP712 {
+    using ERC4337Utils for *;
+
+    bytes32 internal constant _USER_OPERATION_REQUEST =
+        keccak256(
+            "UserOperationRequest(address sender,uint256 nonce,bytes initCode,bytes callData,bytes32 accountGasLimits,uint256 paymasterVerificationGasLimit,uint256 preVerificationGas,bytes32 gasFees,uint48 validAfter,uint48 validUntil)"
```
Amxx, line 30: I'd reorder it so that it matches more closely the `PackedUserOperation` in `Account`
```suggestion
            "UserOperationRequest(address sender,uint256 nonce,bytes initCode,bytes callData,bytes32 accountGasLimits,uint256 preVerificationGas,bytes32 gasFees,uint256 paymasterVerificationGasLimit,uint48 validAfter,uint48 validUntil)"
```

contracts/account/paymaster/PaymasterCore.sol
ernestognw: The thing is that the entrypoint will call `postOp` when there's context but the user op won't revert if the `postOp` reverted. We have 2 options:

1) Revert by default, users who want to implement `validateUserOp` won't have an issue unless they really want to implement `postOp`, where overriding without calling super is ok.
2) Leave `_postOp` virtual, requiring that everyone implements it without knowing that it can just revert and it will be processed by the entrypoint regardless. I see very tempting that users will leave the function empty, which I think is even worse because in those cases they're indeed not calling `super`

So my reasoning is that 1) is safer. wdyt?

contracts/account/paymaster/PaymasterSigner.sol
```diff
@@ -0,0 +1,92 @@
...
+        );
+    }
+
+    /// @dev Decodes the user operation's data from `paymasterAndData`.
+    function decodePaymasterUserOp(
+        PackedUserOperation calldata userOp
+    ) public pure virtual returns (uint48 validAfter, uint48 validUntil, bytes calldata signature) {
+        bytes calldata paymasterData = userOp.paymasterData();
+        return (uint48(bytes6(paymasterData[0:6])), uint48(bytes6(paymasterData[6:12])), paymasterData[12:]);
+    }
```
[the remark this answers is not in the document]
ernestognw: No it's not required. I thought the paymaster should expose a way to understand the `paymasterData`. There's no standard for that and if we leave it internal I would add to the documentation how it's constructed.

Either keeping this public or having documentation is fine

contracts/mocks/account/paymaster/PaymasterSignerECDSAMock.sol
ernestognw: Yes, but they're not exposed as payable, so trying to call `$_addStake` and `$_deposit` will fail

contracts/account/paymaster/PaymasterSigner.sol
ernestognw: > the paymasterPostOpGasLimit, that is missing.

Yes, I don't see how a user can leverage this field to do any harm. The gas usage is in control of the paymaster itself.

If we decide to add it, I would include it in the `UserOperationRequest` type definition instead of the second option with `paymasterAndDataWithoutSig`. I find that last one problematic if anyone wants to customize the format of the `paymasterData` further.

contracts/account/paymaster/PaymasterCore.sol
Amxx: What about:

3. do nothing by default. Users can do whatever they want. Returning an empty context would mean the `postOp` is not called, which is ok. Returning a non empty context would mean the `postOp` is called, and does nothing, whuch is also ok. Users that want to work with a context can return it ... and override `_postOp`. When doing so, they can call super (recommandend good practice when overriding) and that would not revert.

Amxx: [A bit like the `_afterTokenTransfer` hook in v4](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.9.6/contracts/token/ERC20/ERC20.sol#L350-L364)

contracts/account/paymaster/PaymasterSigner.sol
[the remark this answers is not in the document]
Amxx: IMO this should be internal.

contracts/mocks/account/paymaster/PaymasterSignerECDSAMock.sol
Amxx: hooo, that is interresting. Maybe we should change that about the hardhat-exposed plugin

Amxx: please checkout https://github.com/frangio/hardhat-exposed/pull/38

contracts/account/paymaster/PaymasterSigner.sol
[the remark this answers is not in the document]
ernestognw: Sounds good, made it internal

contracts/account/paymaster/PaymasterCore.sol
ernestognw: I felt 3 would be dangerous too. If the linearization reaches the last `_postOp` that doesn't call super, it may be missing another override at the same hierarchic level. (e.g. `contract A is B, PaymasterCore {}` where B implements `_postOp`. I think B would be skipped).

I'm keeping it empty for now, I don't think such situation is common

contracts/mocks/account/paymaster/PaymasterSignerECDSAMock.sol
ernestognw: Looks good, is there a side effect of making functions payable? I wonder if it could break something on the vanilla repository

contracts/account/paymaster/PaymasterCore.sol
Amxx: In option 3, if one of the override doesn't call super, then you are definitelly "losing" some logic (just like the transfer hooks) ... but it would be clearly recommanded that you should call super.

In option 1, there is no super call, so you are certain that logic is dropped is you don't manually linearise the calls. IMO that is the most tricky/error prone approach.

contracts/mocks/account/paymaster/PaymasterSignerECDSAMock.sol
Amxx: My guess is that making payable just removes a check that msg.value is 0. Adding a check should not change any behavior. All the calls that are performed without value should not be affected.

contracts/account/paymaster/PaymasterSigner.sol
Amxx: >Would you think it's fine if we include paymasterPostOpGasLimit in the signature hash?

Its less generic, but yes that would be equivalent to "sign everything"

contracts/account/paymaster/PaymasterSigner.sol
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * @dev Extension of {PaymasterCore} that adds signature validation. See {SignerECDSA}, {SignerP256} or {SignerRSA}.
+ *
+ * Example of usage:
+ *
+ * ```solidity
+ * contract MyPaymasterECDSASigner is PaymasterSigner, SignerECDSA {
+ *     constructor() EIP712("MyPaymasterECDSASigner", "1") {
+ *       // Will revert if the signer is already initialized
+ *       _initializeSigner(signerAddr);
```
Amxx, line 21:
```suggestion
 *       _setSigner(signerAddr);
```

contracts/account/paymaster/PaymasterCore.sol
```diff
@@ -0,0 +1,124 @@
...
+    /// @dev Calls {IEntryPointStake-depositTo}.
+    function _deposit(address to, uint256 value) internal virtual {
+        entryPoint().depositTo{value: value}(to);
+    }
+
+    /// @dev Variant of {_deposit} that deposits `msg.value` to the paymaster itself (i.e. `address(this)`).
+    function _deposit() internal virtual {
+        _deposit(address(this), msg.value);
+    }
+
```
Amxx, line 96:
```suggestion
```

contracts/account/paymaster/PaymasterCore.sol
```diff
@@ -0,0 +1,120 @@
...
+ * through the internal functions {_deposit}, {_withdraw}, {_addStake}, {_unlockStake} and {_withdrawStake}.
+ *
+ * * Deposits are used to pay for user operations.
+ * * Stakes are used to guarantee the paymaster's reputation and obtain more flexibility in accessing storage.
+ *
+ * NOTE: See [Paymaster's unstaked reputation rules](https://eips.ethereum.org/EIPS/eip-7562#unstaked-paymasters-reputation-rules)
+ * for more details on the paymaster's storage access limitations.
+ */
+abstract contract PaymasterCore is IPaymaster {
+    /// @dev Unauthorized call to the account.
```
Amxx, line 25: IMO this error (and the equivalent in AccountCore) should be renamed to something like `EntryPointRestricted(address caller)`

```suggestion
    /// @dev Unauthorized call to the paymaster.
```

contracts/mocks/docs/account/paymaster/MyPaymaster.sol
```diff
@@ -0,0 +1,18 @@
...
+contract MyPaymaster is PaymasterCore {
+    /// @dev Paymaster user op validation logic
+    function _validatePaymasterUserOp(
+        PackedUserOperation calldata userOp,
+        bytes32 userOpHash,
+        uint256 requiredPreFund
+    ) internal override returns (bytes memory context, uint256 validationData) {
+        // Custom validation logic
+    }
+}
```
Amxx, line 18:
```suggestion

    // Staking & unstacking logic (permissioned)
}
```

contracts/mocks/docs/account/paymaster/MyPaymasterECDSA.sol
```diff
@@ -0,0 +1,14 @@
...
+
+import {PaymasterSigner, EIP712} from "../../../../account/paymaster/PaymasterSigner.sol";
+import {PackedUserOperation} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
+import {SignerECDSA} from "../../../../utils/cryptography/SignerECDSA.sol";
+
+contract MyPaymasterECDSA is PaymasterSigner, SignerECDSA {
+    constructor(address paymasterSignerAddr) EIP712("MyAccountECDSA", "1") {
+        _setSigner(paymasterSignerAddr);
+    }
+}
```
Amxx, line 14:
```suggestion

    // Staking & unstacking logic (permissioned)
}
```

ernestognw: he _authorizeWithdraw should make it clearer, do you agree?

test/account/paymaster/Paymaster.behavior.js
```diff
@@ -0,0 +1,96 @@
...
+      const userOp = { ...this.userOp };
+      userOp.callData = this.accountMock.interface.encodeFunctionData('execute', [
+        encodeMode({ callType: CALL_TYPE_BATCH }),
+        encodeBatch([
+          {
+            target: this.target.target,
+            data: this.target.interface.encodeFunctionData('mockFunction'),
+          },
+        ]),
+      ]);
```
Amxx, line 32: This creates a copy of `this.userOp`, and sets the callData ... but then that is never used.

Bellow, operation is produced using `this.userOp` that was not touched.

test/account/paymaster/Paymaster.behavior.js
```diff
@@ -0,0 +1,96 @@
...
+      const operation = await this.accountMock.createUserOp(this.userOp);
+      const userSignedUserOp = await this.signUserOp(operation);
+      const paymasterSignedUserOp = await this.paymasterSignUserOp(userSignedUserOp, 0, 0);
```
Amxx, line 35: Shouldn't the paymaster sign first, so that the paymaster signature can be placed in the userOp, and then the user signs that?

test/account/paymaster/Paymaster.behavior.js
```diff
@@ -0,0 +1,96 @@
...
+
+      await expect(entrypoint.balanceOf(this.mock)).to.eventually.eq(this.deposit);
+      const handleOpsTx = await entrypoint.handleOps([paymasterSignedUserOp.packed], this.receiver);
+      await expect(entrypoint.balanceOf(this.mock)).to.eventually.be.lessThan(this.deposit);
+      expect(handleOpsTx).to.emit(this.target, 'MockFunctionCalledExtra');
+      expect(handleOpsTx).to.not.changeEtherBalance(this.accountMock);
```
Amxx, line 41: await are missing here. When adding them the tests fail.

test/account/paymaster/Paymaster.behavior.js
```diff
@@ -0,0 +1,96 @@
...
+    it('reverts if the caller is not the entrypoint', async function () {
+      await expect(this.mock.connect(this.other).postOp(0, '0x', 0, 0)).to.be.reverted;
+    });
+  });
+
+  describe('deposit lifecycle', function () {
+    it('deposits and withdraws effectively', async function () {
+      const value = 100n;
+      const depositTx = await this.mock.connect(this.depositor).deposit({ value });
+      expect(depositTx).to.changeEtherBalance([this.depositor, entrypoint], [value * -1n, value]);
```
Amxx, line 65: invalid syntax + not awaited

Amxx (approved):

PR #72, opened 2025-01-28
Allow key updates in Signers

Amxx:
- Change the internal setters so that they can be reused if someone wants to update the signer's key
- in SignerP256, check that the provided key is valid.


ernestognw:
Overall I agree with the change but I always felt it was a footgun and developers must consciously introduce the ability to set the key again. Especially since accounts are initializable, I feel it's very easy someone leaves reinitialization exposed (e.g. setting the key again)

ernestognw (dismissed):
Nits, but LGTM

PR #75, opened 2025-02-01
Add PaymasterNFT and PaymasterERC20

Amxx:
Includes a slight refactor of the behavior tests


Amxx:
Note, used this PR to update missing entries in the Changelog (we should have a CI check for that)

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## XX-XX-XXXX
+
+- `PaymasterNFT`: Extension of `PaymasterCore` that approves sponsoring of user operation based on ownership of an ERC-721 NFT.
+- `PaymasterERC20`: Extension of `PaymasterCore` that sponsores transaction against payment in ERC-20 tokens.
```
Amxx, line 4:
```suggestion
- `PaymasterERC20`: Extension of `PaymasterCore` that sponsors user operations against payment in ERC-20 tokens.
```

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,101 @@
...
+        bytes calldata paymasterData
+    )
+        internal
+        view
+        virtual
+        returns (IERC20 token, uint48 validAfter, uint48 validUntil, uint256 tokenPrice, address guarantor);
+
+    function _tokenPriceDenominator() internal view virtual returns (uint256) {
+        return 1e6;
+    }
```
arr00, line 95: Worth documenting this?

Amxx: Yes, this is missing some documentation!

One thing we could consider is to 
- remove that function
- put the denominator in the fetch details 
- pass it in the context.

arr00: Could we not just require the necessary scaling to be done in the `_fetchDetails` function? Meaning, just always have a denominator of 1e18.

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,115 @@
...
+        address indexed user,
+        address indexed guarantor,
+        uint256 tokenAmount,
+        uint256 tokenPrice,
+        bool paidByGuarantor
+    );
+
+    // Over-estimations: ERC-20 balances/allowances may be cold and contracts may not be optimized
+    uint256 private constant POST_OP_COST = 30_000;
+    uint256 private constant POST_OP_COST_WITH_GUARANTOR = 50_000;
```
ernestognw, line 31: I think it's worth specifying how these are calculated. The values seems fine but if I'm a developer using the library, the first thing I'd ask myself is what happens if the underlying ERC-20 consumes even more gas than this over-estimation.

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,115 @@
...
+        );
+
+        uint256 prefundAmount = (maxCost +
+            (guarantor == address(0)).ternary(POST_OP_COST, POST_OP_COST_WITH_GUARANTOR) *
+            userOp.maxFeePerGas()).mulDiv(tokenPrice, _tokenPriceDenominator());
+
+        return (
+            abi.encodePacked(userOpHash, token, prefundAmount, tokenPrice, userOp.sender, guarantor),
+            token
+                .trySafeTransferFrom(guarantor == address(0) ? userOp.sender : guarantor, address(this), prefundAmount)
```
ernestognw, line 51: One thing I'm worried about the guarantor is that if they approve this contract, then anyone can use them as guarantor. 

I guess it depends on the implementation of `_fetchDetails`. For example, if the guarantor is extracted from the `paymasterAndData` field, then each operation must validate a signature from the guarantor.

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * @dev Internal function that returns the repayment details for a given user operation
+     *
+     * This may be implemented in any number of ways, including
+     * * Hardcoding values (only one token supported)
+     * * Getting the price from an onchain oracle
+     * * Getting the (signed) values through the userOp's paymasterData
+     *
+     * The paymaster can also decide to not support guarantors, and always return address(0) for that part.
```
ernestognw, line 94: Following my previous comment, I think this makes it clearer

```suggestion
     * The paymaster can also decide to not support guarantors, and always return address(0) for that part.
     *
     * NOTE: If a guarantor is supported, make sure that it can't be used arbitrarily to pay operations.
     * Concretely, if the guarantor is extracted from the `userOp`, then it'll be possible to use any
     * address who's approved this contract
```

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,115 @@
...
+     */
+    function _fetchDetails(
+        PackedUserOperation calldata userOp,
+        bytes32 userOpHash
+    )
+        internal
+        view
+        virtual
+        returns (IERC20 token, uint48 validAfter, uint48 validUntil, uint256 tokenPrice, address guarantor);
```
ernestognw, line 103: I understand the price is denominated in ETH, wouldn't be clearer to label it as `tokenPriceInNative` to signal the prices is in native currency. Not a fan of the name though

```suggestion
        returns (IERC20 token, uint48 validAfter, uint48 validUntil, uint256 tokenPriceInNative, address guarantor);
```

contracts/account/paymaster/PaymasterERC20.sol
```diff
@@ -0,0 +1,115 @@
...
+
+    /// @dev Denominator used for interpreting the `tokenPrice` returned by {_fetchDetails} as "fixed point".
+    function _tokenPriceDenominator() internal view virtual returns (uint256) {
+        return 1e18;
```
ernestognw, line 107: Isn't denominator too high? For example, if the `maxFeePerGas` is 1e9, then any price less than 1e9 will be zeroed out.

Am I missing something?

contracts/account/paymaster/PaymasterERC20.sol
Amxx: It was estimated by "looking" at gas cost, and its relevance is "checked" in the unit tests:

```javascript
expect(tokenAmount)
  .to.be.greaterThan(actualGasCost * 2n)
  .to.be.lessThan((actualGasCost * 2n * 110n) / 100n);
```

If the underlying ERC20 conumes even more gas, then the result is that the cost of the PostOp, which is beared (in Eth) by the paymaster would not be fully refunded (in ERC20, by the account, to the paymaster). So the paymaster may lose a bit of money (or make less than expected). 

That is if the paymaster is accepting this token as payment (its the paymaster's decision to decide what ERC20 they accept)

contracts/account/paymaster/PaymasterERC20.sol
Amxx: How would you fell about: 

```suggestion
     * The paymaster can also decide to not support guarantors, and always return address(0) for that part.
     *
     * NOTE: If a guarantor is supported, make sure that it can't be used arbitrarily to pay operations.
     * Concretely, if the guarantor is extracted from the `userOp`, make sure that it provided explicit consent to 
     * support that user operation, for example by verifying a signature.
```

contracts/account/paymaster/PaymasterERC20.sol
Amxx: It used to be 1e6, then @arr00 said it was too low and 1e18 would be better. This is virtual, so devs can override it.

Note that this is used in `Math.mulDiv` to "scale" the token price (in native). So you could put anything here, it would just re-scale the token price. I'm not sure how `maxFeePerGas` has an impact here.

contracts/account/paymaster/PaymasterERC20.sol
Amxx: what about keeping the name as it is, but document that better in the natspec ?

contracts/account/paymaster/PaymasterNFT.sol
```diff
@@ -0,0 +1,49 @@
...
+        _token = token_;
+        emit PaymasterNFTTokenSet(token_);
+    }
+
+    /**
+     * @dev Internal validation of whether the paymaster is willing to pay for the user operation.
+     * Returns the context to be passed to postOp and the validation data.
+     *
+     * NOTE: The `context` returned is `bytes(0)`. Developers overriding this function MUST
+     * override {_postOp} to process the context passed along.
```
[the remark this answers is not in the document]
Amxx:
```suggestion
     * NOTE: The default `context` is `bytes(0)`. Developers that add a context when overriding this function MUST
     * also override {_postOp} to process the context passed along.
```

ernestognw (commented):
I feel the PaymasterERC20 should be split into two contracts. One without guarantor and another that adds it as an extension.

Basically, the PaymasterERC20 might just be:

```solidity
abstract contract PaymasterERC20 is PaymasterCore {
    using ERC4337Utils for *;
    using Math for *;
    using SafeERC20 for IERC20;

    event UserOperationSponsored(
        bytes32 indexed userOpHash,
        address indexed funder,
        uint256 tokenAmount,
        uint256 tokenPrice
    );

    /// @inheritdoc PaymasterCore
    function _validatePaymasterUserOp(
        bytes32 userOpHash,
        uint256 maxCost
    ) internal virtual override returns (bytes memory context, uint256 validationData) {
        (uint256 validationData_, IERC20 token, uint256 tokenPrice) = _fetchPrice(userOp);

        uint256 prefundAmount = _ERC20Cost(maxCost, userOp.maxFeePerGas(), tokenPrice);
        address funder = _funder(userOp);

        // if validation is obviously failed, don't even try to do the ERC-20 transfer
        return
            (validationData_ != ERC4337Utils.SIG_VALIDATION_FAILED &&
                token.trySafeTransferFrom(funder, address(this), prefundAmount))
                ? (abi.encodePacked(userOpHash, token, prefundAmount, tokenPrice, funder), validationData_)
                : (bytes(""), ERC4337Utils.SIG_VALIDATION_FAILED);
    }

    /// @inheritdoc PaymasterCore
    function _postOp(
        PostOpMode /* mode */,
        bytes calldata context,
        uint256 actualGasCost,
        uint256 actualUserOpFeePerGas
    ) internal virtual override {
        bytes32 userOpHash = bytes32(context[0x00:0x20]);
        IERC20 token = IERC20(address(bytes20(context[0x20:0x34])));
        uint256 prefundAmount = uint256(bytes32(context[0x34:0x54]));
        uint256 tokenPrice = uint256(bytes32(context[0x54:0x74]));
        address funder = address(bytes20(context[0x74:0x88]));

        uint256 actualAmount = _ERC20Cost(actualGasCost, actualUserOpFeePerGas, tokenPrice);

        token.safeTransfer(funder, prefundAmount - actualAmount);
        emit UserOperationSponsored(userOpHash, funder, actualAmount, tokenPrice);
    }

    function _fetchPrice(
        PackedUserOperation calldata userOp
    ) internal view virtual returns (uint256 validationData, IERC20 token, uint256 tokenPrice);

    function _funder(PackedUserOperation calldata userOp) internal view virtual returns (address) {
        return userOp.sender;
    }

    /// @dev Denominator used for interpreting the `tokenPrice` returned by {_fetchOracleDetails} as "fixed point".
    function _tokenPriceDenominator() internal view virtual returns (uint256) {
        return 1e18;
    }

    function _ERC20Cost(uint256 cost, uint256 feePerGas, uint256 tokenPrice) internal view virtual returns (uint256) {
        return (cost + _postOpCost() * feePerGas).mulDiv(tokenPrice, _tokenPriceDenominator());
    }

    function _postOpCost() internal view virtual returns (uint256) {
        // Over-estimations: ERC-20 balances/allowances may be cold and contracts may not be optimized
        return 30_000;
    }

    /// @dev Public function that allows the withdrawer to extract ERC-20 tokens resulting from gas payments.
    function withdrawTokens(IERC20 token, address recipient, uint256 amount) public virtual onlyWithdrawer {
        if (amount == type(uint256).max) amount = token.balanceOf(address(this));
        token.safeTransfer(recipient, amount);
    }
}
```

While the `PaymasterERC20Guarantor` extension would be:

```solidity
abstract contract PaymasterERC20Guarantor is PaymasterERC20 {
    event UserOperationGuaranteed(bytes32 indexed userOpHash, address indexed user, address indexed guarantor);

    function _funder(PackedUserOperation calldata userOp) internal view virtual override returns (address) {
        (uint256 validationData, address guarantor) = _fetchGuarantor(userOp);
        return
            (validationData == ERC4337Utils.SIG_VALIDATION_SUCCESS && guarantor != address(0))
                ? userOp.sender
                : guarantor;
    }

    function _postOpCost() internal view virtual override returns (uint256) {
        return super._postOpCost() + 15_000;
    }

    function _fetchGuarantor(
        PackedUserOperation calldata userOp
    ) internal view virtual returns (uint256 validationData, address guarantor);
}
```

wdyt @Amxx ?
(NOTE: May not be an accurate example)

contracts/account/paymaster/PaymasterNFT.sol
```diff
@@ -0,0 +1,49 @@
...
+pragma solidity ^0.8.20;
+
+import {IERC721} from "@openzeppelin/contracts/interfaces/IERC721.sol";
+import {ERC4337Utils, PackedUserOperation} from "@openzeppelin/contracts/account/utils/draft-ERC4337Utils.sol";
+import {PaymasterCore} from "./PaymasterCore.sol";
+
+/**
+ * @dev Extension of {PaymasterCore} that supports account based on ownership of an ERC-721 token
+ */
+abstract contract PaymasterNFT is PaymasterCore {
```
ernestognw, line 12: I think `PaymasterNFT` was a mistake, and we should've called it `PaymasterERC721`, did we discussed it? It's just that now it seems pretty arbitrary when reviewing 😅

contracts/account/paymaster/PaymasterNFT.sol
```diff
@@ -0,0 +1,49 @@
...
+    IERC721 private _token;
+
+    event PaymasterNFTTokenSet(IERC721 token);
```
ernestognw, line 15: Accordingly

```suggestion
    event PaymasterERC721TokenSet(IERC721 token);
```

contracts/account/paymaster/PaymasterNFT.sol
```diff
@@ -0,0 +1,49 @@
...
+        PackedUserOperation calldata userOp,
+        bytes32 /* userOpHash */,
+        uint256 /* maxCost */
+    ) internal virtual override returns (bytes memory context, uint256 validationData) {
+        return (
+            bytes(""),
+            token().balanceOf(userOp.sender) == 0
+                ? ERC4337Utils.SIG_VALIDATION_FAILED
+                : ERC4337Utils.SIG_VALIDATION_SUCCESS
+        );
```
ernestognw, line 47: `balanceOf` reverts if the `userOp.sender` is `address(0)`, however, this is unreachable with `address(0)` if the entrypoint is the canonical (or ERC-4337 compliant). I think it's worth a note

```suggestion
            // balanceOf reverts if the `userOp.sender` is the address(0), so this becomes unreachable with address(0)
            // assuming a compliant entrypoint (`_validatePaymasterUserOp` is called after `validateUserOp`),
            token().balanceOf(userOp.sender) == 0
                ? ERC4337Utils.SIG_VALIDATION_FAILED
                : ERC4337Utils.SIG_VALIDATION_SUCCESS
        );
```

contracts/mocks/ERC20Mock.sol
ernestognw, line 1: Can't we use hardhat-exposed's `include` in settings? Similar to:

```js
module.exports = {
  ...,
  exposed: {
    include: ['./lib/@openzeppelin-contracts/contracts/token/ERC20/ERC20.sol'],
  },
};
```

This way we can get rid of a couple of mocks

contracts/account/paymaster/PaymasterNFT.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {IERC721} from "@openzeppelin/contracts/interfaces/IERC721.sol";
+import {ERC4337Utils, PackedUserOperation} from "@openzeppelin/contracts/account/utils/draft-ERC4337Utils.sol";
+import {PaymasterCore} from "./PaymasterCore.sol";
+
+/**
+ * @dev Extension of {PaymasterCore} that supports account based on ownership of an ERC-721 token
```
ernestognw, line 10: imo, it should be made clear that the mechanism is ownership of at least one token from an ERC-721 collection:

```suggestion
 * @dev Extension of {PaymasterCore} that supports account based on ownership of an ERC-721 token.
 *
 * This paymaster will sponsor user operations if the user has at least 1 token of the token specified
 * during construction (or via {_setToken}).
```

contracts/mocks/ERC20Mock.sol
ernestognw: Tried this but didn't work 😞, it's not blocking though

ernestognw (approved):

contracts/account/paymaster/PaymasterNFT.sol
Amxx: That was the name in the grant agreement.

I'd be ok with changing it, but I would go even further and call it something like `PaymasterERC721Owner` to clarify that the ownership gives you the right and you are not actually giving away the NFT as a payment.

contracts/mocks/ERC20Mock.sol
Amxx: Yes, I tried that as well. I think I tried to change the hardhat-expose plugin to support exposing things that are not in `contract` if they are imported, but we never did that.

Amxx:
Merging. and opening another issue to change PaymasterNFT's name

PR #76, opened 2025-02-03
Typehash: constant private with _TYPEHASH suffix

Amxx:


arr00 (approved):
Note that we have a combination of all different visibilities on the typehashes in the vanilla repo (public, internal, private).

Amxx:
Interresting. Then I guess this is more about the suffix than about making it private.

We should decide something.

ernestognw (approved):
My experience has been that typehashes are needed for testing. So generally mocks redeclare it (especially testing in Foundry). For those cases I would consider keeping it internal, LGTM though.

Amxx:
> My experience has been that typehashes are needed for testing. 

I'm curious about that part. (ethers) signers take a definition of the 712 structure and generate the typehash automatically.

I ran `grep -R TYPEHASH openzeppelin-contracts/test`, and the only thing I found was in `test/metatx/ERC2771Forwarder.t.sol`

PR #77, opened 2025-02-03
Add husky to devDependencies

ericglau:
package.json has `"prepare": "husky"`

Since this Community Contracts package is not yet published to NPM, consuming packages must install it through git.  According to [NPM's life cycle scripts](https://docs.npmjs.com/cli/v11/using-npm/scripts#life-cycle-scripts), the `prepare` script of this package must be run:
> NOTE: If a package being installed through git contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.

Therefore `husky` should be declared as a `devDependency` so that this `prepare` script can be run by consumers.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,8 @@
         "@axelar-network/axelar-gmp-sdk-solidity": "^5.10.0"
       },
       "devDependencies": {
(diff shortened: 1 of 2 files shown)


arr00 (approved):
Good catch

PR #80, opened 2025-02-10
Add ERC-6900 account extension

james-toussaint:
In progress works:
- [x] installValidation
- [ ] uninstallValidation
- [x] installExecution
- [ ] uninstallExecution
- [x] validateUserOp
- [x] rawSignatureValidation
- [x] execute
- [x] executeBatch
- [x] executeWithRuntimeValidation
- [x] fallback

### DIFF
--- a/contracts/account/extensions/AccountERC6900.sol
+++ b/contracts/account/extensions/AccountERC6900.sol
@@ -51,7 +51,6 @@ abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount
     mapping(ModuleEntity moduleEntity => Validation) private _validations;
     mapping(bytes4 selector => Execution) private _executions;
     mapping(bytes4 interfaceId => bool supported) private _interfaceIds;
-    mapping(bytes4 selector => address) private _fallbacks;
 
     error ERC6900ModuleInterfaceNotSupported(address module, bytes4 expectedInterface);
     error ERC6900AlreadySetSelectorForValidation(ModuleEntity validationFunction, bytes4 selector);
@@ -69,6 +68,7 @@ abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount
         bytes4 executionSelector
     );
     error ERC6900InvalidExecuteTarget();
+    error ERC6900MissingFallbackExecutionModule(address module);
 
     struct Validation {
         EnumerableSet.Bytes32Set selectors;
@@ -409,16 +409,19 @@ abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount
     }
 
     /**
-     * @dev Fallback function that delegates the call to the installed handler for the given selector.
+     * @dev Fallback function that delegates the call to the installed execution module for the given selector.
      *
      */
     function _fallback() internal virtual returns (bytes memory) {
-        // TODO
+        address module = _fallbackExecutionModule(msg.sig);
+        require(module != address(0), ERC6900MissingFallbackExecutionModule(module));
+        (bool success, bytes memory returndata) = module.call{value: msg.value}(msg.data);
+        return Address.verifyCallResult(success, returndata);
     }
 
-    /// @dev Returns the fallback handler for the given selector. Returns `address(0)` if not installed.
-    function _fallbackHandler(bytes4 selector) internal view virtual returns (address) {
-        return _fallbacks[selector];
+    /// @dev Returns the fallback execution module for the given selector. Returns `address(0)` if not installed.
+    function _fallbackExecutionModule(bytes4 selector) internal view virtual returns (address) {
+        return _executions[selector].module;
     }
 
     function _decodeValidationConfig(
@@ -453,21 +456,6 @@ abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount
         bytes calldata signature
     ) internal pure virtual returns (ModuleEntity moduleEntity, bytes calldata innerSignature) {
         return (ModuleEntity.wrap(bytes24(signature[0:24])), signature[24:]);
-    }
-
-    /**
-     * @dev Extract the function selector from initData/deInitData for MODULE_TYPE_FALLBACK
-     *
-     * NOTE: If we had calldata here, we would could use calldata slice which are cheaper to manipulate and don't
-     * require actual copy. However, this would require `_installModule` to get a calldata bytes object instead of a
-     * memory bytes object. This would prevent calling `_installModule` from a contract constructor and would force
-     * the use of external initializers. That may change in the future, as most accounts will probably be deployed as
-     * clones/proxy/ERC-7702 delegates and therefore rely on initializers anyway.
-     */
-    function _decodeFallbackData(
-        bytes memory data
-    ) internal pure virtual returns (bytes4 selector, bytes memory remaining) {
-        return (bytes4(data), data.slice(4));
     }
 
     // TODO: Remove flat function and update test
--- a/contracts/account/extensions/AccountERC6900.sol
+++ b/contracts/account/extensions/AccountERC6900.sol
@@ -1,0 +1,494 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.24;
+
+import {PackedUserOperation, IAccount} from "@openzeppelin/contracts/interfaces/draft-IERC4337.sol";
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
+import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
+import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
+import {Packing} from "@openzeppelin/contracts/utils/Packing.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
+import {Calldata} from "@openzeppelin/contracts/utils/Calldata.sol";
+import {ERC165Checker} from "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
+import {IERC6900ModularAccount, IERC6900Module, IERC6900ValidationModule, IERC6900ExecutionModule, ValidationConfig, ModuleEntity, ValidationFlags, HookConfig, ExecutionManifest, ManifestExecutionHook, ManifestExecutionFunction, Call} from "../../interfaces/IERC6900.sol";
+import {ERC6900Utils} from "../utils/ERC6900Utils.sol";
+import {ERC7739} from "../../utils/cryptography/ERC7739.sol";
+import {AccountCore} from "../AccountCore.sol";
+
+/**
+ * @dev Extension of {AccountCore} that implements support for ERC-6900 modules.
+ *
+ * To comply with the ERC-1271 support requirement, this contract implements {ERC7739} as an
+ * opinionated layer to avoid signature replayability across accounts controlled by the same key.
+ *
+ * This contract does not implement validation logic for user operations since these functionality
+ * is often delegated to self-contained validation modules. Developers must install a validator module
+ * upon initialization (or any other mechanism to enable execution from the account):
+ *
+ * ```solidity
+ * contract MyAccountERC6900is AccountERC6900, Initializable {
+ *     constructor() EIP712("MyAccount", "1") {}
+ *
+ *   function installValidation(
+ *       ValidationConfig validationConfig,
+ *       bytes4[] calldata selectors,
+ *       bytes calldata installData,
+ *       bytes[] calldata hooks
+ *   ) public initializer {
+ *     _installValidation(validationConfig, selectors, installData, hooks);
+ *   }
+ * }
+ * ```
+ *
+ */
+
+abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount {
+    using Bytes for *;
+    using ERC6900Utils for *;
+    using ERC6900Utils for ModuleEntity;
+    using EnumerableSet for *;
+    using Packing for bytes32;
+
+    mapping(ModuleEntity moduleEntity => Validation) private _validations;
+    mapping(bytes4 selector => Execution) private _executions;
+    mapping(bytes4 interfaceId => bool supported) private _interfaceIds;
+    mapping(bytes4 selector => address) private _fallbacks;
+
+    struct Validation {
+        EnumerableSet.Bytes32Set selectors;
+        ValidationFlags validationFlags;
+        EnumerableSet.Bytes32Set validationHooks;
+        EnumerableSet.Bytes32Set executionHooks;
+    }
+
+    struct Execution {
+        address module;
+        bool skipRuntimeValidation;
+        bool allowGlobalValidation;
+        EnumerableSet.Bytes32Set executionHooks;
+    }
+
+    /// @dev See {_fallback}.
+    fallback(bytes calldata) external payable virtual returns (bytes memory) {
+        return _fallback();
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function accountId() public view virtual returns (string memory) {
+        // vendorname.accountname.semver
+        return "@openzeppelin/community-contracts.AccountERC6900.v0.0.0";
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function installValidation(
+        ValidationConfig validationConfig,
+        bytes4[] calldata selectors,
+        bytes calldata installData,
+        bytes[] calldata hooks
+    ) public virtual onlyEntryPointOrSelf {
+        _installValidation(validationConfig, selectors, installData, hooks);
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function uninstallValidation(
+        ModuleEntity validationFunction,
+        bytes calldata uninstallData,
+        bytes[] calldata hookUninstallData
+    ) public virtual onlyEntryPointOrSelf {
+        // _uninstallModule(moduleTypeId, module, deInitData);
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function installExecution(
+        address module,
+        ExecutionManifest memory manifest,
+        bytes calldata installData
+    ) public virtual onlyEntryPointOrSelf {
+        _installExecution(module, manifest, installData);
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function uninstallExecution(
+        address module,
+        ExecutionManifest calldata manifest,
+        bytes calldata uninstallData
+    ) public virtual onlyEntryPointOrSelf {
+        // _uninstallModule(moduleTypeId, module, deInitData);
+    }
+
+    /// @dev Executes a transaction from the entry point or the account itself. See {_execute}.
+    function execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) public payable virtual onlyEntryPointOrSelf returns (bytes memory) {
+        return _execute(target, value, data);
+    }
+
+    function executeBatch(Call[] calldata calls) public payable virtual onlyEntryPointOrSelf returns (bytes[] memory) {
+        bytes[] memory returnedData = new bytes[](calls.length);
+        for (uint256 i = 0; i < calls.length; i++) {
+            returnedData[i] = _execute(calls[i].target, calls[i].value, calls[i].data);
+        }
+        return returnedData;
+    }
+
+    function executeWithRuntimeValidation(
+        bytes calldata data,
+        bytes calldata authorization
+    ) public payable virtual onlyEntryPointOrSelf returns (bytes memory) {
+        ModuleEntity validationModuleEntity = ModuleEntity.wrap(bytes24(authorization[:24]));
+        bytes calldata validationAuth = authorization[24:];
+        IERC6900ValidationModule(validationModuleEntity.module()).validateRuntime(
+            address(this),
+            validationModuleEntity.entityId(),
+            msg.sender,
+            msg.value,
+            data,
+            validationAuth
+        );
+        return _execute(address(this), msg.value, data);
+    }
+
+    /**
+     * @dev Validates a user operation with {_signableUserOpHash} and returns the validation data
+     * if the module specified by the first 20 bytes of the nonce key is installed. Falls back to
+     * {AccountCore-_validateUserOp} otherwise.
+     *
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -51,8 +51,7 @@
   },
   "lint-staged": {
     "*.{js,ts}": [
-      "prettier --log-level warn --ignore-path .gitignore --check",
-      "eslint"
+      "prettier --log-level warn --ignore-path .gitignore --check"
```
Amxx, line 54: Why that change, are there fundamental issues with eslint here ?

contracts/account/utils/ERC6900Utils.sol
```diff
@@ -0,0 +1,124 @@
...
+    //     return ModuleEntity.wrap(Packing.extract_32_24(ValidationConfig.unwrap(validationConfig), 0));
+    // }
+
+    // function getValidationFlags(ValidationConfig validationConfig) internal pure returns (ValidationFlags) {
+    //     return ValidationFlags.wrap(uint8(Packing.extract_32_1(ValidationConfig.unwrap(validationConfig), 21)));
+    // }
+
+    // HookConfig
+
+    function extractHookConfig(bytes calldata hook) internal pure returns (HookConfig) {
```
Amxx, line 96: It feels like `getHookData` and `extractHookConfig` are duplicated code. Can we get rid of one?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+ *   }
+ * }
+ * ```
+ *
+ */
+
+abstract contract AccountERC6900 is AccountCore, ERC7739, IERC6900ModularAccount {
+    using Bytes for *;
+    using ERC6900Utils for *;
+    using ERC6900Utils for ModuleEntity;
```
Amxx, line 48:
```suggestion
    using ERC6900Utils for *;
```

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+    /// @dev Executes a transaction from the entry point or the account itself. See {_execute}.
+    function execute(
+        address target,
+        uint256 value,
+        bytes calldata data
+    ) public payable virtual onlyEntryPointOrSelf returns (bytes memory) {
+        return _execute(target, value, data);
+    }
+
+    function executeBatch(Call[] calldata calls) public payable virtual onlyEntryPointOrSelf returns (bytes[] memory) {
```
Amxx, line 128:
```suggestion
    /// @inheritdoc IERC6900ModularAccount
    function executeBatch(Call[] calldata calls) public payable virtual onlyEntryPointOrSelf returns (bytes[] memory) {
```

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+        bytes[] memory returnedData = new bytes[](calls.length);
+        for (uint256 i = 0; i < calls.length; i++) {
+            returnedData[i] = _execute(calls[i].target, calls[i].value, calls[i].data);
+        }
+        return returnedData;
+    }
+
+    function executeWithRuntimeValidation(
```
Amxx, line 136:
```suggestion
    /// @inheritdoc IERC6900ModularAccount
    function executeWithRuntimeValidation(
```

contracts/interfaces/IERC6900.sol
```diff
@@ -0,0 +1,351 @@
...
+
+struct Call {
+    // The target address for the account to call.
+    address target;
+    // The value to send with the call.
+    uint256 value;
+    // The calldata for the call.
+    bytes data;
+}
+interface IERC6900ModularAccount {
```
Amxx, line 50:
```suggestion
}

interface IERC6900ModularAccount {
```

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+            uint32(Packing.extract_32_4(ValidationConfig.unwrap(validationConfig), 20)),
+            Packing.extract_32_1(ValidationConfig.unwrap(validationConfig), 21)
+        );
+    }
+
+    /**
+     * @dev Extracts the validator from the user operation.
+     *
+     */
+    function _extractUserOpValidator(PackedUserOperation calldata userOp) internal pure virtual returns (ModuleEntity) {
```
Amxx, line 429: `_extractSignatureValidator` already does that. Maybe we can drop this function, and only use `_extractSignatureValidator`

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+        address module;
+        bool skipRuntimeValidation;
+        bool allowGlobalValidation;
+        EnumerableSet.Bytes32Set executionHooks;
+    }
+
+    /// @dev See {_fallback}.
+    fallback(bytes calldata) external payable virtual returns (bytes memory) {
+        return _fallback();
+    }
```
Amxx, line 74: you should probably pass the input.
```suggestion
    fallback(bytes calldata data) external payable virtual returns (bytes memory) {
        return _fallback(data);
    }
```

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+    ) internal view virtual override returns (bool) {
+        if (signature.length < 25) return false;
+        (ModuleEntity validationModuleEntity, bytes calldata innerSignature) = _extractSignatureValidator(signature);
+        (address module, uint32 entityId) = validationModuleEntity.unpack();
+        // If the validation function is attempted to be used for signature validation
+        // and the flag isSignatureValidation is set to false, validation MUST revert
+        require(
+            !_validations[validationModuleEntity].validationFlags.isSignatureValidation(),
+            ERC6900Utils.ERC6900BadSignatureValidation(validationModuleEntity)
+        );
```
Amxx, line 231: Is it ok to revert here, or should it return `false` ?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+    /**
+     * @dev ERC-6900 execution logic.
+     *
+     * Reverts if the call type is not supported.
+     */
+    function _execute(address target, uint256 value, bytes calldata data) internal virtual returns (bytes memory) {
+        (bool success, bytes memory returndata) = target.call{value: value}(data);
+
+        require(success, ERC6900Utils.ERC6900ExecutionFailed(target, value, data));
+        return returndata;
```
Amxx, line 252: Is the `ERC6900Utils.ERC6900ExecutionFailed` standard ? Otherwise I think we should just do something that bubble the revert reason, for example using `Address.verifyCallResult`

@arr00 do you know what the ERC says ?

arr00: There are no errors specified in the ERC.

contracts/account/utils/ERC6900Utils.sol
```diff
@@ -0,0 +1,124 @@
...
+    error ERC6900ExecutionSelectorConflictingWithERC6900Function(address module, bytes4 selector);
+    error ERC6900BadUserOpValidation(ModuleEntity moduleEntity);
+    error ERC6900BadSignatureValidation(ModuleEntity moduleEntity);
+    error ERC6900MissingValidationForSelector(bytes4 selector);
+    error ERC6900ExecutionSelectorNotAllowedForGlobalValidation(
+        ModuleEntity validationModuleEntity,
+        bytes4 executionSelector
+    );
+    error ERC6900InvalidExecuteTarget();
+    error ERC6900ExecutionFailed(address target, uint256 value, bytes data);
```
Amxx, line 35: If the error are not emited in the utils library, we usually define them in the contract itself

contracts/account/extensions/AccountERC6900.sol
Amxx: so we should probably be doing
```suggestion
        (bool success, bytes memory returndata) = target.call{value: value}(data);
        return Address.verifyCallResult(success, returndata);
```

test/account/extensions/AccountERC6900.behavior.js
```diff
@@ -0,0 +1,148 @@
...
+        const moduleId = this.randomContract.target; // not a validation module
+        const installData = ethers.hexlify(ethers.randomBytes(256));
+        const entityId = ethers.hexlify('0x11223344');
+        const validationFlags = ethers.hexlify('0x11');
+        const validationConfig = ethers.concat([moduleId, entityId, validationFlags]);
+        const selectors = [ethers.hexlify('0x11223344')];
+        const hooks = [ethers.hexlify(ethers.randomBytes(32))];
+        await expect(this.mockFromEntrypoint.installValidation(validationConfig, selectors, installData, hooks))
+          .to.be.revertedWithCustomError(this.mock, 'ERC6900ModuleInterfaceNotSupported')
+          .withArgs(moduleId, '0x46c0c1b4');
```
Amxx, line 38: this value should be defined as a constant

contracts/account/extensions/AccountERC6900.sol
arr00: Note this is aligned with ERC7579 impl https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/03a756ced4410403f4f11746b483c0619733c633/contracts/account/extensions/AccountERC7579.sol#L70-L72

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+                )
+            );
+        } else {
+            // validation functions have a configurable range of applicability.
+            // This can be configured with selectors installed to a validation
+            require(
+                !validation.selectors.contains(executionSelector),
+                ERC6900Utils.ERC6900MissingValidationForSelector(executionSelector)
+            );
+        }
```
arr00, line 194: This part of the ERC is somewhat ambiguous to me. Should a validation that has the `isGlobal` flag be able to also specify selectors for which it is valid? Which approach do you think is better?

Note: if we do go forward with this interpretation, maybe we should disallow adding validation selectors when the `isGlobal` flag is true? https://github.com/jeremyjams/openzeppelin-community-contracts/blob/a0c58324195ab1bff1f4b8d8d9c4e655c6a8ec8d/contracts/account/extensions/AccountERC6900.sol#L280-L285

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+        if (executionSelector == IERC6900ModularAccount.execute.selector) {
+            (address target, , ) = abi.decode(userOp.callData[4:], (address, uint256, bytes));
+            require(target != address(this), ERC6900Utils.ERC6900InvalidExecuteTarget());
+        }
+        if (executionSelector == IERC6900ModularAccount.executeBatch.selector) {
+            Call[] memory calls = abi.decode(userOp.callData[4:], (Call[]));
+            for (uint256 i = 0; i < calls.length; i++) {
+                require(calls[i].target != address(this), ERC6900Utils.ERC6900InvalidExecuteTarget());
+            }
+        }
```
arr00, line 206: Given that there are 3 different validation flows, maybe this should be in its own function? Could we have the validation be part of these functions prior to running?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+            data,
+            validationAuth
+        );
+        return _execute(address(this), msg.value, data);
+    }
+
+    /**
+     * @dev Validates a user operation with {_signableUserOpHash} and returns the validation data
+     * if the module specified by the first 20 bytes of the nonce key is installed. Falls back to
+     * {AccountCore-_validateUserOp} otherwise.
```
arr00, line 156: Is the fallback currently implemented?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+     *
+     * This function delegates the signature validation to a validation module if the first 20 bytes of the
+     * signature correspond to an installed validator module.
+     *
+     * See {_extractSignatureValidator} for the module extraction logic.
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
```
arr00, line 222: Given https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80#discussion_r1949962170 it seems like super is not being called, as such, when is this function utilized?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+                execution.executionHooks.add(bytes32(abi.encodePacked(module, entityId, isPreHook, isPostHook))),
+                ERC6900Utils.ERC6900AlreadySetExecutionHookForExecution()
+            );
+        }
+        // The account SHOULD add all supported interfaces as specified in the manifest.
+        bytes4[] memory interfaceIds = manifest.interfaceIds;
+        for (uint256 i = 0; i < interfaceIds.length; i++) {
+            bytes4 interfaceId = interfaceIds[i];
+            require(!_interfaceIds[interfaceId], "Interface already set"); // TODO use custom error
+            _interfaceIds[interfaceId] = true;
```
arr00, line 380: What if two execution functions add the same interface id? Are all situations not allowed?

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+        bytes calldata installData,
+        bytes[] calldata hooks
+    ) internal virtual {
+        (ModuleEntity moduleEntity, ValidationFlags _validationFlags) = validationConfig.unpack();
+        (address module, uint32 entityId) = moduleEntity.unpack();
+        // Modules MUST implement ERC-165 for IModule.
+        require(
+            ERC165Checker.supportsInterface(module, type(IERC6900Module).interfaceId),
+            ERC6900Utils.ERC6900ModuleInterfaceNotSupported(module, type(IERC6900Module).interfaceId)
+        );
```
arr00, line 277: We can't always call `supportsInterface` since the module may be an EOA (or other non-module) used for direct validation.

package.json
jeremyjams: Only a local issue resolved by bumping my node version to a greater minor.
`package.json` has now been set back to its original state.

contracts/account/utils/ERC6900Utils.sol
jeremyjams: Yes see [Remove duplicated HookConfig extractor](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/078c232c8f5437fc0d864810ffad72a391de5444).

contracts/account/extensions/AccountERC6900.sol
```diff
@@ -0,0 +1,494 @@
...
+        _installValidation(validationConfig, selectors, installData, hooks);
+    }
+
+    /// @inheritdoc IERC6900ModularAccount
+    function uninstallValidation(
+        ModuleEntity validationFunction,
+        bytes calldata uninstallData,
+        bytes[] calldata hookUninstallData
+    ) public virtual onlyEntryPointOrSelf {
+        // _uninstallModule(moduleTypeId, module, deInitData);
```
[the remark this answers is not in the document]
jeremyjams: Added wip checklist [here](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80#issue-2843155484).

contracts/account/extensions/AccountERC6900.sol
jeremyjams: https://eips.ethereum.org/EIPS/eip-6900#validation-functions:
> If the validation function is attempted to be used for signature validation and the flag isSignatureValidation is set to false, validation MUST revert.

contracts/account/extensions/AccountERC6900.sol
jeremyjams: Thank you, updated in [Bubble revert reason when execute fails](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/ace9b0300545008a3d312bf6cee0cba999d88d13).

contracts/account/utils/ERC6900Utils.sol
jeremyjams: Ok, see [Move errors to AccountERC6900](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/dbb5521286bcd9ff10343c93be7109a6940674a3).

contracts/account/extensions/AccountERC6900.sol
jeremyjams: Thank you, indeed fallback was not implemented at all, please see [Init fallback](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/19df12ef2a73862cfc9eebf04525f02e0a617034) now.

contracts/account/extensions/AccountERC6900.sol
jeremyjams: The `AccountERC6900` being an `ERC7739`, this function is at least called from there, what do you think?
https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/03a756ced4410403f4f11746b483c0619733c633/contracts/utils/cryptography/ERC7739.sol#L39
https://eips.ethereum.org/EIPS/eip-6900#validation-call-flow
> Signature validation happens within the account’s implementation of the function isValidSignature, defined in ERC-1271.

contracts/account/extensions/AccountERC6900.sol
arr00: Ah, I was referring to the fallback to super mentioned in `_validateUserOp`. Sorry for the confusion.

contracts/account/extensions/AccountERC6900.sol
jeremyjams: Ok thanks, see [Check validation applicability](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/6e39db13e8a61945a675f57c83676424d0cfee4d), are you thinking about that?

contracts/account/extensions/AccountERC6900.sol
jeremyjams: You are right, a same interface can be added multiple times as ref implem show, see [Count supported interfaces](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/21949b099c4c2a5c564a2a934cd6062599d554a0).

contracts/account/extensions/AccountERC6900.sol
jeremyjams: I'm not sure to understand, can a module be an EOA?

contracts/account/extensions/AccountERC6900.sol
Amxx: fair enough.

contracts/account/extensions/AccountERC6900.sol
Amxx: @arr00 mentioned to me that ERC-6900 requires 1271 to be implemented using validation modules, and that having ERC-7739 + `_rawSignatureValidation` is not the right approach.

However, having a quick look at ERC-6900 and I am not seeing that. @arr00, could you please direct me to the relevant part ?

arr00: Here is the paragraph on the validation call flow. https://eips.ethereum.org/EIPS/eip-6900#validation-call-flow

contracts/account/extensions/AccountERC6900.sol
arr00: See the paragraph on [direct call validation](https://eips.ethereum.org/EIPS/eip-6900#direct-call-validation) in the ERC. Yes, you can install a validation function with the module being an EOA.

contracts/account/extensions/AccountERC6900.sol
arr00: Sorry I was a little confused here. I think the callflow is sound. That said, do you think it might make more sense to implement `ERC7739` on the module side as opposed to the account side?

contracts/account/extensions/AccountERC6900.sol
jeremyjams: Ok, about your question "Should a validation that has the isGlobal flag be able to also specify selectors for which it is valid?"

We can see from the ERC:

1. > The standard supports a isGlobalValidation flag for validation functions, which means that this function is added to a global validation pool and can validate any of the execution functions that expose themselves to global validation via allowGlobalValidation flag.
2. > validation functions [...] can be configured with selectors installed to a validation

From than I understand that:
1. Any execution function being declared as allowed for global validation can be validated by a validation set as global
2. and Any execution function can be validated by a validation having the execution selector installed

From that I understand that a same validation could validate both rules, then I feel we should not "disallow adding validation selectors when the isGlobal flag is true". What do think, do you still stick to your first impression?

contracts/account/extensions/AccountERC6900.sol
jeremyjams: "not convinced we want to use 7739 in the case of an ERC6900 account"
I think I get your point, preventing signature replays should be the responsibility of the validator module. What you suggest is aligned with [related reference validator module](https://github.com/erc6900/reference-implementation/blob/develop/src/modules/validation/SingleSignerValidationModule.sol#L95-L96).

Please see new "[Expose ERC-1271 signature validation](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/80/commits/a5668399f58053401b6a00b31ccb4cc0d0ab01e2)" commit.

contracts/account/extensions/AccountERC6900.sol
arr00: Yes that looks good

contracts/account/extensions/AccountERC6900.sol
arr00: I think we could go either way. The reference impl actually does something similar to what we have here already (either global or selector based, not both). Go with whatever you think makes more sense for how validation functions will be utilized.

contracts/account/extensions/AccountERC6900.sol
adamegyed: Hey all, I wanted to mention here that the ERC-6900 [spec](https://eips.ethereum.org/EIPS/eip-6900) does not require that this check happens onchain, and we realized that having this as non-standard behavior in the reference implementation is probably confusing. So, we've decided to remove it from the RI here: https://github.com/erc6900/reference-implementation/pull/214

Per our discussion with @Amxx and @ernestognw, we're still requiring this to be implemented by modules, and moving the responsibility of checking to be off-chain (i.e. by wallets) prior to installing a module.

PR #81, opened 2025-02-13
Add FUNDING.json

ernestognw:


Amxx:
> Remove drips wallet, keep only.

What is the issue with having drips here ? Maybe its not used today, but its not going to cause any issues is it ? It feels like having consistency (the same file on every repo) would be simpler to manage for us than having different version everywhere.

arr00:
If we do have the same file should we just use a symlink?

xdaluca:
> > Remove drips wallet, keep only.
> 
> What is the issue with having drips here ? Maybe its not used today, but its not going to cause any issues is it ? It feels like having consistency (the same file on every repo) would be simpler to manage for us than having different version everywhere.

you are right I think its fine to keep the wallet address

PR #82, opened 2025-02-13
ERC-7786 N-of-M Aggregator

Amxx:


contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+    error ERC7786AggregatorGatewayAlreadyRegistered(address gateway);
+    error ERC7786AggregatorGatewayNotRegistered(address gateway);
+    error ERC7786AggregatorThresholdViolation();
+
+    /****************************************************************************************************************
+     *                                        S T A T E   V A R I A B L E S                                         *
+     ****************************************************************************************************************/
+
+    /// @dev address of the matching router for a given CAIP2 chain
+    mapping(string caip2 => string) private _remoteRouters;
```
frangio, line 55: Rename router to aggregator.

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+
+    /// @dev Tracking of the received message pending final delivery
+    mapping(bytes32 id => Tracker) private _trackers;
+
+    /// @dev List of authorized IERC7786 gateways (M is the length of this set)
+    EnumerableSet.AddressSet private _gateways;
+
+    /// @dev Threshold for message reception (the threshold opf the sending side is applied on the receiving side)
```
frangio, line 63:
```suggestion
    /// @dev Threshold for message reception
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+        }
+        _setThreshold(threshold_);
+    }
+
+    // ============================================ IERC7786GatewaySource ============================================
+
+    /// @inheritdoc IERC7786GatewaySource
+    function supportsAttribute(bytes4 selector) external view returns (bool) {
+        for (uint256 i = 0; i < _gateways.length(); ++i)
+            if (!IERC7786GatewaySource(_gateways.at(i)).supportsAttribute(selector)) return false;
```
frangio:
```suggestion
        for (uint256 i = 0; i < _gateways.length(); ++i) {
            if (!IERC7786GatewaySource(_gateways.at(i)).supportsAttribute(selector)) return false;
        }
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+        return true;
+    }
+
+    /// @inheritdoc IERC7786GatewaySource
+    function sendMessage(
+        string calldata destinationChain,
+        string memory receiver, // using memory instead of calldata avoids stack too deep error
+        bytes memory payload, // using memory instead of calldata avoids stack too deep error
+        bytes[] memory attributes // using memory instead of calldata avoids stack too deep error
```
frangio, line 97:
```suggestion
    /// @inheritdoc IERC7786GatewaySource
    /// @dev Using memory instead of calldata avoids stack too deep errors
    function sendMessage(
        string calldata destinationChain,
        string memory receiver,
        bytes memory payload,
        bytes[] memory attributes
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+
+    function _setThreshold(uint8 newThreshold) internal virtual {
+        if (newThreshold == 0 || _threshold > _gateways.length()) revert ERC7786AggregatorThresholdViolation();
+        _threshold = newThreshold;
+        emit ThresholdUpdated(newThreshold);
+    }
+
+    // NOTE: once a router is registered for a given chainId, it cannot be updated
+    function _registerRemoteRouter(string memory caip2, string memory router) internal virtual {
+        if (bytes(_remoteRouters[caip2]).length > 0) revert ERC7786AggregatorRemoteAlreadyRegistered(caip2);
```
frangio, line 272: Is there a concrete reason why it shouldn't be possible to overwrite an existing value?

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+
+    // ============================================== IERC7786Receiver ===============================================
+
+    /// @inheritdoc IERC7786Receiver
+    function executeMessage(
+        string calldata sourceChain, // CAIP-2 chain identifier
+        string calldata sender, // CAIP-10 account address (does not include the chain identifier)
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable whenNotPaused returns (bytes4) {
```
frangio: Can you add docs explaining why this should be payable even though that implies `drainEth` is needed?

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+    function pause() public virtual onlyOwner {
+        _pause();
+    }
+
+    function unpause() public virtual onlyOwner {
+        _unpause();
+    }
+
+    /// @dev Recovery method in case ether is ever received through {executeMessage}
+    function drainEth(address payable to) public virtual onlyOwner {
```
frangio, line 247:
```suggestion
    /// @dev Recovery method in case value is ever received through {executeMessage}
    function sweep(address payable to) public virtual onlyOwner {
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+        if (msg.value > 0) revert ERC7786AggregatorValueNotSupported();
+        // address of the remote router, revert if not registered
+        string memory router = getRemoteRouter(destinationChain);
+
+        // wrapping the payload
+        bytes memory wrappedPayload = abi.encode(++_nonce, msg.sender.toChecksumHexString(), receiver, payload);
+
+        // Post on all gateways
+        Outbox[] memory outbox = new Outbox[](_gateways.length());
+        bool needsId = false;
```
frangio, line 109: We can just use `outbox.length > 0`.

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+        }
+
+        if (needsId) {
+            outboxId = keccak256(abi.encode(outbox));
+            emit OutboxDetails(outboxId, outbox);
+        }
+
+        emit MessagePosted(
+            outboxId,
+            CAIP10.local(msg.sender),
```
frangio, line 133: We've already done ASCII conversion at this point, can we reuse it?

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+    ) external payable whenNotPaused returns (bytes4) {
+        // Check sender is a trusted remote router
+        if (!_remoteRouters[sourceChain].equal(sender)) revert ERC7786AggregatorInvalidCrosschainSender();
+
+        // Message reception tracker
+        bytes32 id = keccak256(abi.encode(sourceChain, sender, payload, attributes));
+        Tracker storage tracker = _trackers[id];
+
+        // If call originates from a trusted gateway
+        if (_gateways.contains(msg.sender) && !tracker.receivedBy[msg.sender]) {
```
frangio, line 157: Any reason not to revert if `msg.sender` is not a trusted gateway?

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+                (sourceChain, originalSender, unwrappedPayload, attributes)
+            );
+            // slither-disable-next-line reentrancy-no-eth
+            (bool success, bytes memory returndata) = receiver.parseAddress().call(call);
+
+            if (
+                success &&
+                returndata.length >= 32 &&
+                abi.decode(returndata, (bytes4)) == IERC7786Receiver.executeMessage.selector
+            ) {
```
frangio, line 192: There's a weird edge case where the call succeeds but doesn't return the magic value. I think the gateway should revert in this case. It doesn't currently.

In particular, I don't think we should return the magic value if the underlying gateway doesn't.

frangio (changes requested):

contracts/crosschain/ERC7786Aggregator.sol
Amxx: The logic was that the admin could setup new chains, but not update/break existing links. The is to ensure that something that worked in the past would not be broken by the admin.

This logic comes from the Axelar adaptor.

But given that the admin can add gateways,  remove gateways, and update the threshold ... maybe that doesn't bring much value to the end user.

Should we remove that check ?

contracts/crosschain/ERC7786Aggregator.sol
Amxx: outbox.length is always the number of gateway (the M in N-of-M). 

Outbox doesn't contain entries "only for gateways that have a non zero outboxid", it will include entries for all gateways. This boolean checks if at least one of them is non-zero.

contracts/crosschain/ERC7786Aggregator.sol
Amxx: Last time I tried doing that, the additional variables caused stack too deep.

If we wind a way to not get that error, I'd do it!

contracts/crosschain/ERC7786Aggregator.sol
Amxx: See https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/82#discussion_r1985292425

contracts/crosschain/ERC7786Aggregator.sol
Amxx: This is related to the question about "not reverting if msg.sender is not a trusted gateway"

My idea is that I am not 100% sure how the incomming message can be replaied if it fails. So I thought that:
- aknowledgment by the gateway should always be registered.
  - don't revert if execution reverts
  - alwasy return the magic value to the gateway (if the message relayed is properly formated)
- if we reach consensus, and the execution fails, we should be able to retry the execution
  - I think this should only involve the aggregator (no need to dive into the protocoles to have one of them retrigger)
  - This should be node by anyone.
  
So that is why other msg.sender don't necessarily revert. An execution by someone not trusted does
- check that the message was not executed yet. 
  - If it was already executed, revert
- check that the message achieved the threshold
  - If it reached the threshold (and was not executed) try executing it again
  - If it did not reach consensus, do do nothing (maybe we should revert)

Amxx: Does that make sens ?

frangio: Ok, so the goal is to allow anyone to retry failed execution. That makes sense, but the edge case stands: if we call `executeMessage` on the receiver and the receiver succeeds but _doesn't return the magic value_, we had some effect on a receiver that did not acknowledge receipt like the protocol mandates and we should revert out of caution. If we had a way to revert the subcall only but still register the message as received I would do that... but that adds complexity that I'd rather not add now, and we can also assume that the underlying gateway has its own replay mechanism if needed for some reason.

contracts/crosschain/ERC7786Aggregator.sol
frangio, line 158: I'd rather do something like this:

```suggestion
        if (_gateways.contains(msg.sender)) {
            require(!tracker.receivedBy[msg.sender]));
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,278 @@
...
+            // Count number of time received
+            tracker.receivedBy[msg.sender] = true;
+            ++tracker.countReceived;
+            emit Received(id, msg.sender);
+
+            // if already executed, leave gracefully
+            if (tracker.executed) return IERC7786Receiver.executeMessage.selector;
+        } else {
+            // if already executed, revert
+            if (tracker.executed) revert ERC7786AggregatorAlreadyExecuted();
```
frangio, line 168: How do you explain the difference in behavior here? Why revert if execute but only when the caller isn't a gateway?

contracts/crosschain/ERC7786Aggregator.sol
Amxx: that is saying "a registered gateway cannot perform the retry if its aknoledgment was already registered"

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,281 @@
...
+            );
+            // slither-disable-next-line reentrancy-no-eth
+            (bool success, bytes memory returndata) = receiver.parseAddress().call(call);
+
+            if (!success) {
+                // rollback to enable retry
+                tracker.executed = false;
+                emit ExecutionFailed(id);
+            } else if (
+                returndata.length >= 32 && abi.decode(returndata, (bytes4)) == IERC7786Receiver.executeMessage.selector
```
frangio, line 195:
```suggestion
                abi.decode(returndata, (bytes4)) == IERC7786Receiver.executeMessage.selector
```

I understand that ABI decoding could revert, but length < 32 is not the only reason, it could also have nonzero padding. I don't think we should check for every condition so we should just let it fail.

contracts/crosschain/ERC7786Aggregator.sol
Amxx: If the call was already executed:
- there is no reason for a third party to try to trigger the execution
- a gateway that is registered (and has not yet aknoledge) is just sending an aknoledgement AFTER the message was executed. We agreed this should not revert.

contracts/crosschain/ERC7786Aggregator.sol
frangio: Let's remove it.

contracts/crosschain/ERC7786Aggregator.sol
frangio: Hm, I see.

Can you write this into comments in the code? There is a lot implicit. It's not obvious that this function is used for retries.

Amxx: see: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/82/commits/a4399b8d2796863a51e0fd5febda60672a9859b9

contracts/crosschain/ERC7786Aggregator.sol
Amxx: see: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/82/commits/1cd890cec940f27e77cef857b9b68e13d2beb0d0

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -0,0 +1,314 @@
...
+    ) public payable virtual whenNotPaused returns (bytes4) {
+        // Check sender is a trusted remote aggregator
+        if (!_remotes[sourceChain].equal(sender)) revert ERC7786AggregatorInvalidCrosschainSender();
+
+        // Message reception tracker
+        bytes32 id = keccak256(abi.encode(sourceChain, sender, payload, attributes));
+        Tracker storage tracker = _trackers[id];
+
+        // If call originates from a trusted gateway
+        if (_gateways.contains(msg.sender) && !tracker.receivedBy[msg.sender]) {
```
frangio, line 196:
```suggestion
        // If call is first from a trusted gateway
        if (_gateways.contains(msg.sender) && !tracker.receivedBy[msg.sender]) {
```

PR #83, opened 2025-02-17
Refactor the way ERC-1271 validation is handled by modular ERC-7579 accounts

Amxx:
Currently (commit 03a756ced4410403f4f11746b483c0619733c633) the Account framework comes with some quirks:

- `AccountERC7579` depends on `ERC7739` to implement ERC-1271, and implements the abstract signer interface using validation modules.
  - This means that any signature handeled by validation module will have to also be ERC-7739 wrapped, even though the ERC-7579 modules have an interface that supports mechanism for cross-account replay protection. In cases where modules would want to implement ERC-7739 or similar (which they probably should), we would be wrapping 2 layers of protection.
  - There is a side effect where a userOperation that would fail the `IERC7579Validator.validateUserOp` check would be re-submitted to a validation module's `IERC7579Validator.isValidSignatureWithSender` (through the `super._validateUserOp`)
  
The goals of this PR are to:
- preserve ERC-7739 as a recommanded "default" in `Account`, working with abstract signer (such as EIP-7702)
- change ERC-7579 to not act as a signer, but instead implement `_isValidSignature` through modules directly (without going to an abstract signer)
- remove the posibility for `IERC7579Validator.isValidSignatureWithSender` to approve/be part of the user operation validation workflow.
- Maintain the ability for abstract signer based ERC-7739 signature to co-exist with ERC-7579 module based validation (that don't have the 7739 layer)


Amxx:
One argument in support of this change is that ERC-7579 validation module may be used with "session keys", or more genreally signing keys that are oblivious to the wallet implementation (they only know about the module). Such keys would not be 7739 aware, and would not work if the account does an ERC-7739 preprocessing before calling the module.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -164,6 +163,30 @@ abstract contract AccountERC7579 is
...
+        if (signature.length >= 20) {
+            (address module, bytes calldata innerSignature) = _extractSignatureValidator(signature);
+            // if module is not installed, skip
+            if (isModuleInstalled(MODULE_TYPE_VALIDATOR, module, Calldata.emptyBytes())) {
+                // try validation, skip any revert
+                try IERC7579Validator(module).isValidSignatureWithSender(address(this), hash, innerSignature) returns (
+                    bytes4 magic
+                ) {
+                    if (magic == IERC1271.isValidSignature.selector) return magic;
+                } catch {}
```
ernestognw, line 184: To confirm, ERC-7579 is not explicit whether the function should revert. So we're basically protecting against an unexpected revert. Is this correct?

Also side note but this is the kind of code that may be better to do in assembly. For another PR

ernestognw (approved):

ernestognw:
For the record. I thought about signature resolution while working on the original implementation and my initial posture was a bit more defensive.

The general design rules I followed were:

1) Always make sure to call super
2) Let the account protect himself by default using ERC-7739

I understand that forcing ERC-7739 is not ideal and the more we converse about this I get more convinced that is ok to allow the account decide what they want to sign. I initially thought the user might not be aware and thus, a risk is that they sign a message that's potentially replayable without noticing.

So far I agree with the tradeoffs of this PR.

PR #85, opened 2025-02-21
Add ERC-7579 Social Recovery Executor Module (In Progress)

gonzaotc:
### Overview
Introduces a Social Recovery Executor Module that implements account recovery functionality following the ERC-7579 standard. The module enables accounts to designate guardians who can collectively initiate and execute recovery operations in case of key loss or compromise.

### Key Features
- **Configurable Threshold**: Customizable M-of-N threshold for recovery signatures
- **Timelock Protection**: Enforced delay between recovery initiation and execution
- **Recovery Cancellation & configuration**: Allows the ERC-7579 Account and designated Guardians to cancel ongoing recovery attempts and reconfigure the recovery module.
- **Replay Attack Protection**: Implements EIP-712 typed data signing with nonce protection

### On-going
- [x] Implemented core functionality
- [x] Added comprehensive tests
- [x] Enhance guardian signature security by including `executionCalldata` in the EIP-712 RECOVERY_TYPEHASH and validating at execution.
- [x] Improve O(n^2) guardian signature verification
- [x] Prepare module documentation
- [ ] Enhance module docs (remove @notice and @param in favor of @dev, clarify explanations)
- [ ] Gas testing & optimizations
- [ ] Improve and sophisticate testing

*Note*: updated submodules for making use of `EnumerableSet.clear`


package.json
Amxx, line 1: Are those changes needed here ?

test/helpers/erc7579-modules.js
```diff
@@ -0,0 +1,17 @@
+const SocialRecoveryExecutorHelper = {
+  RecoveryStatus: {
+    NotStarted: 0,
+    Started: 1,
+    Cancelled: 2,
+  },
```
Amxx, line 6: Using `const {Enum } = require("lib/@openzeppelin-contracts/test/helpers/enums");` you can do 

```suggestion
RecoveryStatus: Enum('NotStarted', 'Started', 'Cancelled');
```

test/helpers/erc7579-modules.js
```diff
@@ -0,0 +1,17 @@
...
+  RECOVERY_MESSAGE_TYPE: {
+    RecoveryMessage: [
+      { name: 'account', type: 'address' },
+      { name: 'nonce', type: 'uint256' },
+    ],
```
Amxx, line 11: Similarly, using `const { formatType } = require('@openzeppelin/contracts/test/helpers/eip712');` you can do

```suggestion
    RecoveryMessage: formatType({ account: 'address', nonce: 'uint256' }),
```

test/account/extensions/ERC7579Modules/SocialRecoveryExecutorModule.test.js
```diff
@@ -0,0 +1,708 @@
...
+  const newSigner = ethers.Wallet.createRandom();
+
+  // ERC-4337 account
+  const helper = new ERC4337Helper();
+  const env = await helper.wait();
+  const accountMock = await helper.newAccount('$AccountERC7579Mock', [
+    'AccountERC7579',
+    '1',
+    validatorMock.target,
+    ethers.solidityPacked(['address'], [initialSigner.address]),
```
Amxx, line 33: That should be just the same
```suggestion
    initialSigner.address,
```

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+
+    error RecoveryAlreadyStarted();
+    error RecoveryNotStarted();
+    error RecoveryNotReady();
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                CONSTANTS & STORAGE
+    //////////////////////////////////////////////////////////////////////////*/
+
+    string public constant NAME = "SocialRecoveryExecutor";
```
Amxx, line 69: I think we need to rationalize here. 

Right now there are:
- public contracts (`NAME` and `VERSION` that cannot be overriden)
- external view functions (`name()` and `version()` that are not virtual)
- name and version in EIP712 (set at construction, that can be fetched using `IERC5267.eip712Domain()`

since the first two are NOT required by any module interface, I'd remove them. All that should be set trough constructor params

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+    string public constant VERSION = "0.0.1";
+    uint8 public constant MAX_GUARDIANS = 32;
```
Amxx, line 71: I'd use a `function maxGuardians() public view virtual { return 32; }` function that can be overriden

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+            revert InvalidTimelock();
+        }
+
+        for (uint256 i = 0; i < _guardians.length; i++) {
+            if (_guardians[i] == address(0)) {
+                revert InvalidGuardian();
+            }
+            bool added = _recoveryConfigs[account].guardians.add(_guardians[i]);
+            if (!added) {
+                revert InvalidGuardians();
```
Amxx, line 157: Its more of a "duplicate" or "already active" then an "invalid" issue.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+    /// @notice Executes the recovery process after timelock period
+    /// @dev Can only execute calls to reconfigure installed Validator Modules
+    /// @param account The account to execute recovery for
+    /// @param guardianSignatures Array of GuardianSignature authorizing the recovery
+    /// @param executionCalldata The encoded call to reconfigure a Validator Module
+    function executeRecovery(
+        address account,
+        GuardianSignature[] calldata guardianSignatures,
+        bytes calldata executionCalldata
+    ) external whenRecoveryIsReady(account) nonReentrant {
```
Amxx, line 225: We don't need `nonReentrant` here, do we?

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+        (address targetValidatorModule, , ) = ERC7579Utils.decodeSingle(executionCalldata);
+
+        // check the target is an installed Validator Module on the account.
+        if (!IERC7579ModuleConfig(account).isModuleInstalled(MODULE_TYPE_VALIDATOR, targetValidatorModule, "")) {
+            revert InvalidInstalledValidatorModule();
+        }
+
+        // create single call mode
+        Mode mode = ERC7579Utils.encodeMode(
+            ERC7579Utils.CALLTYPE_SINGLE, // enforce single call
```
Amxx, line 250: Forcing single is not great. What if you want to do multiple operations during the recovery.

IMO the call should just be
```Address.functionCall(account, executionCalldata)```;

You may do some check that about the first 4 bytes of the executionCalldata (selector) ... to make sure it actually encodes a `executeFromExecutor` ... but honestly, what else is an executionModule going to call anyway ?

File: contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol [hunk 9]
```diff
@@ -0,0 +1,475 @@
...
+
+        emit RecoveryStarted(account);
+    }
+
+    /// @notice Executes the recovery process after timelock period
+    /// @dev Can only execute calls to reconfigure installed Validator Modules
+    /// @param account The account to execute recovery for
+    /// @param guardianSignatures Array of GuardianSignature authorizing the recovery
+    /// @param executionCalldata The encoded call to reconfigure a Validator Module
+    function executeRecovery(
```
Amxx, line 221: I looks like there is a massive security flaw here, the signature does NOT bind to the executionCalldata !!!

If I intercept the tx in the memory pool, can I front run it with a different `executionCalldata`, and use the signature from the guardians to recover to my benefit ? I don't see what is stopping this from happening.

Overall I think the design should have the guardian sign what the recovery will do, and commit that when doing `startRecovery`. That way, its clear what is going to be done.

The execute should probably not need signatures, since they were already provided at the begining. And the guardians should be able to change their mind and cancel if they realise their recovery is not good.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+            if (!guardianSignatureIsValid(account, guardianSignatures[i])) {
+                return false;
+            }
+            // check for signature duplication @TBD optimize O(n^2)
+            address currentSigner = guardianSignatures[i].signer;
+            for (uint256 j = 0; j < i; j++) {
+                if (guardianSignatures[j].signer == currentSigner) {
+                    return false;
+                }
+            }
```
Amxx, line 315: This is O(n²), which is bad.

How would you make that O(n) ?

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+    /// @return True if all signatures are valid, false otherwise
+    function guardianSignaturesAreValid(
+        address account,
+        GuardianSignature[] calldata guardianSignatures
+    ) public view returns (bool) {
+        if (guardianSignatures.length > 32) {
+            revert MaxGuardians();
+        }
+
+        for (uint256 i = 0; i < guardianSignatures.length; i++) {
```
Amxx, line 304: This loop is calling
- guardianSignatureIsValid
  - guardianSignatureIsValid
    - _getRecoveryDigest
      - allocates memory (`keccak256(abi.encode(...))`
the hash is always the same, so you are effectivelly leaking memory for no reason.

You should probably cache some values, and inline some functions.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                MODULE MANAGEMENT
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Adds a new guardian to the account's recovery configuration
+    /// @dev Can only be called by the account itself
+    /// @param guardian Address of the new guardian
+    function addGuardian(address guardian) external {
```
Amxx, line 345: this logic is duplicated in `onInstall`.

You should probably have an internal `function _addGuardian(address account, address guardian) internal`

That you could call from here and from `onInstall`

Amxx: In general, we try to always provide internal virtual version of all these function (with account as a parameter), and then have external functions call the internal ones with msg.sender.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+
+        // execute the recovery.
+        AccountERC7579(payable(account)).executeFromExecutor(Mode.unwrap(mode), executionCalldata);
+
+        emit RecoveryExecuted(account, targetValidatorModule);
+    }
+
+    /// @notice Cancels an ongoing recovery process
+    /// @dev Can only be called by the ERC-7579 Account itself
+    function cancelRecovery() external whenRecoveryIsTimelockOrReady(msg.sender) {
```
Amxx, line 270: If you don't increase the nonce, the same signature can be re-used immediatly.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,475 @@
...
+    /// @param account The account to query
+    /// @return RecoveryStatus enum value (NotStarted, Timelock, or Ready)
+    function getRecoveryStatus(address account) public view returns (RecoveryStatus) {
+        RecoveryConfig storage recoveryConfig = _recoveryConfigs[account];
+
+        if (recoveryConfig.recoveryStart == 0) {
+            return RecoveryStatus.NotStarted;
+        }
+
+        if (block.timestamp < recoveryConfig.recoveryStart + recoveryConfig.timelock) {
```
Amxx, line 426: duplicated sload on `recoveryStart`

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+    event ModuleInstalledReceived(address indexed account, bytes data);
+
+    event RecoveryCancelled(address indexed account);
+    event RecoveryExecuted(address indexed account);
+    event RecoveryStarted(address indexed account);
+
+    event ThresholdChanged(address indexed account, uint256 indexed threshold);
+    event TimelockChanged(address indexed account, uint256 indexed timelock);
+    event GuardianRemoved(address indexed account, address indexed guardian);
+    event GuardianAdded(address indexed account, address indexed guardian);
```
Amxx, line 73: can you explain the rational for indexing everything ?

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+            data,
+            (address[], uint256, uint256)
+        );
+        if (_guardians.length == 0) {
+            revert InvalidGuardians();
+        }
+        if (_threshold == 0 || _threshold > _guardians.length) {
+            revert InvalidThreshold();
+        }
+        if (_timelock == 0) {
```
Amxx, line 142: What is the rational for accepting `1` but refusing `0` ?

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+            revert InvalidTimelock();
+        }
+
+        for (uint256 i = 0; i < _guardians.length; i++) {
+            _addGuardian(account, _guardians[i]);
+        }
+
+        _recoveryConfigs[account].threshold = _threshold;
+        _recoveryConfigs[account].timelock = _timelock;
```
Amxx, line 151: any reason to not use `_changeThreshold` and `_changeTimelock` here ?

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+    /// @param account The ERC-7579 Account to validate signatures for
+    /// @param guardianSignatures Array of guardian signatures to validate
+    /// @param digest The digest to validate the signatures against
+    function validateGuardianSignatures(
+        address account,
+        GuardianSignature[] calldata guardianSignatures,
+        bytes32 digest
+    ) external view {
+        _validateGuardianSignatures(account, guardianSignatures, digest);
+    }
```
Amxx, line 255: What is the value of this function ?

If we want it, I'd make it "public virtual", and remove the corresponding internal one.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+    /// @return The current recovery status
+    function getRecoveryStatus(address account) public view virtual returns (RecoveryStatus) {
+        uint256 recoveryStart = _recoveryConfigs[account].recoveryStart;
+        if (recoveryStart == 0) {
+            return RecoveryStatus.NotStarted;
+        }
+        if (block.timestamp < recoveryStart + _recoveryConfigs[account].timelock) {
+            return RecoveryStatus.Started;
+        }
+        return RecoveryStatus.Ready;
```
Amxx, line 304: IMO that is clearer (but I guess its subjective)

```suggestion
        return recoveryStart == 0
            ? RecoveryStatus.NotStarted
            : (block.timestamp < recoveryStart + _recoveryConfigs[account].timelock)
            ? RecoveryStatus.Started
            : RecoveryStatus.Ready;
```

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,462 @@
...
+
+    /// @notice Gets the current recovery status of an ERC-7579 Account
+    /// @param account The ERC-7579 Account to get the recovery status for
+    /// @return The current recovery status
+    function getRecoveryStatus(address account) public view virtual returns (RecoveryStatus) {
+        uint256 recoveryStart = _recoveryConfigs[account].recoveryStart;
+        if (recoveryStart == 0) {
+            return RecoveryStatus.NotStarted;
+        }
+        if (block.timestamp < recoveryStart + _recoveryConfigs[account].timelock) {
```
Amxx, line 301: What is the rational for storing the start, and doing that math when recovering the status versus storing the end (doing the add math when initiating a recovery) ?

I'm not sure if one is clearly better than the other, but we should documetn that choice with regard to the delay changing during the recovery.

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,459 @@
...
+                    guardianSignatures[i].signature
+                )
+            ) {
+                revert InvalidGuardianSignature();
+            }
+            // ensures signers are unique in O(n), requires sorted signatures by signer address in ascending order
+            if (i > 0 && uint160(guardianSignatures[i].signer) <= uint160(guardianSignatures[i - 1].signer)) {
+                revert DuplicatedOrUnsortedGuardianSignatures();
+            }
+        }
```
Amxx, line 364: lets avoid duplicated decode/lookup in calldata
```suggestion
        address lastSigner = address(0);
        for (uint256 i = 0; i < guardianSignatures.length; i++) {
            address signer = guardianSignatures[i].signer;
            if (
                !isGuardian(account, signer) ||
                !SignatureChecker.isValidSignatureNow(
                    signer,
                    digest,
                    guardianSignatures[i].signature
                )
            ) {
                revert InvalidGuardianSignature();
            }
            if (uint160(signer) <= uint160(lastSigner)) 
                revert DuplicatedOrUnsortedGuardianSignatures();
            lastSigner = signer;
        }
```

Amxx (commented):
This PR contains a lot of different visibility and overridability.

There are external, public and internal functions, some of which are virtual, some of which are not. Is there any undocummented logic to it? Note that at OZ we try to only use `public virtual` and `internal virtual` (and `private`)

contracts/account/extensions/ERC7579Modules/SocialRecoveryExecutor.sol
Amxx: having the signature provided in order or signer (and just checking the ordering) is what I was thinking. IMO its the simplest and most effective way. AFAIK that is what most Multisig do.

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,453 @@
...
+        emit RecoveryCancelled(account);
+    }
+
+    /// @notice Adds a new guardian to the account's recovery configuration
+    /// @param account The ERC-7579 Account to add the guardian to
+    /// @param guardian Address of the new guardian
+    function _addGuardian(address account, address guardian) internal virtual {
+        if (guardian == address(0)) {
+            revert InvalidGuardian();
+        }
```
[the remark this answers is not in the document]
ernestognw: Also, in `_validateGuardianSignatures` we're doing:

```solidity
if (uint160(signer) <= uint160(lastSigner)) {
    revert DuplicatedOrUnsortedGuardianSignatures();
}
```

So we can implicitly exclude the `address(0)` by just changing `<=` for `<=`

```solidity
if (uint160(signer) < uint160(lastSigner)) {
    revert DuplicatedOrUnsortedGuardianSignatures();
}
```

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,453 @@
...
+    function _cancelRecovery(address account) internal virtual {
+        _recoveryConfigs[account].pendingExecutionHash = bytes32(0);
+        _recoveryConfigs[account].recoveryStart = 0;
+        emit RecoveryCancelled(account);
+    }
+
+    /// @notice Adds a new guardian to the account's recovery configuration
+    /// @param account The ERC-7579 Account to add the guardian to
+    /// @param guardian Address of the new guardian
+    function _addGuardian(address account, address guardian) internal virtual {
```
ernestognw, line 403: I think the signature of both the `_addGuardian` and `_removeGuardian` functions must allow batch adding/removing guardians natively. It'd be more elegant but it's also opinionated since the account can batch operations. It's just that batching operations at the account level may require extra work than just calling a batched function directly

```solidity
for (uint256 i = 0; i < signers.length; i++) {
    if (!_associatedSigners[account].add(guardians[i]))
        revert AlreadyGuardian(account, guardians[i]);
}
emit GuardiansAdded(account, guardians);
```

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,453 @@
...
+    /// @notice Verifies multiple guardian signatures for a given digest
+    /// @dev Ensures signatures are unique, and threshold is met
+    /// @param account The account the signatures are for
+    /// @param guardianSignatures Array of guardian signatures sorted by signer address in ascending order
+    /// @param digest The digest to verify the signatures against
+    function _validateGuardianSignatures(
+        address account,
+        GuardianSignature[] calldata guardianSignatures,
+        bytes32 digest
+    ) internal view virtual {
```
ernestognw, line 367: It'd be better if this function returns a boolean instead of reverting. This way there can be a public view getter called `validateGuardianSignatures` that allows to test if the guardian signatures construction is correct

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,453 @@
...
+    /// @notice Overload: Allows guardians to cancel a recovery process
+    /// @dev Requires threshold signatures, similar to starting recovery
+    /// @param account The account to cancel recovery for
+    /// @param guardianSignatures Array of guardian signatures authorizing cancellation
+    /// @custom:security Uses same signature threshold as recovery initiation
+    function cancelRecovery(
+        address account,
+        GuardianSignature[] calldata guardianSignatures
+    ) public virtual whenRecoveryIsStartedOrReady(account) {
+        bytes32 digest = _getCancelRecoveryDigest(account, nonces(account));
```
arr00, line 239:
```suggestion
        bytes32 digest = _getCancelRecoveryDigest(account, _useNonce(account));
```

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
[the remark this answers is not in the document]
arr00: If we have `<`, that removes the duplication check. We could increment the initial `lastSigner` from `address(0)` if this is a concern. `isValidSignatureNow` will always fail on sig from address 0 though since we are using `ECDSA.tryRecover`. Don't think we should think too hard here.

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
```diff
@@ -0,0 +1,453 @@
...
+        }
+        _;
+    }
+
+    modifier whenRecoveryIsStartedOrReady(address account) {
+        if (getRecoveryStatus(account) == RecoveryStatus.NotStarted) {
+            revert RecoveryNotStarted();
+        }
+        _;
+    }
```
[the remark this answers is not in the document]
ernestognw: > but note that cancelRecovery depends on the status to be one of <Started, Ready>

Oh this is an cool observation! We've had this situation in the past. My suggestion is representing the state as a bitmap, so more than 1 state can be checked with an logical and (`&`) operation:

See https://github.com/OpenZeppelin/openzeppelin-contracts/blob/acd4ff74de833399287ed6b31b4debf6b2b35527/contracts/governance/Governor.sol#L734-L750

contracts/account/extensions/ERC7579Modules/ERC7579SocialRecoveryExecutor.sol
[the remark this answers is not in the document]
ernestognw: Right, is a good observation that `<=` disallows duplication.

Still, I don't think is worth checking that `guardian != address(0)`. Agree there could be a human/software error but I consider that out of scope of the library.

PR #88, opened 2025-03-07
Add messageId in IERC7786Receiver.executeMessage

Amxx:


contracts/crosschain/axelar/AxelarGatewayDestination.sol
```diff
@@ -53,6 +84,7 @@ abstract contract AxelarGatewayDestination is AxelarGatewayBase, AxelarExecutabl
         );
 
         bytes4 result = IERC7786Receiver(receiver.parseAddress()).executeMessage(
+            messageId,
```
Amxx, line 87: Here we have a choice between:
- the axelar adaptor passes the axelar commandId to the receiver (stringified)
- the axelar adaptor passes an empty string (and hides the axelar adaptor)

Note that the AxelarExecutor (that is part of  `@axelar-network/axelar-gmp-sdk-solidity`) is implement in a way that doesn't pass the commandId to the "execution logic". If axelar believes its not necessary for execution, maybe we can accept that the adaptor doesn't forward it further and passes an empty string instead.

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -216,9 +217,10 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
             // prevent re-entry
             tracker.executed = true;
 
+            // TODO: Should the messageId be empty here?
```
Amxx, line 220: Lets discuss that

contracts/crosschain/axelar/AxelarGatewayDestination.sol
frangio: If the id exists we should forward it.

frangio: > Note that the AxelarExecutor (that is part of `@axelar-network/axelar-gmp-sdk-solidity`) is implement in a way that doesn't pass the commandId to the "execution logic".

Apparently this is only true for an older version of `AxelarExecutable`. Their v6 does pass the command id. We should just update the dep and keep using that.

https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/v6.0.6/contracts/executable/AxelarExecutable.sol#L60-L61

contracts/crosschain/ERC7786Aggregator.sol
frangio: I propose to use `keccak256(abi.encode(nonce, msg.sender))`, where `nonce` is the one in the payload.

frangio: Ah I see you're already using the `id`. That seems fine too.

frangio:
```suggestion
```

frangio (approved):

PR #91, opened 2025-03-13
Update link to ERC-7802 in ERC20Bridgeable

ericglau:
ERC-7802 was merged to the ERCs repository, so the link can be updated to point to the one from master.


arr00 (approved):
Good catch

PR #93, opened 2025-03-14
ERC7579 Social Recovery Module

Amxx:
Alternative to #85

loosely based on https://eips.ethereum.org/EIPS/eip-7093


contracts/account/modules/ERC7579SocialRecovery.sol
```diff
@@ -0,0 +1,371 @@
...
+        keccak256("StartRecovery(address account,bytes recovery,uint256 nonce)");
+    bytes32 private constant CANCEL_RECOVERY_TYPEHASH =
+        keccak256("CancelRecovery(address account,bytes recovery,uint256 nonce)");
+
+    struct AccountConfig {
+        address verifier;
+        EnumerableMapExtended.BytesToUintMap guardians;
+        Checkpoints.Trace160 thresholds;
+        bytes recoveryCall;
+        uint48 expiryTime;
```
arr00, line 31:
```suggestion
        address verifier;
        uint48 expiryTime;
        EnumerableMapExtended.BytesToUintMap guardians;
        Checkpoints.Trace160 thresholds;
        bytes recoveryCall;
```
Packing

contracts/account/modules/ERC7579SocialRecovery.sol
```diff
@@ -0,0 +1,371 @@
...
+            config.guardians.set(identity, property);
+        }
+
+        // threshold
+        for (uint256 i = 0; i < recoveryConfigArgs.thresholds.length; ++i) {
+            config.thresholds.push(
+                recoveryConfigArgs.thresholds[i].threshold,
+                recoveryConfigArgs.thresholds[i].lockPeriod
+            );
+        }
```
arr00, line 239: Should require sorting by threshold?

contracts/account/modules/ERC7579SocialRecovery.sol
```diff
@@ -0,0 +1,371 @@
...
+
+abstract contract ERC7579SocialRecovery is EIP712, Nonces, IERC7579Module, ISocialRecovery, IRecoveryPolicyVerifier {
+    using Checkpoints for *;
+    using EnumerableMapExtended for *;
+    using SafeCast for *;
+
+    bytes32 private constant START_RECOVERY_TYPEHASH =
+        keccak256("StartRecovery(address account,bytes recovery,uint256 nonce)");
+    bytes32 private constant CANCEL_RECOVERY_TYPEHASH =
+        keccak256("CancelRecovery(address account,bytes recovery,uint256 nonce)");
```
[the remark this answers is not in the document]
Amxx: True, we could skip recovery here.

contracts/account/modules/ERC7579SocialRecovery.sol
Amxx: This is not an array push, its a checkpoint.trace push. It already includes a sorting check internally

- [push](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol#L39-L45)
  - [_insert](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol#L39-L45)
    - if (length > 0)
      - [if (lastKey > key) revert](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/structs/Checkpoints.sol#L147-L149)

contracts/account/modules/ERC7579SocialRecovery.sol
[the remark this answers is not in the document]
Amxx: The only difference that makes is you can sign a valid cancel before the recovery is even proposed. not sure if we should be concerned by that

PR #94, opened 2025-03-24
Wormhole adaptor

Amxx:
Replaces #13

### DIFF
--- a/contracts/crosschain/wormhole/WormholeGatewayBase.sol
+++ b/contracts/crosschain/wormhole/WormholeGatewayBase.sol
@@ -5,11 +5,11 @@ pragma solidity ^0.8.27;
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 import {CAIP10} from "@openzeppelin/contracts/utils/CAIP10.sol";
 import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
-import {IWormholeRelayer} from "wormhole-solidity-sdk/interfaces/IWormholeRelayer.sol"; // VaaKey
+import {IWormholeRelayer} from "wormhole-solidity-sdk/interfaces/IWormholeRelayer.sol";
 
 abstract contract WormholeGatewayBase is Ownable {
     IWormholeRelayer internal immutable _wormholeRelayer;
-    uint16 internal immutable _currentChain;
+    uint16 internal immutable _wormholeChainId;
 
     mapping(string caip2 => bytes32 remoteGateway) private _remoteGateways;
     mapping(string caip2 => uint24 wormholeId) private _caipToWormholeEquivalence;
@@ -34,12 +34,12 @@ abstract contract WormholeGatewayBase is Ownable {
         _;
     }
 
-    constructor(IWormholeRelayer wormholeRelayer, uint16 currentChain) {
+    constructor(IWormholeRelayer wormholeRelayer, uint16 wormholeChainId) {
         _wormholeRelayer = wormholeRelayer;
-        _currentChain = currentChain;
+        _wormholeChainId = wormholeChainId;
     }
 
-    function gateway() public view virtual returns (address) {
+    function relayer() public view virtual returns (address) {
         return address(_wormholeRelayer);
     }
 
--- a/contracts/crosschain/wormhole/WormholeGatewayBase.sol
+++ b/contracts/crosschain/wormhole/WormholeGatewayBase.sol
@@ -1,0 +1,78 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.27;
+
+import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
+import {CAIP10} from "@openzeppelin/contracts/utils/CAIP10.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import {IWormholeRelayer} from "wormhole-solidity-sdk/interfaces/IWormholeRelayer.sol"; // VaaKey
+
+abstract contract WormholeGatewayBase is Ownable {
+    IWormholeRelayer internal immutable _wormholeRelayer;
+    uint16 internal immutable _currentChain;
+
+    mapping(string caip2 => bytes32 remoteGateway) private _remoteGateways;
+    mapping(string caip2 => uint24 wormholeId) private _caipToWormholeEquivalence;
+    mapping(uint16 wormholeId => string caip2) private _wormholeToCaipEquivalence;
+
+    /// @dev A remote gateway has been registered for a chain.
+    event RegisteredRemoteGateway(string caip2, bytes32 gatewayAddress);
+
+    /// @dev A chain equivalence has been registered.
+    event RegisteredChainEquivalence(string caip2, uint16 wormholeId);
+
+    /// @dev Error emitted when an unsupported chain is queried.
+    error UnsupportedChain(string caip2);
+    error UnsupportedChain2(uint16 wormholeId);
+
+    error ChainEquivalenceAlreadyRegistered(string caip2);
+    error RemoteGatewayAlreadyRegistered(string caip2);
+    error UnauthorizedCaller(address);
+
+    modifier onlyWormholeRelayer() {
+        require(msg.sender == address(_wormholeRelayer), UnauthorizedCaller(msg.sender));
+        _;
+    }
+
+    constructor(IWormholeRelayer wormholeRelayer, uint16 currentChain) {
+        _wormholeRelayer = wormholeRelayer;
+        _currentChain = currentChain;
+    }
+
+    function gateway() public view virtual returns (address) {
+        return address(_wormholeRelayer);
+    }
+
+    function supportedChain(string memory caip2) public view virtual returns (bool) {
+        return _caipToWormholeEquivalence[caip2] & (1 << 16) != 0;
+    }
+
+    function fromCAIP2(string memory caip2) public view virtual returns (uint16) {
+        uint24 wormholeId = _caipToWormholeEquivalence[caip2];
+        require(wormholeId & (1 << 16) != 0, UnsupportedChain(caip2));
+        return uint16(wormholeId);
+    }
+
+    function toCAIP2(uint16 wormholeId) public view virtual returns (string memory caip2) {
+        caip2 = _wormholeToCaipEquivalence[wormholeId];
+        require(bytes(caip2).length > 0, UnsupportedChain2(wormholeId));
+    }
+
+    function getRemoteGateway(string memory caip2) public view virtual returns (bytes32 remoteGateway) {
+        remoteGateway = _remoteGateways[caip2];
+        require(remoteGateway != bytes32(0), UnsupportedChain(caip2));
+    }
+
+    function registerChainEquivalence(string calldata caip2, uint16 wormholeId) public virtual onlyOwner {
+        require(_caipToWormholeEquivalence[caip2] == 0, ChainEquivalenceAlreadyRegistered(caip2));
+        _caipToWormholeEquivalence[caip2] = wormholeId | (1 << 16);
+        _wormholeToCaipEquivalence[wormholeId] = caip2;
+        emit RegisteredChainEquivalence(caip2, wormholeId);
+    }
+
+    function registerRemoteGateway(string calldata caip2, bytes32 remoteGateway) public virtual onlyOwner {
+        require(_remoteGateways[caip2] == bytes32(0), RemoteGatewayAlreadyRegistered(caip2));
+        _remoteGateways[caip2] = remoteGateway;
+        emit RegisteredRemoteGateway(caip2, remoteGateway);
+    }
+}
--- a/contracts/mocks/crosschain/wormhole/WormholeRelayerMock.sol
+++ b/contracts/mocks/crosschain/wormhole/WormholeRelayerMock.sol
@@ -1,0 +1,32 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.27;
+
+import {IWormholeRelayer} from "wormhole-solidity-sdk/interfaces/IWormholeRelayer.sol";
+import {IWormholeReceiver} from "wormhole-solidity-sdk/interfaces/IWormholeReceiver.sol";
+import {toUniversalAddress} from "wormhole-solidity-sdk/utils/UniversalAddress.sol";
+
+contract WormholeRelayerMock {
+    uint64 private _seq;
+
+    function sendPayloadToEvm(
+        uint16 targetChain,
+        address targetAddress,
+        bytes memory payload,
+        uint256 receiverValue,
+        uint256 gasLimit
+    ) external payable returns (uint64) {
+        // TODO: check that destination chain is local
+
+        uint64 seq = _seq++;
+        IWormholeReceiver(targetAddress).receiveWormholeMessages{value: receiverValue, gas: gasLimit}(
+            payload,
+            new bytes[](0),
+            toUniversalAddress(msg.sender),
+            targetChain,
+            keccak256(abi.encode(seq))
+        );
+
+        return seq;
+    }
+}
--- a/remappings.txt
+++ b/remappings.txt
@@ -2,3 +2,4 @@
 @openzeppelin/contracts-upgradeable/=lib/@openzeppelin-contracts-upgradeable/contracts/
 @openzeppelin/community-contracts/=contracts/
 @axelar-network/axelar-gmp-sdk-solidity/=node_modules/@axelar-network/axelar-gmp-sdk-solidity/
+wormhole-solidity-sdk/=lib/wormhole-solidity-sdk/src/
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,3 +9,6 @@
 [submodule "lib/forge-std"]
 	path = lib/forge-std
 	url = https://github.com/foundry-rs/forge-std.git
+[submodule "lib/wormhole-solidity-sdk"]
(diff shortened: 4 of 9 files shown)


Amxx:
TODO: fix lint rules to skip `lib/*`

remappings.txt
```diff
@@ -1,6 +1,7 @@
 @openzeppelin/contracts/=lib/@openzeppelin-contracts/contracts/
 @openzeppelin/contracts-upgradeable/=lib/@openzeppelin-contracts-upgradeable/contracts/
 @openzeppelin/community-contracts/=contracts/
-@axelar-network/axelar-gmp-sdk-solidity/=lib/axelar-gmp-sdk-solidity/
+@axelar-network/axelar-gmp-sdk-solidity/=node_modules/@axelar-network/axelar-gmp-sdk-solidity/
```
Amxx, line 4:
```suggestion
@axelar-network/axelar-gmp-sdk-solidity/=lib/axelar-gmp-sdk-solidity/
```

contracts/crosschain/wormhole/WormholeGatewayBase.sol
frangio, line 1: Let's do the same as #202 and flatten everything down to one contract.

contracts/crosschain/utils/ERC7786Attributes.sol
```diff
@@ -0,0 +1,34 @@
...
+pragma solidity ^0.8.27;
+
+import {IERC7786Attributes} from "../../interfaces/IERC7786Attributes.sol";
+
+/// @dev Library of helper to parse/process ERC-7786 attributes
+library ERC7786Attributes {
+    /// @dev Parse the `requestRelay(uint256,uint256,address)` (0x4cbb573a) attribute into its components.
+    function tryDecodeRequestRelay(
+        bytes memory attribute
+    ) internal pure returns (bool success, uint256 value, uint256 gasLimit, address refundRecipient) {
```
frangio, line 12: We're only using the calldata version in this PR. Can we keep only that one? I have a sense we'd never have attributes in memory so I'd also remove the `*Calldata` prefix from the name. (Perhaps we needed parsing from memory when we used to send attributes to the receiver.)

test/crosschain/utils/ERC7786Attributes.test.js
```diff
@@ -0,0 +1,50 @@
...
+const { loadFixture } = require('@nomicfoundation/hardhat-network-helpers');
+
+const ERC7786Attributes = require('../../helpers/erc7786attributes');
+
+async function fixture() {
+  const mock = await ethers.deployContract('$ERC7786Attributes');
+  return { mock };
+}
+
+// NOTE: here we are only testing the receiver. Failures of the gateway itself (invalid attributes, ...) are out of scope.
```
frangio, line 12:
```suggestion
```

contracts/mocks/crosschain/wormhole/WormholeRelayerMock.sol
```diff
@@ -0,0 +1,32 @@
...
+        uint256 gasLimit
+    ) external payable returns (uint64) {
+        // TODO: check that destination chain is local
+
+        uint64 seq = _seq++;
+        IWormholeReceiver(targetAddress).receiveWormholeMessages{value: receiverValue, gas: gasLimit}(
+            payload,
+            new bytes[](0),
+            toUniversalAddress(msg.sender),
+            targetChain,
```
frangio, line 26: This should be the source chain.

https://github.com/wormhole-foundation/wormhole-solidity-sdk/blob/575181b586a315d8f9813eab82e4cb98b45bc381/src/interfaces/IWormholeReceiver.sol#L48

Any reasonable way to do that in this mock?

contracts/mocks/crosschain/wormhole/WormholeRelayerMock.sol
```diff
@@ -0,0 +1,32 @@
...
+            keccak256(abi.encode(seq))
```
frangio, line 27:
```suggestion
            keccak256(abi.encode(seq)) // Unrealistic value but sufficient for current tests
```

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+import {InteroperableAddress} from "@openzeppelin/contracts/utils/draft-InteroperableAddress.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import {ERC7786Attributes} from "../utils/ERC7786Attributes.sol";
+import {IERC7786GatewaySource} from "../../interfaces/IERC7786.sol";
+import {IERC7786Receiver} from "../../interfaces/IERC7786.sol";
+import {IERC7786Attributes} from "../../interfaces/IERC7786Attributes.sol";
+
+/**
+ * @dev A contract that combines the functionality of both the source and destination gateway
+ * adapters for the Wormhole Network. Allowing to either send or receive messages across chains.
```
frangio, line 20:
```suggestion
 * @dev An ERC-7786 compliant adapter to send and receive messages via Wormhole.
```

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+ *
+ * Note: only EVM chains are currently supported
```
frangio, line 22: Is this because of address encoding?

frangio: Ah I see we're using `sendPayloadToEvm`.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint256 value = msg.value;
+        uint256 gasLimit = 0;
+
+        for (uint256 i = 0; i < attributes.length; ++i) {
+            (relay, value, gasLimit, ) = ERC7786Attributes.tryDecodeRequestRelayCalldata(attributes[i]);
+            require(relay, UnsupportedAttribute(attributes[i].length < 0x04 ? bytes4(0) : bytes4(attributes[i])));
+        }
+
+        if (relay) {
+            // TODO: Do we care about the returned "sequence"?
```
frangio, line 189: Can we use the returned sequence instead of a custom `sendId`?

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+        uint256 value,
+        uint256 gasLimit,
+        address /*refundRecipient*/
+    ) external view returns (uint256) {
+        (uint256 cost, ) = _wormholeRelayer.quoteEVMDeliveryPrice(getWormholeChain(recipient), value, gasLimit);
+        return cost - value;
+    }
+
+    /// @dev Relay a message that was initiated by {sendMessage}.
+    function requestRelay(bytes32 sendId, uint256 gasLimit, address /*refundRecipient*/) external payable {
```
frangio, line 237: Why is `refundRecipient` ignored? There's a version of `sendPayloadToEvm` that takes a refund address.

https://github.com/wormhole-foundation/wormhole-solidity-sdk/blob/575181b586a315d8f9813eab82e4cb98b45bc381/src/interfaces/IWormholeRelayer.sol#L123-L131

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+                recipient,
+                payload
+            );
+            _wormholeRelayer.sendPayloadToEvm{value: msg.value}(
+                getWormholeChain(recipient),
+                getRemoteGateway(recipient),
+                encoded,
+                value,
+                gasLimit
+            );
```
frangio, line 202: Ideally this would be in an internal function and reused in both `sendMessage` and `requestRelay`.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
frangio: Ah, we can't because we're using `sendId` for internal sequencing in the payload.

My point here was that we could return the sequence as the `sendId` from this function.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+        (uint256 cost, ) = _wormholeRelayer.quoteEVMDeliveryPrice(getWormholeChain(recipient), value, gasLimit);
+        return cost - value;
+    }
+
+    /// @dev Relay a message that was initiated by {sendMessage}.
+    function requestRelay(bytes32 sendId, uint256 gasLimit, address /*refundRecipient*/) external payable {
+        PendingMessage memory pmsg = _pending[sendId];
+        require(pmsg.pending, InvalidSendId(sendId));
+
+        // Do we want to do that to get a gas refund? Would it be valuable to keep that information stored?
```
frangio, line 241:
```suggestion
```

I think delete is fine.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+        delete _pending[sendId];
+
+        // TODO: Do we care about the returned "sequence"?
```
frangio, line 244:
```suggestion
```

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+
+        for (uint256 i = 0; i < attributes.length; ++i) {
+            (relay, value, gasLimit, ) = ERC7786Attributes.tryDecodeRequestRelayCalldata(attributes[i]);
+            require(relay, UnsupportedAttribute(attributes[i].length < 0x04 ? bytes4(0) : bytes4(attributes[i])));
+        }
+
+        if (relay) {
+            // TODO: Do we care about the returned "sequence"?
+            bytes memory encoded = abi.encode(
+                sendId,
```
frangio, line 191: ⚠️ This will always be 0!

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+    function sendMessage(
+        bytes calldata recipient, // Binary Interoperable Address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) external payable returns (bytes32 sendId) {
+        bool relay = false;
+        uint256 value = msg.value;
+        uint256 gasLimit = 0;
+
+        for (uint256 i = 0; i < attributes.length; ++i) {
```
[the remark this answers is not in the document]
Amxx: fixed

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
Amxx: fixed

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
Amxx: changed

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,325 @@
...
+ * Note: only EVM chains are currently supported
+ */
+// slither-disable-next-line locked-ether
+contract WormholeGatewayAdapter is IERC7786GatewaySource, IWormholeReceiver, Ownable {
+    using BitMaps for BitMaps.BitMap;
+    using InteroperableAddress for bytes;
+
+    IWormholeRelayer internal immutable _wormholeRelayer;
+    uint16 internal immutable _wormholeChainId;
+    uint24 private constant EVM_ID_MASK = 1 << 16;
```
frangio, line 30: I think `EVM_ID_FLAG` would be a clearer name for me.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,325 @@
...
+            // Parse the attribute details
+            (bool success, uint256 receiverValue, uint256 gasLimit, address refundRecipient) = ERC7786Attributes
+                .tryDecodeRequestRelay(attributes[0]);
+            require(success, InvalidAttributeEncoding(attributes[0]));
+
+            // Send the message
+            _sendMessage(msgId, recipient, payload, msg.sender, msg.value, receiverValue, gasLimit, refundRecipient);
+
+            emit MessageSent(
+                0,
```
frangio, line 220: If we have an id we might as well return it? It can be used to track the MessageRelayed event later.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,325 @@
...
+        // check message validity
+        // - `wormholeSourceAddress` is the remote gateway on the origin chain.
+        require(
+            addr == fromUniversalAddress(wormholeSourceAddress),
+            InvalidOriginGateway(wormholeSourceChain, wormholeSourceAddress)
+        );
+
+        // prevent replay - deliveryHash might not be unique if a message is relayed multiple time
+        require(!_executed[chainId].get(uint256(sendId)), MessageAlreadyExecuted(chainId, sendId));
+        _executed[chainId].set(uint256(sendId));
```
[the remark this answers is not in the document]
Amxx: I was not sure that if we try to re-relay, we would get the same message delivered twice with different deliveryHash. If we know for sure its not possible for a message relay retry to cause the same `_sendMessage` to arrive twice with different deliveryHashes, than yes, we can simplify things.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
frangio: Yes, returning an id means post processing _may_ be necessary.

> I would encourage returning zero is we have no usage for the value.

I was missing that if we return zero there won't be a MessageRelayed event. So it's true there is no usage for it in this case. Fine to leave it as is.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,325 @@
...
+        uint256 value,
+        uint256 gasLimit,
+        address /*refundRecipient*/
+    ) external view returns (uint256) {
+        (uint256 cost, ) = _wormholeRelayer.quoteEVMDeliveryPrice(getWormholeChain(recipient), value, gasLimit);
+        return cost - value;
+    }
+
+    /// @dev Relay a message that was initiated by {sendMessage}.
+    function requestRelay(bytes32 sendId, uint256 gasLimit, address refundRecipient) external payable {
```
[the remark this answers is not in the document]
frangio: That distinction makes some sense but it's implicit in the code and I didn't get it when I read it. In this function, `sendId` is implicitly used as `msgId`. I think the distinction is making the code slightly more confusing so I'd just call everything `sendId`.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
```diff
@@ -0,0 +1,296 @@
...
+        _wormholeChainId = wormholeChainId;
+    }
+
+    /// @dev Returns the local Wormhole relayer
+    function relayer() public view virtual returns (address) {
+        return address(_wormholeRelayer);
+    }
+
+    /// @dev Returns whether a binary interoperable chain id is supported.
+    function supportedChain(bytes memory chain) public view virtual returns (bool) {
```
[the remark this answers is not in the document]
frangio: I don't know, the problem is there is no such thing as a "binary interoperable chain id" as far as I know?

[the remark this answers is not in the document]
Amxx: How would you call an ERC7930 interroperable address that doesn't include an address ?

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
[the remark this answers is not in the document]
Amxx: > `sendId` is implicitly used as `msgId`

I'd put it differently. IMO `sendId` is always as `sendId`, and its in `sendMessage` that line 195 to 206 implicitelly say "we use `msgId` as the `sendId`"

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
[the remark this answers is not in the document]
Amxx: I'm not sure weither it would be possible for someone else to go to wormhole directly and ask for a re-relay, or a top-up, for a message that was initiated by the adatper.

In Axelar it is clearly possible for a third party to come and fund the execution.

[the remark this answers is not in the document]
frangio: There are [`resend` methods](https://wormhole.com/docs/products/messaging/reference/relayer-contract/#resend-generic). I don't understand the guarantees and can't find documentation. Let's leave it like it is now.

contracts/crosschain/wormhole/WormholeGatewayAdapter.sol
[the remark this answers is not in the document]
Amxx: In human readable format it reads like `@eip155:1#<checksum>` for mainnet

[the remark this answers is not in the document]
frangio: I see. Interesting.

PR #95, opened 2025-03-24
Update Account and Paymaster to use Entrypoint v0.8.0

Amxx:
Requires https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5586


contracts/account/Account.sol
```diff
@@ -20,36 +20,4 @@ import {AccountCore} from "./AccountCore.sol";
...
-                        userOp.accountGasLimits,
-                        userOp.preVerificationGas,
-                        userOp.gasFees,
-                        keccak256(userOp.paymasterAndData)
-                    )
-                )
-            );
-    }
-}
+abstract contract Account is AccountCore, ERC721Holder, ERC1155Holder, ERC7739 {}
```
ernestognw, line 22: I would consider removing `Account` whatsoever and just enable `ERC721Holder`, `ERC1155Holder`, `ERC721` and `ERC7739` as toggle features in Wizard.

We can keep this contract until we have the wizard version

contracts/account/AccountCore.sol
```diff
@@ -95,14 +95,15 @@ abstract contract AccountCore is AbstractSigner, IAccount {
     }
 
     /**
-     * @dev Virtual function that returns the signable hash for a user operations. Some implementation may return
-     * `userOpHash` while other may prefer a signer-friendly value such as an EIP-712 hash describing the `userOp`
-     * details.
+     * @dev Virtual function that returns the signable hash for a user operations. Since v0.8.0 of the entrypoint,
+     * `userOpHash` is a EIP-712 hash that can be signed directly.
```
ernestognw, line 99:
```suggestion
     * `userOpHash` is an EIP-712 hash that can be signed directly.
```

test/account/paymaster/PaymasterCore.test.js
```diff
@@ -68,6 +58,7 @@ for (const [name, opts] of Object.entries({
         );
 
     return {
+      helper,
```
ernestognw, line 61: Is this and other `helpers` returned from fixtures needed or just kept for convenience?

I'd remove them if they're not used.

test/helpers/erc4337.js
```diff
@@ -11,48 +11,51 @@ const parseInitCode = initCode => ({
...
   }
 
   async wait() {
-    return (this.env = await this.envAsPromise);
+    this.factory = await this.factoryAsPromise;
+    return this;
   }
 
+  // TODO: Pass the entrypoint in the params (senderCreator has to be discovered accordingly)
   async newAccount(name, extraArgs = [], params = {}) {
```
ernestognw, line 23: I'd just do:

```suggestion
  async newAccount(name, extraArgs = [], params = {
    entrypoint: entrypoint.v08,
    senderCreator: senderCreator.v08,
  }) {
```

ernestognw (approved):

contracts/account/AccountCore.sol
Amxx: AFAIK, we say "a hash" and not "an hash"

test/helpers/erc4337.js
Amxx: If the two are inconsitent we accept that behavior is unexpected ?

Honeslt we don't need that now. We don't have any test set-up with anything other than v08

contracts/account/Account.sol
Amxx: Note: this will require a lot of mocks changes, so I'd definitelly do it in a separate PR.

test/helpers/erc4337.js
Amxx: That syntax does not work. If a param object is provided that doesn't include entrypoint/senderCreator, then the default values are not added.

Implemented a bit differently

ernestognw: Perfect, I like the alternative better:

```js
const env = {
  entrypoint: params.entrypoint ?? entrypoint.v08,
  senderCreator: params.senderCreator ?? senderCreator.v08,
};
```

ernestognw (approved):

PR #96, opened 2025-03-25
Add ZKEmailSigner

ernestognw:
### Context

OpenZeppelin will collaborate with ZKEmail https://github.com/zkemail to provide developers with an easy way to opt-in for this functionality for an account.

This PR includes an adaptation of their [EmailSigner](https://github.com/zkemail/email-tx-builder/blob/main/packages/contracts/src/EmailSigner.sol) to be compatible with our `AbstractSigner`


Amxx:
Is there tooling that we could use to have unit tests ?

Amxx:
I noticed that this abstract signer is quite different from the other ones we already have:
- this abstractSigner requires further inheritance to set things such as the accountSalt(), DKIMRegistry() and verifier(). The other implementation we have storage everything they need (and have setters).
- this other abstractSigners are invisible from the outside, they only use private storage and internal functions. They don't expose anything publicly. This one exposes public functions.

This is not necessarily an issue, but its a pretty significant difference/inconsistency.

ernestognw:
> * this abstractSigner requires further inheritance to set things such as the accountSalt(), DKIMRegistry() and verifier(). The other implementation we have storage everything they need (and have setters).
> * this other abstractSigners are invisible from the outside, they only use private storage and internal functions. They don't expose anything publicly. This one exposes public functions.

Right, these are good observations. For the first point, I didn't implement those functions because I wasn't sure whether it's better to put them in storage or not (now I think it's fine in storage as long as developers can override, I'll update). 

For the second, I think the public functions are equivalent to the public key in each signer. For example, all signers have a public `signer()` getter, so I would say these public functions are a replace for `signer()`. Perhaps if this ZKEmailSigner exposes a `signer()` getter returning the `accountSalt`, `DKIMRegistry` and `verifier` would feel more consistent?

Amxx:
> For the second, I think the public functions are equivalent to the public key in each signer. For example, all signers have a public signer() getter, so I would say these public functions are a replace for signer(). 

Right.

I'm wondering if we should create a library that implements
```solidity
library ZKEmailUtils {
    function verifyEmail(
        bytes32 accountSalt,
        IDKIMRegistry registry,
        IVerifier verifier,
        uint256 commandTemplate,
        EmailAuthMsg memory emailAuthMsg
    ) internal view virtual returns (bool, EmailProofError) { ... }
}
```

(and maybe an override that takes a bytes calldata and decode it to a emailAuthMsg transparently

That can be used in the abstract signer, but also in a 7913 signature verifier. Basically the idea would be to separate the crypto logic from the "integration" logic.

contracts/utils/cryptography/ZKEmailSigner.sol
```diff
@@ -0,0 +1,126 @@
...
+            if (Strings.equal(expectedCommand, trimmedMaskedCommand)) {
+                break;
+            }
+            if (stringCase == 2) {
+                return (false, EmailProofError.Command);
+            }
+        }
+
+        if (!verifier().verifyEmailProof(emailAuthMsg.proof)) return (false, EmailProofError.EmailProof);
+        return (true, EmailProofError.NoError);
```
Amxx, line 124:
```suggestion
        for (uint256 stringCase = 0; stringCase < 2; stringCase++) {
            if (
                CommandUtils.computeExpectedCommand(
                    emailAuthMsg.commandParams,
                    signHashTemplate,
                    stringCase
                ).equal(trimmedMaskedCommand)
            ) {
                if (verifier().verifyEmailProof(emailAuthMsg.proof)) 
                    return (true, EmailProofError.NoError);
                else
                    return (false, EmailProofError.EmailProof);
            }
        }
        return (false, EmailProofError.Command);
```

Amxx, line 124:
```suggestion
        for (uint256 stringCase = 0; stringCase < 2; stringCase++) {
            if (
                CommandUtils.computeExpectedCommand(
                    emailAuthMsg.commandParams,
                    signHashTemplate,
                    stringCase
                ).equal(trimmedMaskedCommand)
            ) {
                if (verifier().verifyEmailProof(emailAuthMsg.proof)) 
                    return (true, EmailProofError.NoError);
                else
                    return (false, EmailProofError.EmailProof);
            }
        }
        return (false, EmailProofError.Command);
```

JohnGuilding (commented):
Implementation looks good in regards to abidance with our logic. Some comments on your discussion:

> I noticed that this abstract signer is quite different from the other ones we already have

Agree it makes sense to keep your signers consistent 

> Perhaps if this ZKEmailSigner exposes a signer() getter returning the accountSalt, DKIMRegistry and verifier would feel more consistent?

Nit: `signer()` feels like the wrong naming here for returning these values

> the public functions are equivalent to the public key in each signer.

Yeah there are equivalencies here to your signers, and then there also ZK Email-specific abstractions.

The “public keys” here would be the public keys stored in the DKIM registry, and also the `accountSalt`. The account salt is an implementation detail of ZK Email in order to provide onchain privacy. More info [here](https://github.com/zkemail/email-tx-builder/tree/main?tab=readme-ov-file#account-code-and-salt). We use the concept of an `accountSalt` in the `EmailAuth` circuits and ZK JWT circuits. Context: The `EmailAuth` circuits are the circuits we use to authenticate onchain actions with email

The `verifier` is kinda more analogous to the `P256` and `RSA` libs you have. The `commandTemplate` is an implementation detail of the specific circuits that this signer would use - the EmailAuth circuits. More detail [here](https://github.com/zkemail/email-tx-builder/tree/main?tab=readme-ov-file#command-template).

> I'm wondering if we should create a library that implements...

Open to the idea of a library. The thinking would then be that we would still store reference to the registry, verifier, salt etc in the signer @Amxx? Or something different

JohnGuilding:
@ernestognw merged a [fix](https://github.com/zkemail/email-tx-builder/pull/118) for the unused import. It's in `main` now

Amxx:
> Open to the idea of a library. The thinking would then be that we would still store reference to the registry, verifier, salt etc in the signer @Amxx? Or something different

It would be something like:
```solidity
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.20;

import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
import {IVerifier, EmailProof} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
import {EmailAuthMsg} from "@zk-email/email-tx-builder/interfaces/IEmailTypes.sol";
import {CommandUtils} from "@zk-email/email-tx-builder/libraries/CommandUtils.sol";
import {AbstractSigner} from "./AbstractSigner.sol";

library ZKEmailUtils {
    using Strings for string;

    enum EmailProofError {
        NoError,
        DKIMPublicKeyHash, // The DKIM public key hash verification fails
        MaskedCommandLength, // The masked command length exceeds the maximum
        SkippedCommandPrefixSize, // The skipped command prefix size is invalid
        Command, // The command format is invalid
        EmailProof // The email proof verification fails
    }

    function isValidZKEmail(
        EmailAuthMsg memory emailAuthMsg,
        IDKIMRegistry dkimregistry,
        IVerifier verifier
    ) internal view returns (bool, EmailProofError) {
        if (!dkimregistry.isDKIMPublicKeyHashValid(emailAuthMsg.proof.domainName, emailAuthMsg.proof.publicKeyHash)) {
            return (false, EmailProofError.DKIMPublicKeyHash);
        } else if (bytes(emailAuthMsg.proof.maskedCommand).length > verifier.commandBytes()) {
            return (false, EmailProofError.MaskedCommandLength);
        } else if (emailAuthMsg.skippedCommandPrefix >= verifier.commandBytes()) {
            return (false, EmailProofError.SkippedCommandPrefixSize);
        } else {
            string[] memory signHashTemplate = new string[](2);
            signHashTemplate[0] = "signHash";
            signHashTemplate[1] = "{uint}";
            
            // Construct an expectedCommand from template and the values of emailAuthMsg.commandParams.
            string memory expectedCommand = "";
            string memory trimmedMaskedCommand = CommandUtils.removePrefix(
                emailAuthMsg.proof.maskedCommand,
                emailAuthMsg.skippedCommandPrefix
            );
            for (uint256 stringCase = 0; stringCase < 2; stringCase++) {
                if (
                    CommandUtils.computeExpectedCommand(
                        emailAuthMsg.commandParams,
                        signHashTemplate,
                        stringCase
                    ).equal(expectedCommand, trimmedMaskedCommand)
                ) {
                    if (verifier.verifyEmailProof(emailAuthMsg.proof)) 
                        return (true, EmailProofError.NoError);
                    else 
                        return (false, EmailProofError.EmailProof);
                }
            }
            return (false, EmailProofError.Command);
        }
    }
}

contract ZKEmailSigner is AbstractSigner {
    using ZKEmailUtils for *;

    // Need setters for these
    bytes32 private _accountSalt;
    IDKIMRegistry private _registry;
    IVerifier private _verifier;
    uint256 private _commandTemplate;


    function accountSalt() public view virtual returns (bytes32) { return _accountSalt; }
    function DKIMRegistry() public view virtual returns (IDKIMRegistry) { return _registry; }
    function verifier() public view virtual returns (IVerifier) { return _verifier; }
    function commandTemplate() public view virtual returns (uint256) { return _commandTemplate; }

    function setAccountSalt(bytes32 accountSalt_) internal virtual { _accountSalt = accountSalt_; }
    function setDKIMRegistry(IDKIMRegistry registry_) internal virtual { _registry = registry_; }
    function setVerifier(IVerifier verifier_) internal virtual { _verifier = verifier_; }
    function setCommandTemplate(uint256 commandTemplate_) internal virtual { _commandTemplate = commandTemplate_; }

    /// @inheritdoc AbstractSigner
        bytes32 hash,
        bytes calldata signature
        EmailAuthMsg memory emailAuthMsg = abi.decode(signature, (EmailAuthMsg));
        if (
            emailAuthMsg.commandParams[0] == hash &&
            emailAuthMsg.templateId == commandTemplate() &&
            emailAuthMsg.proof.accountSalt == accountSalt()
        ) {
            (bool verified, ) = emailAuthMsg.isValidZKEmail(DKIMRegistry(), verifier());
            return verified;
        } else {
            return false;
        }
    }
}
```

JohnGuilding:
> It would be something like:...

Yeah I'm a fan of the approach outlined.

This signer would still differ from your existing signers, as it would not contain the exact functions `_setSigner` or `signer()`. Is this ok for you or would you want to change something else here?

The abstractions are quite different so it would feel strange to return the 4 zkemail-specific variables from a function called `signer()` imo.

package.json
```diff
@@ -17,10 +17,10 @@
     "prepare-docs": "scripts/prepare-docs.sh",
     "lint": "npm run lint:js && npm run lint:sol",
     "lint:fix": "npm run lint:js:fix && npm run lint:sol:fix",
-    "lint:js": "prettier --log-level warn --ignore-path .gitignore '**/*.{js,ts}' --check && eslint .",
-    "lint:js:fix": "prettier --log-level warn --ignore-path .gitignore '**/*.{js,ts}' --write && eslint . --fix",
-    "lint:sol": "prettier --log-level warn --ignore-path .gitignore '{contracts,test}/**/*.sol' --check && solhint '{contracts,test}/**/*.sol'",
-    "lint:sol:fix": "prettier --log-level warn --ignore-path .gitignore '{contracts,test}/**/*.sol' --write",
```
ernestognw, line 23: According to `prettier --help`, the default value already is `[.gitignore, .prettierignore]`

contracts/utils/cryptography/ZKEmailSigner.sol
```diff
@@ -0,0 +1,126 @@
...
+        (bool verified, ) = _verifyEmail(emailAuthMsg);
+        return verified && abi.decode(emailAuthMsg.commandParams[0], (bytes32)) == hash;
+    }
+
+    /// @dev Internal function to verify an email authenticated message that doesn't revert and returns a boolean and the error instead.
+    function _verifyEmail(EmailAuthMsg memory emailAuthMsg) internal view virtual returns (bool, EmailProofError) {
+        if (commandTemplate() != emailAuthMsg.templateId) return (false, EmailProofError.CommandTemplate);
+        string[] memory signHashTemplate = new string[](2);
+        signHashTemplate[0] = "signHash";
+        signHashTemplate[1] = "{uint}";
```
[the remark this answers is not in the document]
ernestognw: It makes sense to use constants as part of the library. For the `"{uint}"` one, I'm using the constant from your [`CommandUtils` library](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96/files#diff-2ec326e3c739b9eb617929730d8a8c6500f0ae5d8be920140ade5ada9345cdb8R38).

Perhaps we add an `template` input to the `isValidZKEmail`?

```
function isValidZKEmail(
        EmailAuthMsg memory emailAuthMsg,
        IDKIMRegistry dkimregistry,
        IVerifier verifier,
        string memory template
    ) internal view returns (bool, EmailProofError) {
        if (!dkimregistry.isDKIMPublicKeyHashValid(emailAuthMsg.proof.domainName, emailAuthMsg.proof.publicKeyHash)) {
            return (false, EmailProofError.DKIMPublicKeyHash);
        } else if (bytes(emailAuthMsg.proof.maskedCommand).length > verifier.commandBytes()) {
            return (false, EmailProofError.MaskedCommandLength);
        } else if (emailAuthMsg.skippedCommandPrefix >= verifier.commandBytes()) {
            return (false, EmailProofError.SkippedCommandPrefixSize);
        } else {
            // Construct an expectedCommand from template and the values of emailAuthMsg.commandParams.
            string memory trimmedMaskedCommand = CommandUtils.removePrefix(
                emailAuthMsg.proof.maskedCommand,
                emailAuthMsg.skippedCommandPrefix
            );
            for (uint256 stringCase = 0; stringCase < 2; stringCase++) {
                if (
                    CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, stringCase).equal(
                        trimmedMaskedCommand
                    )
                ) {
                    if (verifier.verifyEmailProof(emailAuthMsg.proof)) return (true, EmailProofError.NoError);
                    else return (false, EmailProofError.EmailProof);
                }
            }
            return (false, EmailProofError.Command);
        }
    }
```

In this way the implementation can construct the template out of constants. One of them being `signHash`.

ernestognw:
> (and maybe an override that takes a bytes calldata and decode it to a emailAuthMsg transparently

The issue with this is that a library can't have virtual functions.  So there's no way to override it.

> That can be used in the abstract signer, but also in a 7913 signature verifier. Basically the idea would be to separate the crypto logic from the "integration" logic.

Yes, I agree that a library would be the best way to abstract the logic. I updated the PR accordingly.

> The “public keys” here would be the public keys stored in the DKIM registry, and also the accountSalt. The account salt is an implementation detail of ZK Email in order to provide onchain privacy. More info [here](https://github.com/zkemail/email-tx-builder/tree/main?tab=readme-ov-file#account-code-and-salt).

Added this as [context to the current `accountSalt()` function NatSpec](https://github.com/OpenZeppelin/openzeppelin-community-contracts/commit/813b226136740b055801ae90ec80e63db023f27e#diff-7f09cf6e3ebdf9e6b443ae44df8cc11d59fbe8a832155d5b2d9d68afac9e8bc4R26-R32). I think it's important so that developers know how to override it if they want.

Amxx:
> I guess @Amxx point is that the ZKEmailUtils.isValidZKEmail may receive an IERC7913 instance instead of an IVerifier one

That was not what I had in mind

package.json
Amxx: we should also change that on the vanila repo

contracts/utils/cryptography/ZKEmailSigner.sol
```diff
@@ -0,0 +1,101 @@
...
+        _commandTemplate = commandTemplate_;
+    }
+
+    /** @dev Authenticate the email sender and authorize the message in the email command.
+     *
+     * NOTE: This function only verifies the authenticity of the email and command, without
+     * handling replay protection. The calling contract should implement its own mechanisms
+     * to prevent replay attacks, similar to how nonces are used with ECDSA signatures.
+     */
+    function verifyEmail(EmailAuthMsg memory emailAuthMsg) public view virtual {
```
Amxx, line 79: Do we want that function ? I'm not sure we really do.

Note that it doesn't verify the email corresponds to the commandTemplate and the accountSalt for this signer.

Amxx:
> Is there tooling that we could use to have unit tests ?

Any news on that part. Code looks quite good.

JohnGuilding:
> Any news on that part. Code looks quite good.

You can reference [these](https://github.com/zkemail/email-tx-builder/blob/main/packages/contracts/test/EmailSigner.t.sol) for unit testing. Although they are forge tests not hardhat tests

contracts/utils/cryptography/ZKEmailSigner.sol
```diff
@@ -0,0 +1,101 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {IERC1271} from "@openzeppelin/contracts/interfaces/IERC1271.sol";
```
Amxx, line 5:
```suggestion
```

contracts/utils/cryptography/ZKEmailSigner.sol
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /// @inheritdoc AbstractSigner
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        EmailAuthMsg memory emailAuthMsg = abi.decode(signature, (EmailAuthMsg));
+        if (
+            abi.decode(emailAuthMsg.commandParams[0], (bytes32)) == hash &&
```
Amxx, line 89: This would avoid some soldiity revert and handle them more gracefully (return false)
```suggestion
            emailAuthMsg.commandParams.length > 0 &&
            emailAuthMsg.commandParams[0].length >= 0x20 &&
            bytes32(emailAuthMsg.commandParams[0]) == hash &&
```

Amxx: But in any case the `EmailAuthMsg memory emailAuthMsg = abi.decode(signature, (EmailAuthMsg));` part can always fail. So many its not worth it.

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,58 @@
...
+            string[] memory signHashTemplate = new string[](2);
+            signHashTemplate[0] = "signHash";
+            signHashTemplate[1] = CommandUtils.UINT_MATCHER;
+
+            // Construct an expectedCommand from template and the values of emailAuthMsg.commandParams.
+            string memory trimmedMaskedCommand = CommandUtils.removePrefix(
+                emailAuthMsg.proof.maskedCommand,
+                emailAuthMsg.skippedCommandPrefix
+            );
+            for (uint256 stringCase = 0; stringCase < 2; stringCase++) {
```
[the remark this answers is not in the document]
Amxx: So we can avoid doing the loop and only need to check one index ?

contracts/utils/cryptography/ZKEmailSigner.sol
Amxx: In what situation will that function be used ?

contracts/utils/cryptography/ZKEmailUtils.sol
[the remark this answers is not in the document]
zkfriendly: correct.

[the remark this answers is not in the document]
JohnGuilding: Good catch @zkfriendly. So we could do something like this?

```solidity
        string memory trimmedMaskedCommand = CommandUtils.removePrefix(
            emailAuthMsg.proof.maskedCommand,
            emailAuthMsg.skippedCommandPrefix
        );
        uint256 stringCase = 0;
        string memory expectedCommand = CommandUtils.computeExpectedCommand(
            emailAuthMsg.commandParams,
            signHashTemplate,
            stringCase
        );
        if (trimmedMaskedCommand != expectedCommand) {
            revert InvalidCommand();
        }
```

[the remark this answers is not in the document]
zkfriendly: Yes. Computing expected command with `stringCase` 0 will produce the expected command correctly.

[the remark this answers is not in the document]
ernestognw: Thanks for pointing it out @zkfriendly! Given @JohnGuilding [suggested to make the command a constant](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96#discussion_r2023855878), I was thinking there could be 2 functions:

```solidity
function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier
) internal view returns (EmailProofError) {
  string[] memory signHashTemplate = new string[](2);
  signHashTemplate[0] = "signHash";
  signHashTemplate[1] = CommandUtils.UINT_MATCHER;
  return isValidZKEmail(emailAuthMsg, dkimregistry, verifier, signHashTemplate);
}

function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier,
    string[] memory template
) internal view returns (EmailProofError) {
  ...
}
```

If that were the case, I think the loop would be needed, right?
The rationale behind these 2 functions is that developers who want to use the `signHash` template by default shouldn't be providing it as an argument.

JohnGuilding:
> You seem convinced that we should keep the signers consistency in this regard, curious to hear your thoughts.

No strong opinion on consistency - happy to defer to you guys on what `AbstractSigner` **should** expose. If you guys are happy with exposing those 4 variables and not having the signer functions, then that sounds good to me

ernestognw:
> > You seem convinced that we should keep the signers consistency in this regard, curious to hear your thoughts.
> 
> No strong opinion on consistency - happy to defer to you guys on what `AbstractSigner` **should** expose. If you guys are happy with exposing those 4 variables and not having the signer functions, then that sounds good to me

Got it. I am also fine exposing these 4 functions instead of a `signer`, we can't guarantee such external function would be there if it's not standardized, so I don't see any negative impact.

contracts/utils/cryptography/ZKEmailUtils.sol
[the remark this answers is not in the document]
zkfriendly: Interesting idea.

> If that were the case, I think the loop would be needed, right?

In that case, yeah, we would need to keep the loop.

It's worth noting that `templateId` (or [`_commandTemplate` here](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/4b6426c08e9981cbf7d266bd10c986ad7083cce7/contracts/utils/cryptography/SignerZKEmail.sol#L50) — maybe `templateId` is a more accurate name) is also part of the proof and should be checked to match the expected `templateId`. The idea is that `templateId` acts like a namespace, so even if two different contracts use the same template, a proof from one app can't be reused on another.

I noticed that `templateId` isn't being checked in the `isValidZKEmail` library function. That might be fine if the calling function is handling the check — like [`_rawSignatureValidation`](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/4b6426c08e9981cbf7d266bd10c986ad7083cce7/contracts/utils/cryptography/SignerZKEmail.sol#L139) does. But yeah, just pointing this out for more context — you'd basically need some kind of mapping between registered `templateId`s and the actual templates and check them if reply protection is a concern.

Also just noticed that [`verifyEmail`](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/4b6426c08e9981cbf7d266bd10c986ad7083cce7/contracts/utils/cryptography/SignerZKEmail.sol#L119) isn’t doing the `templateId` check either.

contracts/utils/cryptography/ZKEmailSigner.sol
ernestognw: I think it makes sense to fail more gracefully. If the signer is part of an account this would go through the Entrypoint, and it may revert if a wrong `commandParams` is sent.

contracts/utils/cryptography/ZKEmailSigner.sol
Amxx: I'm really not convinced. 

Userop simulation is done using a standard process that starts at the entrypoint. It's not done at the account level.

Other signatures will be checked using 1271 endpoint (if available)

contracts/utils/cryptography/ZKEmailUtils.sol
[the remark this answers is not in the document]
zkfriendly: > Is there any guideline to pick a templateId? 

No particular guideline for this. It only needs to be unique. This is an example of how we derive the `templateId` in [`EmailSignerFactory` ](https://github.com/zkemail/email-tx-builder/blob/db9b113165c822c716ef6e72a573d9504428b60d/packages/contracts/src/EmailSignerFactory.sol#L56):

```solidity
uint256 templateId = uint256(
    keccak256(abi.encodePacked("EMAIL-SIGNER", accountSalt))
);
```

> still not sure if we'll need that function

I personally find this to be useful for the reason you mentioned, just checking to see if a proof will pass or not has come in handy before.

contracts/utils/cryptography/ZKEmailSigner.sol
Amxx: Entrypoint is protected against the account reverting.

ernestognw: Yes but reverting will affect reputation, so we want to avoid it, don't we?

contracts/utils/cryptography/ZKEmailUtils.sol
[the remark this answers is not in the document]
Amxx: @zkfriendly can you confirm this is correct https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96/commits/11b197ffa05251133137ae5b184856a5d64ea4bb?

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,96 @@
...
+    ) internal view returns (EmailProofError) {
+        if (!dkimregistry.isDKIMPublicKeyHashValid(emailAuthMsg.proof.domainName, emailAuthMsg.proof.publicKeyHash)) {
+            return EmailProofError.DKIMPublicKeyHash;
+        } else if (bytes(emailAuthMsg.proof.maskedCommand).length > verifier.commandBytes()) {
+            return EmailProofError.MaskedCommandLength;
+        } else if (emailAuthMsg.skippedCommandPrefix >= verifier.commandBytes()) {
+            return EmailProofError.SkippedCommandPrefixSize;
+        } else if (
+            !CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, 0).equal(
+                CommandUtils.removePrefix(emailAuthMsg.proof.maskedCommand, emailAuthMsg.skippedCommandPrefix)
```
Amxx, line 86: Note: `CommandUtils.removePrefix` is equivalent (but less efficient) to `Bytes.slice` for the openzeppelin/contracts

Amxx: `removePrefix` uses a for loop, `slice` does a `mcopy`.

contracts/utils/cryptography/ZKEmailUtils.sol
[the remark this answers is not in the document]
zkfriendly: > can you confirm if this is correct? https://github.com/OpenZeppelin/openzeppelin-community-contracts/commit/11b197ffa05251133137ae5b184856a5d64ea4bb?

Now that we have two `isValidZKEmail` functions—one with the `signHash` template hardcoded and the other that takes the `template` as an argument—we need to keep the loop. If we only want to support one template (`signHash`), we can safely remove the loop. However, if we want to support dynamic template verification as well, we'll need to keep the loop, because it's now possible for a template to use the ETH address matcher.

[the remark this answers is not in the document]
ernestognw: Another option is to have 3 `isValidZKEmailFunctions`, one that uses `signHash` by default, another that receives a `template`, and a third one that receives a `template` and a `stringCase`.

```solidity
enum Case {
    LOWERCASE, // Converts the command to hex lowercase.
    UPPERCASE, // Converts the command to hex uppercase.
    CHECKSUM // Computes a checksum of the command.
}

// signHash by default
function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier
) internal view returns (EmailProofError)  {
    string[] memory signHashTemplate = new string[](2);
    signHashTemplate[0] = "signHash";
    signHashTemplate[1] = CommandUtils.UINT_MATCHER;

    return isValidZKEmail(emailAuthMsg, dkimregistry, verifier, signHashTemplate, Case.LOWERCASE);
}

// custom template - try all stringCases
function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier,
    string[] memory template
) internal view returns (EmailProofError) {
    EmailProofError err = _validateEmailAuthParams(emailAuthMsg, dkimregistry, verifier);
    for (uint256 i = 0; i < uint8(type(Case).max) && err != EmailProofErrror.NoError; i++) {
        err = _validateCommandAndProof(emailAuthMsg, verifier, template, Case(i));
    }
    return err;
}

// custom template - try only specific stringCase
function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    Case stringCase
) internal view returns (EmailProofError)  {
    EmailProofError err = _validateEmailAuthParams(emailAuthMsg, dkimregistry, verifier);
    return
        err == EmailProofError.NoError
            ? _validateCommandAndProof(emailAuthMsg, verifier, template, stringCase)
            : err;
}
```

<details>
<summary>Code for </summary>
```solidity
function isValidZKEmail(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier,
    string[] memory template,
    Case stringCase
) internal view returns (EmailProofError) {
    EmailProofError err = _validateEmailAuthParams(emailAuthMsg, dkimregistry, verifier);
    return
        err == EmailProofError.NoError
            ? _validateCommandAndProof(emailAuthMsg, verifier, template, stringCase)
            : err;
}

function _validateEmailAuthParams(
    EmailAuthMsg memory emailAuthMsg,
    IDKIMRegistry dkimregistry,
    IVerifier verifier
) private view returns (EmailProofError) {
    if (!dkimregistry.isDKIMPublicKeyHashValid(emailAuthMsg.proof.domainName, emailAuthMsg.proof.publicKeyHash)) {
        return EmailProofError.DKIMPublicKeyHash;
    } else if (bytes(emailAuthMsg.proof.maskedCommand).length > verifier.commandBytes()) {
        return EmailProofError.MaskedCommandLength;
    } else if (emailAuthMsg.skippedCommandPrefix >= verifier.commandBytes()) {
        return EmailProofError.SkippedCommandPrefixSize;
    }
    return EmailProofError.NoError;
}
```
<details>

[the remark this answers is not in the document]
ernestognw: Tried the previous idea in [this commit](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96/commits/3e9d0e33cadf29a2ed0d8c4eb2db238af4d707ec). I like how it looks.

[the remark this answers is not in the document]
Amxx: The loop will override any error detected at an earlier stage.

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,144 @@
...
+        EmailAuthMsg memory emailAuthMsg,
+        IDKIMRegistry dkimregistry,
+        IVerifier verifier,
+        string[] memory template
+    ) internal view returns (EmailProofError) {
+        EmailProofError err = validateDKIMAndCommandFormat(emailAuthMsg, dkimregistry, verifier);
+        for (uint256 i = 0; i < uint8(type(Case).max) && err != EmailProofError.NoError; i++) {
+            err = validateExpectedCommandAndProof(emailAuthMsg, verifier, template, Case(i));
+        }
+        return err;
```
Amxx, line 82: If `validateDKIMAndCommandFormat` return an error, this will ignore it

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,144 @@
...
+    ) internal view returns (EmailProofError) {
+        string memory expectedCommand = CommandUtils.computeExpectedCommand(
+            emailAuthMsg.commandParams,
+            template,
+            uint8(stringCase)
+        );
+        if (!expectedCommand.equal(emailAuthMsg.proof.maskedCommand)) {
+            return EmailProofError.MismatchedCommand;
+        }
+        return verifier.verifyEmailProof(emailAuthMsg.proof) ? EmailProofError.NoError : EmailProofError.EmailProof;
```
Amxx, line 142: What is the point of doing that inside the `stringCase` loop ? It doesn't depend on the loop.

contracts/utils/cryptography/ZKEmailUtils.sol
Amxx: I think its even worst. If `validateDKIMAndCommandFormat` return `NoError` then the loop condition is never true, and you never try any of the loop values ... and you return the `NoError` value comming from above.

ernestognw: Yes, there was a check missing, the correct function should look like:

```solidity
EmailProofError err = validateDKIMAndCommandFormat(emailAuthMsg, dkimregistry, verifier);
if (err != EmailProofError.NoError) return err;
for (uint256 i = 0; i < uint8(type(Case).max) && err != EmailProofError.NoError; i++) {
    err = validateExpectedCommandAndProof(emailAuthMsg, verifier, template, Case(i));
}
return err;
```

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,145 @@
...
+     */
+    function isValidZKEmail(
+        EmailAuthMsg memory emailAuthMsg,
+        IDKIMRegistry dkimregistry,
+        IVerifier verifier,
+        string[] memory template
+    ) internal view returns (EmailProofError) {
+        EmailProofError err = validateDKIMAndCommandFormat(emailAuthMsg, dkimregistry, verifier);
+        if (err != EmailProofError.NoError) return err;
+        for (uint256 i = 0; i < uint8(type(Case).max) && err != EmailProofError.NoError; i++) {
```
Amxx, line 80: That change is still broken. If `err` is `NoError` you don't enter the loop.

ernestognw: I'll try the alternative approach with `_commandMatch`. Next time please write a suggestion, it's difficult to follow "if this then that, but then this, otherwise we should do this other", I understand the function is broken.

contracts/utils/cryptography/ZKEmailUtils.sol
ernestognw: I don't understand this question

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,131 @@
...
+    function _commandMatch(
+        EmailAuthMsg memory emailAuthMsg,
+        string[] memory template,
+        Case stringCase
+    ) private pure returns (bool) {
+        return
+            stringCase == Case.ANY
+                ? _matchAnyCase(emailAuthMsg, template)
+                : _matchCase(emailAuthMsg, template, stringCase);
+    }
```
Amxx, line 112: Would it be worth inlining this directly in `isValidZKEmail` ?

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,131 @@
...
+
+    function _matchAnyCase(EmailAuthMsg memory emailAuthMsg, string[] memory template) private pure returns (bool) {
+        return
+            _matchCase(emailAuthMsg, template, Case.LOWERCASE) ||
+            _matchCase(emailAuthMsg, template, Case.UPPERCASE) ||
+            _matchCase(emailAuthMsg, template, Case.CHECKSUM);
+    }
+
+    function _matchCase(
```
Amxx, line 121: This function should never be called with ANY. Its private, so this is fine, but maybe we should document that part

contracts/utils/cryptography/ZKEmailUtils.sol
ernestognw: It's either this:

```solidity
       else if (
            !(
                stringCase == Case.ANY
                    ? _matchAnyCase(emailAuthMsg, template)
                    : _matchCase(emailAuthMsg, template, stringCase)
            )
        ) {
            return EmailProofError.MismatchedCommand;
        } else {
            return verifier.verifyEmailProof(emailAuthMsg.proof) ? EmailProofError.NoError : EmailProofError.EmailProof;
        }
```

Or this:

```solidity
else if (
      stringCase == Case.ANY
          ? _matchAnyCase(emailAuthMsg, template)
          : _matchCase(emailAuthMsg, template, stringCase)
  ) {
      return verifier.verifyEmailProof(emailAuthMsg.proof) ? EmailProofError.NoError : EmailProofError.EmailProof;
  } else {
      return EmailProofError.MismatchedCommand;
  }
```

And I wasn't a fan of either

Amxx: I don't see anything wrong, but I guess this is personal preference.

Another option would be 
```solidity
    function _commandMatch(
        EmailAuthMsg memory emailAuthMsg,
        string[] memory template,
        Case stringCase
    ) private pure returns (bool) {
        return
            stringCase == Case.ANY
            ?
                CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.LOWERCASE)).equal(emailAuthMsg.proof.maskedCommand) ||
                CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.UPPERCASE)).equal(emailAuthMsg.proof.maskedCommand) ||
                CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.CHECKSUM)).equal(emailAuthMsg.proof.maskedCommand)
            :
                CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(stringCase)).equal(emailAuthMsg.proof.maskedCommand);
    }
```

Which IMO would be the best, except that the linter rewrites it as 
```solidity
    function _commandMatch(
        EmailAuthMsg memory emailAuthMsg,
        string[] memory template,
        Case stringCase
    ) private pure returns (bool) {
        return
            stringCase == Case.ANY
                ? CommandUtils
                    .computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.LOWERCASE))
                    .equal(emailAuthMsg.proof.maskedCommand) ||
                    CommandUtils
                        .computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.UPPERCASE))
                        .equal(emailAuthMsg.proof.maskedCommand) ||
                    CommandUtils
                        .computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(Case.CHECKSUM))
                        .equal(emailAuthMsg.proof.maskedCommand)
                : CommandUtils.computeExpectedCommand(emailAuthMsg.commandParams, template, uint8(stringCase)).equal(
                    emailAuthMsg.proof.maskedCommand
                );
    }
```

which is horrible

ernestognw: Yeah that's horrible. Let's do this one:

```
else if (
      stringCase == Case.ANY
          ? _matchAnyCase(emailAuthMsg, template)
          : _matchCase(emailAuthMsg, template, stringCase)
  ) {
      return verifier.verifyEmailProof(emailAuthMsg.proof) ? EmailProofError.NoError : EmailProofError.EmailProof;
  } else {
      return EmailProofError.MismatchedCommand;
  }
```

Amxx: This is linted compatible, and might actually be cheaper because of the mload caching
@ernestognw WDYT? 

```solidity
    function _commandMatch(
        EmailAuthMsg memory emailAuthMsg,
        string[] memory template,
        Case stringCase
    ) private pure returns (bool) {
        string memory commandParams = emailAuthMsg.commandParams; // Not a memory copy
        string memory maskedCommand = emailAuthMsg.maskedCommand; // Not a memory copy
        return
            stringCase == Case.ANY
                ? (commandParams.computeExpectedCommand(template, uint8(Case.LOWERCASE)).equal(proof.maskedCommand) ||
                    commandParams.computeExpectedCommand(template, uint8(Case.UPPERCASE)).equal(proof.maskedCommand) ||
                    commandParams.computeExpectedCommand(template, uint8(Case.CHECKSUM)).equal(proof.maskedCommand))
                : commandParams.computeExpectedCommand(template, uint8(stringCase)).equal(proof.maskedCommand);
    }
```

ernestognw:
Hey @JohnGuilding and @zkfriendly. Adding you as CODEOWNERS of the library and signer. This way you would get notified if new contributions come in. Let me know if I should add/remove someone

contracts/utils/cryptography/SignerZKEmail.sol
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * @dev See {AbstractSigner-_rawSignatureValidation}. Validates a raw signature by:
+     *
+     * 1. Decoding the email authentication message from the signature
+     * 2. Verifying the hash matches the command parameters
+     * 3. Checking the template ID matches
+     * 4. Validating the account salt
+     * 5. Verifying the email proof
+     */
+    function _rawSignatureValidation(
```
JohnGuilding, line 136: What is the rationale for having `verifyEmail` and `_rawSignatureValidation`?

contracts/mocks/docs/account/MyAccountZKEmail.sol
JohnGuilding, line 1: For my understanding, what is the difference between `MyAccountZKEmail` and `AccountZKEmailMock`? Can see the mock is used in testing but unsure of the purpose of `MyAccountZKEmail`

contracts/utils/cryptography/SignerZKEmail.sol
JohnGuilding: Duplicate question to https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96#discussion_r2026793241

contracts/utils/cryptography/ZKEmailSigner.sol
JohnGuilding: > I was thinking to verify that the EmailAuthMsg is valid externally. For example, a frontend can call this function before executing a user op

Could another option be to follow the same pattern as your ECDSA library? And have a `verifyEmail` and a `tryVerifyEmail`? And keeping that behaviour in the ZKEmailUtils instead of this signer

contracts/utils/cryptography/SignerZKEmail.sol
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * @dev Unique identifier for owner of this contract defined as a hash of an email address and an account code.
+     *
+     * An account code is a random integer in a finite scalar field of https://neuromancer.sk/std/bn/bn254[BN254] curve.
+     * It is a private randomness to derive a CREATE2 salt of the user's Ethereum address
+     * from the email address, i.e., userEtherAddr := CREATE2(hash(userEmailAddr, accountCode)).
+     *
+     * The account salt is used for:
+     *
+     * * User Identification: Links the email address to a specific Ethereum address securely and deterministically.
```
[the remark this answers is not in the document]
ernestognw: This makes sense. Updating

contracts/utils/cryptography/SignerZKEmail.sol
```diff
@@ -0,0 +1,146 @@
...
+import {AbstractSigner} from "./AbstractSigner.sol";
+import {ZKEmailUtils} from "./ZKEmailUtils.sol";
+
+/**
+ * @dev Implementation of {AbstractSigner} using https://docs.zk.email[ZKEmail] signatures.
+ *
+ * ZKEmail enables secure authentication and authorization through email messages, leveraging
+ * DKIM signatures from a trusted {DKIMRegistry} and zero-knowledge proofs enabled by a {verifier}
+ * contract that ensures email authenticity without revealing sensitive information. This contract
+ * implements the core functionality for validating email-based signatures in smart contracts.
```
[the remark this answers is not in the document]
ernestognw: Pretty well written, adding it without changes. Thanks!

contracts/mocks/docs/account/MyAccountZKEmail.sol
ernestognw: Good point. The difference is that `MyAccountZKEmail` is used for the documentation. Here's a preview:
https://deploy-preview-96--community-contracts.netlify.app/community-contracts/0.0.1/account-abstraction#selecting_a_signer

You'll notice that the mock uses a constructor, which is not possible if the account were to be used in ERC-4337 infrastructure where it needs to be initialized by a factory. So the docs example is suitable for that case

contracts/utils/cryptography/ZKEmailSigner.sol
ernestognw: I think the `try` option is useful when the verification function may revert. In this case it returns an enum error so it's enough IMO. I'm removing this function from the signer for now, I think we mostly prefer not to have it

contracts/utils/cryptography/SignerZKEmail.sol
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * @dev Authenticates the email sender and authorizes the message in the email command.
+     *
+     * NOTE: This function only verifies the authenticity of the email and command, without
+     * handling replay protection. The calling contract should implement its own mechanisms
+     * to prevent replay attacks, similar to how nonces are used with ECDSA signatures.
+     */
+    function verifyEmail(EmailAuthMsg memory emailAuthMsg) public view virtual {
+        if (emailAuthMsg.templateId != templateId() || emailAuthMsg.proof.accountSalt != accountSalt()) {
+            revert InvalidEmailProof(ZKEmailUtils.EmailProofError.EmailProof);
```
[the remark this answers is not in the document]
ernestognw: I removed the function. We may consider passing some of this validation down to the library, perhaps extending the `EmailProofError` if you think is worth

contracts/utils/cryptography/SignerZKEmail.sol
ernestognw: Initially the rationale was just having an external verification function, but it's true that's an implementation detail and `_rawSignatureValidation` must be enough. I removed the `verifyEmail` function

contracts/mocks/docs/account/MyAccountZKEmail.sol
JohnGuilding: understood thanks

contracts/utils/cryptography/SignerZKEmail.sol
JohnGuilding: Got it, sounds good

contracts/utils/cryptography/SignerZKEmail.sol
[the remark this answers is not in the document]
JohnGuilding: I'm not super super opinionated. There would be a small devex improvement to using the utils library as you could get specific errors for invalid template ids and salts. However, it would make the interfaces less simple.

Outlining rough changes:

```solidity
        // signer 
        ...
        return (
            abi.decode(emailAuthMsg.commandParams[0], (bytes32)) == hash &&
            emailAuthMsg.isValidZKEmail(
                DKIMRegistry(),
                verifier(),
                accountSalt(),
                templateId()
            ) == ZKEmailUtils.EmailProofError.NoError);
        ...
```

```solidity
        // utils  
        ...
        
        enum EmailProofError {
            NoError,
            InvalidTemplateId, // ...
            InvalidAccountSalt, // ...
            DKIMPublicKeyHash, // The DKIM public key hash verification fails
            MaskedCommandLength, // The masked command length exceeds the maximum
            SkippedCommandPrefixSize, // The skipped command prefix size is invalid
            MismatchedCommand, // The command does not match the proof command
            EmailProof // The email proof verification fails
        }
        
        function isValidZKEmail(
            EmailAuthMsg memory emailAuthMsg,
            IDKIMRegistry dkimregistry,
            IVerifier verifier,
            bytes32 accountSalt,
            uint256 templateId
        ) internal view returns (EmailProofError) {
        ...
```

I would say I'm leaning slightly towards doing this but not strongly in favour. Wdyt?

contracts/utils/cryptography/SignerZKEmail.sol
```diff
@@ -0,0 +1,133 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.24;
+
+import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
+import {IVerifier, EmailProof} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
```
benceharomi, line 6:
```suggestion
import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
```
`EmailProof` is an unused import

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,120 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.24;
+
+import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
+import {IVerifier, EmailProof} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
```
benceharomi, line 8:
```suggestion
import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
```
unused import

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,120 @@
...
+import {EmailAuthMsg} from "@zk-email/email-tx-builder/interfaces/IEmailTypes.sol";
+import {CommandUtils} from "@zk-email/email-tx-builder/libraries/CommandUtils.sol";
+import {AbstractSigner} from "./AbstractSigner.sol";
```
benceharomi, line 11:
```suggestion
```
unused import

benceharomi (commented):
@ernestognw the code looks great! added some comments to the unused imports I found

contracts/utils/cryptography/ZKEmailUtils.sol
```diff
@@ -0,0 +1,120 @@
...
+        if (!dkimregistry.isDKIMPublicKeyHashValid(emailAuthMsg.proof.domainName, emailAuthMsg.proof.publicKeyHash)) {
+            return EmailProofError.DKIMPublicKeyHash;
+        } else if (bytes(emailAuthMsg.proof.maskedCommand).length > verifier.commandBytes()) {
+            return EmailProofError.MaskedCommandLength;
+        } else if (emailAuthMsg.skippedCommandPrefix >= verifier.commandBytes()) {
+            return EmailProofError.SkippedCommandPrefixSize;
+        } else if (!_commandMatch(emailAuthMsg, template, stringCase)) {
+            return EmailProofError.MismatchedCommand;
+        } else {
+            return verifier.verifyEmailProof(emailAuthMsg.proof) ? EmailProofError.NoError : EmailProofError.EmailProof;
```
benceharomi, line 100: @JohnGuilding do we care about the order of checks here? If we don't wouldn't it be cheaper to go in this order?
```
// 1. MaskedCommandLength
// 2. SkippedCommandPrefixSize
// 3. MismatchedCommand
// 4. DKIMPublicKeyHash
```

ernestognw:
> @ernestognw the code looks great! added some comments to the unused imports I found

Thanks for your feedback @benceharomi, I just updated removing the unused imports 😄

contracts/utils/cryptography/ZKEmailUtils.sol
ernestognw: I think we can make this change. Generally, we don't worry about the cost of the failure case, especially because we don't know what branch is most likely to fail. For example, if the error distribution is 80% DKIMPublicKeyHash, we should prioritize that one first.

I'm updating the code since I don't think the order is a big issue.

ernestognw:
All good on our side, gentlemen. The only thing we may be missing is just a real email proof that we can use for testing, currently everything is mocked.

Captura de pantalla 2025-04-10 a la(s) 10 08 39 p m

Also, I noticed it may not be possible not ensure the signer doesn't revert as we want for the ERC-4337 validation phase. The issue is that we're always doing an `abi.decode()` over the signature in `_rawValidateSignature`, so reverting is as easy as just passing a wrong value. I'm adding a [minimum length checl](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96/files#diff-a26349dade685464648402a2be79d6887e725a9dda5848eb8dc1443753a6c156R128-R146) that helps some tests pass but the only alternative might be directly calldata decoding, which is not an easy task and may require heavy use of assembly.

ernestognw:
Merging, thanks everyone for their feedback ❤️

PR #97, opened 2025-03-28
Deprecate Account and rename AccountCore to Account

Amxx:
Fixes https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/95#discussion_r2013319539


ernestognw (approved):
Updated the documentation where I thought it was needed. I'm also including this change and #95 in the CHANGELOG

PR #98, opened 2025-03-29
Add SignerERC7913 and verifiers for P256 and RSA

Amxx:
Includes
- `contracts/interfaces/IERC7913.sol`: ERC-7913 interface
- `contracts/utils/cryptography/ERC7913SignatureVerifierP256.sol` a stateless verifier for P256 keys. This is unopinionated and production ready, so I don't think it needs to be a mock
- `contracts/utils/cryptography/ERC7913SignatureVerifierRSA.sol` a stateless verifier for RSA keys. This is unopinionated and production ready, so I don't think it needs to be a mock
- `contracts/utils/cryptography/ERC7913Utils.sol` library for checking signer's signatures using ERC-7913 (with ECDSA/ERC-1271 fallback)
- `contracts/utils/cryptography/SignerERC7913.sol` abstract signer that uses an ERC-7913 signer.


contracts/utils/cryptography/ERC7913SignatureVerifierP256.sol
```diff
@@ -0,0 +1,26 @@
...
+
+import {P256} from "@openzeppelin/contracts/utils/cryptography/P256.sol";
+import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
+
+/**
+ * @dev ERC-7913 signature verifier that support P256 (secp256r1) keys.
+ */
+contract ERC7913SignatureVerifierP256 is IERC7913SignatureVerifier {
+    /// @inheritdoc IERC7913SignatureVerifier
+    function verify(bytes calldata key, bytes32 hash, bytes calldata signature) external view returns (bytes4) {
```
Amxx, line 13:
```suggestion
    function verify(bytes calldata key, bytes32 hash, bytes calldata signature) public view virtual returns (bytes4) {
```

contracts/utils/cryptography/ERC7913SignatureVerifierP256.sol
```diff
@@ -0,0 +1,26 @@
...
+import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
+
+/**
+ * @dev ERC-7913 signature verifier that support P256 (secp256r1) keys.
+ */
+contract ERC7913SignatureVerifierP256 is IERC7913SignatureVerifier {
+    /// @inheritdoc IERC7913SignatureVerifier
+    function verify(bytes calldata key, bytes32 hash, bytes calldata signature) public view virtual returns (bytes4) {
+        // Signature length may be 0x40 or 0x41.
+        if (key.length == 0x40 && signature.length > 0x3f && signature.length < 0x42) {
```
Amxx, line 15:
```suggestion
        if (key.length == 0x40 && signature.length >= 0x40) {
```

contracts/utils/cryptography/ERC7913Utils.sol
```diff
@@ -0,0 +1,74 @@
...
+
+import {SignatureChecker} from "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
+import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
+import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
+
+/**
+ * @dev Helper library to verify key signatures following the ERC-7913 standard, with fallback to ECDSA and ERC-1271
+ * when the signer's key is empty (as specified in ERC-7913)
+ */
+library ERC7913Utils {
```
ernestognw, line 13: I think I would prefer having just one implementation of the `isValidSignatureNow` function. The rationale is that `signature` always ends up in memory for any valid verification, and a portion of `signer` will be copied too. So I doubt there's actual benefits in having two versions. I'll simplify for now following #109

Amxx: The memory version uses a slice, that does a memory copy. 

Overall, calldata versions are often cheaper because they don't copy to memory until the very last point (when the Abi call is encoded) contrary to the memory version that copy to memory at least once more (in this case 2 times)

Overall not a big deal. Having a single "memory" version is probably good enough. Just explaining the logic

ernestognw: Yeah I saw the memory version is required, so I'd agree with just one "memory" version for now unless there are significant savings measured.

Thanks for answering though, now please you go 👶🏻

ernestognw (dismissed):

ernestognw (approved):

PR #99, opened 2025-03-31
Audit fixes for v5.3

Amxx:


contracts/account/extensions/AccountERC7579.sol
```diff
@@ -259,14 +268,16 @@ abstract contract AccountERC7579 is Account, IERC1271, IERC7579Execution, IERC75
...
-     * rest of the data to be sent to the handler when calling {IERC7579Module-onUninstall}.
+     * For the fallback module type, the `deInitData` is expected to be the (packed) concatenation of a 4-byte
+     * selector and the rest of the data to be sent to the handler when calling {IERC7579Module-onUninstall}.
      *
      * Requirements:
      *
      * * Module must be already installed. Reverts with {ERC7579UninstalledModule} otherwise.
      */
     function _uninstallModule(uint256 moduleTypeId, address module, bytes memory deInitData) internal virtual {
+        require(supportsModule(moduleTypeId), ERC7579Utils.ERC7579UnsupportedModuleType(moduleTypeId));
```
arr00, line 279: Thoughts on reverting in the else clause instead of checking here?

Amxx: I was thinking "what if someone wants to disable a type of module". For example, keep validation module but stop supporting execution module.

- They would update the `supportsModule` function.
- That would prevent installing
- If they try to uninstall, they would get a `ERC7579UninstalledModule` error instead of a `ERC7579UnsupportedModuleType`.

I don't have strong opinion either way. @ernestognw wdyt ?

Amxx: Note that `supportsModule` doesn't perform any storage operation.

arr00: That's reasonable.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -336,9 +347,10 @@ abstract contract AccountERC7579 is Account, IERC1271, IERC7579Execution, IERC75
      * ```
      * <module address (20 bytes)> | <key (4 bytes)> | <nonce (8 bytes)>
      * ```
-     * NOTE: The default behavior of this function replicated the behavior of
+     * NOTE: The default behavior of this function replicates the behavior of
      * https://github.com/rhinestonewtf/safe7579/blob/bb29e8b1a66658790c4169e72608e27d220f79be/src/Safe7579.sol#L266[Safe adapter] and
      * https://github.com/etherspot/etherspot-prime-contracts/blob/cfcdb48c4172cea0d66038324c0bae3288aa8caa/src/modular-etherspot-wallet/wallet/ModularEtherspotWallet.sol#L227[Etherspot's Prime Account].
+     * https://github.com/erc7579/erc7579-implementation/blob/16138d1afd4e9711f6c1425133538837bd7787b5/src/MSAAdvanced.sol#L247[ERC7579 reference implementation]
```
arr00, line 353:
```suggestion
     * https://github.com/rhinestonewtf/safe7579/blob/bb29e8b1a66658790c4169e72608e27d220f79be/src/Safe7579.sol#L266[Safe adapter],
     * https://github.com/etherspot/etherspot-prime-contracts/blob/cfcdb48c4172cea0d66038324c0bae3288aa8caa/src/modular-etherspot-wallet/wallet/ModularEtherspotWallet.sol#L227[Etherspot's Prime Account], and
     * https://github.com/erc7579/erc7579-implementation/blob/16138d1afd4e9711f6c1425133538837bd7787b5/src/MSAAdvanced.sol#L247[ERC7579 reference implementation].
```

contracts/account/extensions/AccountERC7579.sol
ernestognw: I slightly prefer to call `supportsModule` first as it's implemented now.

contracts/utils/cryptography/ERC7739Utils.sol
```diff
@@ -56,12 +56,17 @@ library ERC7739Utils {
...
-     * - `contentsHash` is the hash of the underlying data structure or message
-     * - `contentsDescr` is a descriptor of the "contents" part of the the EIP-712 type of the nested signature
+     * - `signature` is the signature for the (ERC-7739) nested struct hash. This signature indirectly signs over the
+     *   original "contents" hash (from the app) and the account's domain separator.
+     * - `APP_DOMAIN_SEPARATOR` is the EIP-712 {EIP712-_domainSeparatorV4} of the application smart contract that is
+     *   requesting the signature verification (though ERC-1271).
+     * - `contentsHash` is the hash of the underlying data structure or message.
+     * - `contentsDescr` is a descriptor of the "contents" part of the the EIP-712 type of the nested signature.
+     *
+     * Note: If decoding fails due to the input format being invalid, this function does not revert and returns empty
```
ernestognw, line 63:
```suggestion
     * NOTE: This function returns empty if the input format is invalid instead of reverting.
```

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -40,7 +40,16 @@ import {Account} from "../Account.sol";
...
+ * * When combined with {ERC7739}, resolution ordering of {isValidSignature} may have an impact ({ERC7739} does not
+ *   call super). Manual resolution might be necessary.
+ * * Static calls (using callType 0xfe) are currently NOT supported.
  * ====
+ *
+ * [WARNING]
+ * ===
+ * * When uninstalling validation module, users should make sure to keep at least one validation mechanism active.
+ *   Uninstalling all the validation mechanisms would brick the account.
+ * ===
```
ernestognw, line 52: I think this can be simplified to just

```suggestion
 * WARNING: Removing all validator modules will render the account inoperable, as no user operations can be validated thereafter.
```

Note I'm purposely not using the term "brick" since I think it may be confusing.

contracts/account/extensions/AccountERC7579.sol
```diff
@@ -40,7 +40,16 @@ import {Account} from "../Account.sol";
  *   internal virtual functions {_extractUserOpValidator} and {_extractSignatureValidator}. Both are implemented
  *   following common practices. However, this part is not standardized in ERC-7579 (or in any follow-up ERC). Some
  *   accounts may want to override these internal functions.
+ * * When combined with {ERC7739}, resolution ordering of {isValidSignature} may have an impact ({ERC7739} does not
+ *   call super). Manual resolution might be necessary.
+ * * Static calls (using callType 0xfe) are currently NOT supported.
```
ernestognw, line 45:
```suggestion
 * * Static calls (using callType `0xfe`) are currently NOT supported.
```

contracts/account/extensions/ERC7821.sol
```diff
@@ -7,7 +7,9 @@ import {IERC7821} from "../../interfaces/IERC7821.sol";
 import {Account} from "../Account.sol";
 
 /**
- * @dev Minimal batch executor following ERC-7821. Only supports basic mode (no optional "opData").
+ * @dev Minimal batch executor following ERC-7821.
+ *
+ * Only supports supports single batch mode (0x01000000000000000000). Does not support optional "opData".
```
ernestognw, line 12:
```suggestion
 * Only supports supports single batch mode (`0x01000000000000000000`). Does not support optional "opData".
```

contracts/account/extensions/AccountERC7579Hooked.sol
```diff
@@ -21,6 +21,8 @@ import {AccountERC7579} from "./AccountERC7579.sol";
 abstract contract AccountERC7579Hooked is AccountERC7579 {
     address private _hook;
 
+    error ERC7579HookModuleAlreadyPresent(address hook);
```
ernestognw, line 24: We need some natspec for this error

ernestognw (approved):

contracts/utils/cryptography/ERC7739.sol
```diff
@@ -38,8 +38,8 @@ abstract contract ERC7739 is AbstractSigner, EIP712, IERC1271 {
      */
     function isValidSignature(bytes32 hash, bytes calldata signature) public view virtual returns (bytes4 result) {
         // For the hash `0x7739773977397739773977397739773977397739773977397739773977397739` and an empty signature,
-        // we return the magic value too as it's assumed impossible to find a preimage for it that can be used maliciously.
-        // Useful for simulation purposes and to validate whether the contract supports ERC-7739.
+        // we return the magic value 0x77390001 as it's assumed impossible to find a preimage for it that can be used
```
arr00, line 41:
```suggestion
        // we return the magic value `0x77390001` as it's assumed impossible to find a preimage for it that can be used
```

arr00 (approved):

PR #102, opened 2025-04-03
Rename PaymasterNFT to PaymasterERC721Owner

### LINKED ISSUE #101: Rename PaymasterNFT
> @ernestognw
> 
> I think `PaymasterNFT` was a mistake, and we should've called it `PaymasterERC721`, did we discussed it? It's just that now it seems pretty arbitrary when reviewing 😅

> @amxx
> 
> That was the name in the grant agreement.
> 
> I'd be ok with changing it, but I would go even further and call it something like PaymasterERC721Owner to clarify that the ownership gives you the right and you are not actually giving away the NFT as a payment.

Amxx:
Fixes #101

### DIFF
--- a/contracts/account/paymaster/PaymasterERC721Owner.sol
+++ b/contracts/account/paymaster/PaymasterERC721Owner.sol
@@ -12,11 +12,11 @@ import {PaymasterCore} from "./PaymasterCore.sol";
  * This paymaster will sponsor user operations if the user has at least 1 token of the token specified
  * during construction (or via {_setToken}).
  */
-abstract contract PaymasterNFT is PaymasterCore {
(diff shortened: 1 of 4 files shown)


contracts/account/paymaster/PaymasterERC721Owner.sol
```diff
@@ -27,10 +27,15 @@ abstract contract PaymasterNFT is PaymasterCore {
         return _token;
     }
 
+    /// @dev Minimum balance necessary. Default: 1 (having a single token is sufficient).
+    function minimumBalance() public virtual returns (uint256) {
```
Amxx, line 31:
```suggestion
    function minimumBalance() public view virtual returns (uint256) {
```

ernestognw (approved):

PR #103, opened 2025-04-03
ERC-7913 signature verifier for ZKEmail

Amxx:
Based on #96 and #98

The relevant file here is `contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol`

### DIFF
--- a/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
+++ b/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
@@ -2,8 +2,6 @@
 
 pragma solidity ^0.8.20;
 
-import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
-
 import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
 import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
 import {EmailAuthMsg} from "@zk-email/email-tx-builder/interfaces/IEmailTypes.sol";
--- a/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
+++ b/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
@@ -11,7 +11,7 @@ import {ZKEmailUtils} from "./ZKEmailUtils.sol";
 /**
  * @dev ERC-7913 signature verifier that supports ZKEmail accounts.
  *
- * This verifier validates signatures produced through ZKEmail's zero-knowledge
+ * This contract verifies signatures produced through ZKEmail's zero-knowledge
  * proofs which allows users to authenticate using their email addresses.
  */
 abstract contract ERC7913SignatureVerifierZKEmail is IERC7913SignatureVerifier {
@@ -25,8 +25,10 @@ abstract contract ERC7913SignatureVerifierZKEmail is IERC7913SignatureVerifier {
         _templateId = templateId_;
     }
 
-    /// @dev An instance of the Verifier contract.
-    /// See https://docs.zk.email/architecture/zk-proofs#how-zk-email-uses-zero-knowledge-proofs[ZK Proofs].
+    /**
+     * @dev An instance of the Verifier contract.
+     * See https://docs.zk.email/architecture/zk-proofs#how-zk-email-uses-zero-knowledge-proofs[ZK Proofs].
+     */
     function verifier() public view virtual returns (IVerifier) {
         return _verifier;
     }
--- a/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
+++ b/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
@@ -1,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
+import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
+import {EmailAuthMsg} from "@zk-email/email-tx-builder/interfaces/IEmailTypes.sol";
+import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
+import {ZKEmailUtils} from "./ZKEmailUtils.sol";
+
+/**
+ * @dev ERC-7913 signature verifier that support ZKEmail accounts.
+ */
+contract ERC7913SignatureVerifierZKEmail is IERC7913SignatureVerifier {
+    using ZKEmailUtils for EmailAuthMsg;
+
+    IVerifier public immutable verifier;
+    uint256 public immutable commandTemplate;
+
+    constructor(IVerifier _verifier, uint256 _commandTemplate) {
+        verifier = _verifier;
+        commandTemplate = _commandTemplate;
+    }
+
+    /// @inheritdoc IERC7913SignatureVerifier
+    function verify(bytes calldata key, bytes32 hash, bytes calldata signature) public view virtual returns (bytes4) {
+        (IDKIMRegistry registry, bytes32 accountSalt) = abi.decode(key, (IDKIMRegistry, bytes32));
+        EmailAuthMsg memory emailAuthMsg = abi.decode(signature, (EmailAuthMsg));
+        if (
+            bytes32(emailAuthMsg.commandParams[0]) == hash &&
+            emailAuthMsg.templateId == commandTemplate &&
+            emailAuthMsg.proof.accountSalt == accountSalt
+        ) {
+            (bool verified, ) = emailAuthMsg.isValidZKEmail(registry, verifier);
+            if (verified) {
+                return IERC7913SignatureVerifier.verify.selector;
+            }
+        }
+        return 0xffffffff;
+    }
+}
--- a/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
+++ b/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-pragma solidity ^0.8.20;
+pragma solidity ^0.8.24;
 
 import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
 import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
--- a/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
+++ b/contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
@@ -9,29 +9,77 @@ import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
 import {ZKEmailUtils} from "./ZKEmailUtils.sol";
 
 /**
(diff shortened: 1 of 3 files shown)


contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
```diff
@@ -0,0 +1,86 @@
...
+import {IDKIMRegistry} from "@zk-email/contracts/DKIMRegistry.sol";
+import {IVerifier} from "@zk-email/email-tx-builder/interfaces/IVerifier.sol";
+import {EmailAuthMsg} from "@zk-email/email-tx-builder/interfaces/IEmailTypes.sol";
+import {IERC7913SignatureVerifier} from "../../interfaces/IERC7913.sol";
+import {ZKEmailUtils} from "./ZKEmailUtils.sol";
+
+/**
+ * @dev ERC-7913 signature verifier that supports ZKEmail accounts.
+ *
+ * This verifier validates signatures produced through ZKEmail's zero-knowledge
```
arr00, line 14:
```suggestion
 * This verifier verifies signatures produced through ZKEmail's zero-knowledge
```
Verifies?

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
```diff
@@ -0,0 +1,86 @@
...
+    IVerifier private immutable _verifier;
+    uint256 private immutable _templateId;
+
+    constructor(IVerifier verifier_, uint256 templateId_) {
+        _verifier = verifier_;
+        _templateId = templateId_;
+    }
+
+    /// @dev An instance of the Verifier contract.
+    /// See https://docs.zk.email/architecture/zk-proofs#how-zk-email-uses-zero-knowledge-proofs[ZK Proofs].
```
arr00, line 29:
```suggestion
    /**
     * @dev An instance of the Verifier contract.
     * See https://docs.zk.email/architecture/zk-proofs#how-zk-email-uses-zero-knowledge-proofs[ZK Proofs].
     */
```

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
```diff
@@ -0,0 +1,86 @@
...
+    /// @dev The command template of the sign hash command.
+    function templateId() public view virtual returns (uint256) {
+        return _templateId;
+    }
+
+    /**
+     * @dev Verifies a zero-knowledge proof of an email signature validated by a {DKIMRegistry} contract.
+     *
+     * The key format is ABI-encoded (IDKIMRegistry, bytes32) where:
+     * - IDKIMRegistry: The registry contract that validates DKIM public key hashes
```
arr00, line 43: Is this a canonical address or one particular to the DKIM public key?

ernestognw: Not really, they even suggest to deploy it yourself: https://docs.zk.email/email-tx-builder/quickstart#build-and-deploy

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
ernestognw: Sounds redundant, let's do "this contract verifies"

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
JohnGuilding: > Is this a canonical address or one particular to the DKIM public key?

actually, we do envision this as a canonical address per chain, as this is the dkim registry that get's updated by our infrastructure. Users can override the dkim keys on the canonical registry if they want

Ofc users can still deploy their own but they would have to update it themselves.

This is not clear from our docs so we will update them!

ernestognw: > actually, we do envision this as a canonical address per chain, as this is the dkim registry that get's updated by our infrastructure. 

This would be amazing! If that's the case we could update the `SignerZKEmail` and this `ERC7913SignatureVerifierZKEmail` to hardcode the address of the canonical registry in the `DKIMRegistry` function, similar to how we [hardcode the Entrypoint in Accounts](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/contracts/account/Account.sol#L49). Ideally I would like developers to not have to worry about locating/deploying the registry

JohnGuilding: yeah ok we could look at doing that. We can run a new deployment and provide an address that would be consistent across chains

test/account/AccountERC7913.test.js
```diff
@@ -15,15 +15,39 @@ const signerECDSA = ethers.Wallet.createRandom();
...
+
 // Minimal fixture common to the different signer verifiers
 async function fixture() {
   // EOAs and environment
-  const [beneficiary, other] = await ethers.getSigners();
+  const [admin, beneficiary, other] = await ethers.getSigners();
   const target = await ethers.deployContract('CallReceiverMockExtended');
 
+  // DKIM Registry for ZKEmail
+  const dkim = await ethers.deployContract('ECDSAOwnedDKIMRegistry');
```
JohnGuilding, line 33: I missed this change from the last PR regarding the use of `ECDSAOwnedDKIMRegistry`.

We should actually use `UserOverrideableDKIMRegistry` which comes from `@zk-email/contracts`.

We soft deprecated `ECDSAOwnedDKIMRegistry` in favour of `UserOverrideableDKIMRegistry` as the latter is overridable and thus provides an escape hatch to users. We should make a stronger note of this in our repo/docs

Since `UserOverrideableDKIMRegistry` is the contract we would use, would it be possible to update all references of `ECDSAOwnedDKIMRegistry` to the that instead?

Example [import](https://github.com/zkemail/email-tx-builder/blob/b49ed9fbc0e87ecaf6dcdb05c4dc0118727d3593/packages/contracts/script/BaseDeployScript.sol#L8)

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
```diff
@@ -0,0 +1,88 @@
...
+ * @dev ERC-7913 signature verifier that supports ZKEmail accounts.
+ *
+ * This contract verifies signatures produced through ZKEmail's zero-knowledge
+ * proofs which allows users to authenticate using their email addresses.
+ */
+abstract contract ERC7913SignatureVerifierZKEmail is IERC7913SignatureVerifier {
+    using ZKEmailUtils for EmailAuthMsg;
+
+    IVerifier private immutable _verifier;
+    uint256 private immutable _templateId;
```
JohnGuilding, line 21: What is the thinking here behind adding verifier and templateId, but not dkimRegistry and accountSalt? I noticed other 7913 verifiers do not have any state variables

ernestognw: Ahh good point. The standard defines stateless verifiers, its only purpose is to implement a universal verification mechanism for the algorithm they implement. In the case of the `ERC7913SignatureVerifierZKEmail`, it should be able to verify for any `dkimRegistry` and `accountSalt`, but sanitizing those is responsibility of the developer using the verifier (e.g. whitelisting `key || verifier` pairs)

test/account/AccountERC7913.test.js
ernestognw: I tried using the `UserOverrideableDKIMRegistry` but it has an initializer and the constructor is already set. I think using ECDSAOwnedDKIMRegistry in the tests is fine. Note that the actual ZKEmailUtils and SignerZKEmail only require the `IDKIMRegistry` interface, so both should be compatible.

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
ernestognw: I've been thinking about this. These variables are immutable in compliance with ERC-7913 statelessness since they're fixed for multiple emails. However, one could argue the `dkimRegistry` could fall into the same category, and similarly, an argument can be made that both must be part of the encoding.

I've come to the conclusion that the most elegant solution is to have all elements in the key, so that developers can override each value if they want to:

```
/**
     * @dev Decodes the key into its components.
     *
     * ```solidity
     * bytes memory key = abi.encode(registry, accountSalt, verifier, templateId);
     * ```
     */
    function _decodeKey(
        bytes calldata key
    )
        internal
        view
        virtual
        returns (IDKIMRegistry registry, bytes32 accountSalt, IVerifier verifier, uint256 templateId)
    {
        return abi.decode(key, (IDKIMRegistry, bytes32, IVerifier, uint256));
    }
```

For example:

```
function _decodeKey(
        bytes calldata key
    )
        internal
        view
        override
        returns (IDKIMRegistry registry, bytes32 accountSalt, IVerifier verifier, uint256 templateId)
    {
       (registry, accountSalt, verifier, templateId) = abi.decode(key, (IDKIMRegistry, bytes32, IVerifier, uint256));
        // Validate any of these against immutable variables (e.g. require(verifier == cachedImmutableVerifier))
        return super._decodeKey(key);
    }
```

I'll push an update

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
ernestognw: Opened an issue to track this: https://github.com/OpenZeppelin/openzeppelin-community-contracts/issues/140

ernestognw:
Looks good, thanks all for your feedback. I'm merging so we can move forward, I want to rename the ERC7913SignatureVerifiers to just ERC7913<Something>Verifier, which I think is clearer.

I'll also work on a proposal to add multiple templateIds in the SignerZKEmail

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
JohnGuilding: Sweet. We'll aim to do the multi-chain deploy by end of next week

test/account/AccountERC7913.test.js
JohnGuilding: > but it has an initializer and the constructor is already set

I don't understand this point, can you clarify?

> It's something we can also emphasize more in our contracts. For example, giving concrete instructions to use the UserOverrideableDKIMRegistry wherever the IDKIMRegistry is declared, wdyt?

The main point I would make is that we should not use `ECDSAOwnedDKIMRegistry` in the production contracts and examples. For tests I am not as opinionated about so happy to go with your call, as long as there are explicit docs that state `UserOverrideableDKIMRegistry` is the registry used in production

contracts/utils/cryptography/ERC7913SignatureVerifierZKEmail.sol
JohnGuilding: > I've come to the conclusion that the most elegant solution is to have all elements in the key, so that developers can override each value if they want to

Sweet sounds good to me

test/account/AccountERC7913.test.js
ernestognw: > I don't understand this point, can you clarify?

Yes, so the `UserOverrideableDKIMRegistry` requires initialization and uses `onlyInitializing` through `__Ownable_init`, which complicates the tests setup. I think I should either be able to write true to the `_initializing` bool in `Initializable` (the other alternative is a top-level call). Creating a proxy has the same issue, we could work around it if the `UserOverrideableDKIMRegistry` was abstract so we can initialize during construction.

> The main point I would make is that we should not use ECDSAOwnedDKIMRegistry in the production contracts and examples. 

Yeah, that makes sense. I'm happy to note it in the documentation in a follow up PR :smile

PR #109, opened 2025-04-12
Add ERC-7913 interface and utils

ernestognw:
Required for #93 and #98

### DIFF
--- a/contracts/utils/cryptography/ERC7913Utils.sol
+++ b/contracts/utils/cryptography/ERC7913Utils.sol
@@ -1,0 +1,49 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.20;
+
+import {SignatureChecker} from "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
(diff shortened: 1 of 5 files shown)


contracts/utils/cryptography/ERC7913Utils.sol
```diff
@@ -0,0 +1,49 @@
...
+     * The signer is a `bytes` object that is the concatenation of an address and optionally a key:
+     * `verifier || key`. A signer must be at least 20 bytes long.
+     *
+     * Verification is done as follows:
+     * - If `signer.length < 20`: verification fails
+     * - If `signer.length == 20`: verification is done using {SignatureChecker}
+     * - Otherwise: verification is done using {IERC7913SignatureVerifier}
+     */
+    function isValidSignatureNow(
+        bytes calldata signer,
```
ernestognw, line 31: It's cool that `signer` is in calldata but I suspect most of the times it will end up in memory as arguments are used to construct the external call. Perhaps it makes sense to turn this into `memory` eventually. Not a blocker

PR #111, opened 2025-04-12
Add actual ERC7702 tests for SignerERC7702

ernestognw:
### Description

So far, we're missing real ERC-7702 since this feature is not enabled in hardhat yet. However, `pectra` and `signAndAttachDelegation` is available in Foundry so it makes sense to include a tests where it does work with ERC-7702


test/account/AccountERC7702.t.sol
```diff
@@ -0,0 +1,100 @@
...
...d90fd5b608490604051907f220266b6000000000000000000000000000000000000000000000000000000008252600482015260406024820152601760448201527f4141393920696e6974436f646520746f6f20736d616c6c0000000000000000006064820152fd5b608490604051907f220266b6000000000000000000000000000000000000000000000000000000008252600482015260406024820152601f60448201527f414131302073656e64657220616c726561647920636f6e7374727563746564006064820152fd5b945050919050601482116153735750505050565b604073ffffffffffffffffffffffffffffffffffffffff7f000000000000000000000000449ed7c3e6fee6a97311d4b55475df59c44add33169301519082601411612a3257833b15612a32575f809461542f96604051978896879586937fc09ad0d900000000000000000000000000000000000000000000000000000000855260048501526040602485015260147fffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffec60448601930191016131b6565b0393f1801561545557615445575b8080806145b7565b5f61544f91612c2d565b5f61543d565b6040513d5f823e3d90fd5b615478604092959493956060835260608301906133bd565b9460208201520152565b73ffffffffffffffffffffffffffffffffffffffff165f525f60205260405f2090815481811061381d5703905560019056fea2646970667358221220a2ee7c02d47f72772240d0dfa7174d99b6049a68ccdf3d4434c3918f6bd9c1e164736f6c634300081c0033"
```
ernestognw, line 48: Either we hardcode the entrypoint code here, we [use `ffi`](https://book.getfoundry.sh/cheatcodes/ffi), or import the Entrypoint from the [account-abstraction repo](https://github.com/eth-infinitism/account-abstraction/blob/develop/contracts/core/EntryPoint.sol) to deploy it.

I don't like neither of those

ernestognw: Fixed with `vm.readFileBinary`

ernestognw:
Merging since these are equivalent to #114

PR #114, opened 2025-04-15
"Real" EIP-7702 account testing in hardhat.

### LINKED ISSUE #106: Testing EIP-7702
Currently, EIP-7702 is "simulated" by
- [deploying an implementation contract](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/test/helpers/erc4337.js#L33)
- [copying the deployed bytecode for this contract to the EOA's address](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/master/test/helpers/erc4337.js#L92)

While this is good enough for simple testing, the second point should eventually be replaced by a proper degelation setup.

---

- [EIP-7702 is available in edr 0.9.0](https://github.com/NomicFoundation/edr/releases/tag/%40nomicfoundation%2Fedr%400.9.0)
- [This was merged in hardhat in PR #6420](https://github.com/NomicFoundation/hardhat/pull/6420) on march 3
- EIP-7702 is supported in [this branch of ethers](https://github.com/ethers-io/ethers.js/tree/eip-7702) [which is released on npm under the `beta-eip-7702` tag](https://www.npmjs.com/package/ethers/v/6.13.6-beta.1)

Amxx:
Fixes #106

### DIFF
--- a/test/helpers/erc4337.js
+++ b/test/helpers/erc4337.js
@@ -1,5 +1,4 @@
-const { ethers, entrypoint, senderCreator } = require('hardhat');
-const { setCode } = require('@nomicfoundation/hardhat-network-helpers');
+const { ethers, config, entrypoint, senderCreator } = require('hardhat');
 
 const { UserOperation } = require('@openzeppelin/contracts/test/helpers/erc4337');
 
@@ -32,7 +31,8 @@ class ERC4337Helper {
     if (params.erc7702signer) {
       const delegate = await accountFactory.deploy(...extraArgs);
       const instance = await params.erc7702signer.getAddress().then(address => accountFactory.attach(address));
-      return new ERC7702SmartAccount(instance, delegate, env);
+      const authorization = await params.erc7702signer.authorize({ address: delegate.target });
+      return new ERC7702SmartAccount(instance, authorization, env);
     } else {
       const initCode = await accountFactory
         .getDeployTransaction(...extraArgs)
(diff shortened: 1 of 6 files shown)


test/helpers/erc4337.js
```diff
@@ -83,27 +83,41 @@ class SmartAccount extends ethers.BaseContract {
...
     super(instance, undefined, env);
-    this.delegate = delegate;
+    this.authorization = authorization;
   }
 
   async deploy() {
-    await ethers.provider.getCode(this.delegate).then(code => setCode(this.target, code));
+    // hardhat signers from @nomicfoundation/hardhat-ethers do not support type 4 txs.
+    // so we rebuild it using "native" ethers
+    await ethers.Wallet.fromPhrase(config.networks.hardhat.accounts.mnemonic, ethers.provider).sendTransaction({
```
ernestognw, line 94: Isn't this tied to the first account? I tried to make `AccountERC7702WithModules` to revert unsuccessfully, but I think we should do something like:

```js
await ethers.Wallet.fromPhrase(this.findAccount(), ethers.provider).sendTransaction(...)

findAccount() {
  for (const account of config.networks.hardhat.accounts) {
    console.log(account);
    if (account.address === this.address) {
      return account;
    }
  }
}
```

Amxx: config.network.hardhat.accounts is not an array. It's an object that contains a mnemonic from which all signers are derived. It does not contain an address.

And yes, this code uses the first one of the bunch.

Amxx: Also, if we make the EOA relay its own delegation, we get into nonce issues that I'd rather avoid.

Amxx: Basically:
- any account can relay
- account needs gas
- account objects needs to support type4 txs (hardhat-ether) signer do not
- if the account used is the signer of the delegate, we need to fix the nounces.

So my choice was to take the first "canonical" account from the hardhat config and instantiate it as a signer that supports type 4 txs

ernestognw: Right! So the "first canonical account" just relays. I forgot we reserved the first account, so that explains why I couldn't make the test fail.

Makes sense then

Amxx: First account is not reserved.

I'm not sure what you think could cause any of that to fail.

ernestognw: > If I remember correctly, the first account is no longer reserved.

Also right, I thought it was still around and now I just realized that an EOA can send a transaction with another's EOA authorization 🤯 didn't realize that but it's extremely convenient

Amxx: That is why you need the authorizations to be signed. Authorizations are like a meta transaction.

In particular, and ERC-4337 relayer can include the authorization with the user operation so that the 4337+7702 logic is deployed and used in a single operation

PR #118, opened 2025-04-23
Add setup to generate real ZKProofs

ernestognw:
### Description

Current ZKEmail tests don't use an actual ZK Proof, so it would be convenient to setup a way to generate real proofs and store them as fixtures that are later used for testign


ernestognw:
Unsuccessful experiment

PR #119, opened 2025-04-23
Fix Account usage examples

ernestognw:
### Description

Most of the accounts still add EIP712 since that was the default before #83 (I think). This PR updates the examples accordingly


ernestognw:
Merging as it's only docs

Issue #30, opened 2024-11-29
Documentation does not link functions from `@openzeppelin/contracts`


ernestognw:
### Description

The documentation API reference is automatically generated based on the vanilla contracts script. This causes the documentation to generate references that aren't available in this repository.

![image](https://github.com/user-attachments/assets/d0e27180-9d37-4dcd-854b-31138cb83048)

To avoid this issue naively, I implemented a [quick fix](https://github.com/OpenZeppelin/openzeppelin-community-contracts/commit/5407d105f1d188ffbf917438428c278edaa01609) to render the names without the reference.

Also, it doesn't solve references written in the documentation text itself:

![image](https://github.com/user-attachments/assets/73259eb8-fa1e-4b58-ac76-ce8f3a03a837)

Perhaps we can easily fix this second issue by linking to the OpenZeppelin documentation URL as we don't expect it to change until 6.0

PR #121, opened 2025-04-25
Add ERC7579 Executor modules

ernestognw:
More generalized alternatives to #93 and #85


contracts/account/modules/ERC7579Executor.sol
```diff
@@ -0,0 +1,56 @@
...
+     * @dev Executes an operation and returns the result data from the executed operation.
+     * See {_execute} for requirements.
+     */
+    function execute(
+        Mode mode,
+        bytes calldata executionCalldata,
+        bytes32 salt
+    ) public virtual returns (bytes[] memory returnData) {
+        return _execute(msg.sender, mode, executionCalldata, salt);
+    }
```
arr00, line 38: What is the point in this?

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,394 @@
...
+ * IMPORTANT: This module assumes the {AccountERC7579} is the ultimate authority and does not restrict
+ * module uninstallation. An account can bypass the time-delay security by simply uninstalling
+ * the module. Consider adding safeguards in your Account implementation if uninstallation
+ * protection is required for your security model.
+ */
+abstract contract ERC7579DelayedExecutor is ERC7579Executor {
+    using Time for *;
+
+    uint32 private constant NO_DELAY = type(uint32).max; // Sentinel value for no delay
+    uint32 private constant EXECUTED = type(uint32).max - 1; // Sentinel value for no delay
```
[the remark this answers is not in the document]
ernestognw: ?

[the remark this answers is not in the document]
arr00: Comment for the second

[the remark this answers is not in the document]
ernestognw: Ah got it, I'll push some updates in general

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -57,32 +65,126 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
...
+        if (sched.canceled) return OperationState.Canceled;
+        if (sched.executed) return OperationState.Executed;
+        if (block.timestamp < sched.scheduledAt + sched.executableAfter) return OperationState.Scheduled;
+        if (block.timestamp > sched.scheduledAt + sched.expiresAfter) return OperationState.Expired;
+        return OperationState.Ready;
+    }
 
-    /// @dev Thrown when trying to execute an operation that has expired.
-    error ERC7579ExecutorOperationExpired(bytes32 operationId, uint48 expiresAt);
+    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's {OperationState-Ready}.
```
gonzaotc, line 122: Note that this allows the `Account` to execute any operation even if the `OperationState` is not ready, since `ERC7579Executor.canExecute` returns `msg.sender == account;`

If this is intended behavior, I would add an explicit note about into the NatSpec:
```suggestion
    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's {OperationState-Ready.
    /// Note: The Account is allowed to bypass this restriction and execute during any OperationState.
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,464 @@
...
+    function canExecute(
+        address account,
+        Mode mode,
+        bytes calldata executionCalldata,
+        bytes32 salt
+    ) public view virtual override returns (bool) {
+        bytes32 id = hashOperation(account, mode, executionCalldata, salt);
+        return state(id) == OperationState.Ready || super.canExecute(account, mode, executionCalldata, salt);
+    }
```
gonzaotc, line 131: Note that this allows the `Account` to execute any operation even if the `OperationState` is not ready, since `ERC7579Executor.canExecute` returns `msg.sender == account;`

If this is intended behavior, I would add an explicit note about into the NatSpec:
```suggestion
    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's {OperationState-Ready}.
    /// Note: The Account is allowed to bypass this restriction and execute during any OperationState.
    function canExecute(
        address account,
        Mode mode,
        bytes calldata executionCalldata,
        bytes32 salt
    ) public view virtual override returns (bool) {
        bytes32 id = hashOperation(account, mode, executionCalldata, salt);
        return state(id) == OperationState.Ready || super.canExecute(account, mode, executionCalldata, salt);
    }
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,464 @@
...
+    ) public view virtual returns (uint32 delay, uint32 pendingDelay, uint48 effectTime) {
+        (uint32 currentDelay, uint32 newDelay, uint48 effect) = _config[account].delay.getFull();
+        bool installed = IERC7579ModuleConfig(msg.sender).isModuleInstalled(MODULE_TYPE_EXECUTOR, address(this), "");
+        return (
+            // Safe downcast since both arguments are uint32
+            uint32(Math.ternary(installed, 0, Math.max(currentDelay, minimumDelay()))),
+            newDelay,
+            effect
+        );
+    }
```
gonzaotc, line 211: I don't fully understand the intended behavior here

> bool installed = IERC7579ModuleConfig(msg.sender).isModuleInstalled(MODULE_TYPE_EXECUTOR, address(this), "");

We take the `address account` as a parameter but check `(msg.sender).isModuleInstalled` instead. Shouldn't we check `isModuleInstalled` on the provided `address account` ?  

> uint32(Math.ternary(installed, 0, Math.max(currentDelay, minimumDelay()))),

If the module is installed, we are returning 0, `Math.max(currentDelay, minimumDelay())` otherwise. I'm confused, shouldn't it be the other way around?

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,464 @@
...
+     * Behaves as a no-op if the module is already installed.
+     *
+     * Requirements:
+     *
+     * * The account (i.e `msg.sender`) must implement the {IERC7579ModuleConfig} interface.
+     * * The {IERC7579ModuleConfig-isModuleInstalled} function must return not revert.
+     * * `initData` must be empty or decode correctly to `(uint32, uint32)`.
+     */
+    function onInstall(bytes calldata initData) public virtual {
+        bool installed = IERC7579ModuleConfig(msg.sender).isModuleInstalled(MODULE_TYPE_EXECUTOR, address(this), "");
```
gonzaotc, line 269: As I reported on a previous PR: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/127#discussion_r2070759015 this will not work unless we re-arrange the order of `ERC7579Account._installModule`, since currently it first registers the module as installed in it's storage, and then calls this function. 

Wold also appreciate your feedback @ernestognw on the linked comment, since you seem interested on applying the approach I was suggesting here.

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: Right, good point, don't you feel it's enough with the description in the original `execute` in ERC7579Executor. This function is linking to the parent definition

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: Good catch, it was a copy-paster overshight and we should use `account` and invert the `installed` flag as you pointed out.

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,464 @@
...
+    }
+
+    /**
+     * @dev Cleans up the {getDelay} and {getExpiration} values by scheduling them to `0`
+     * and respecting the previous delay and expiration values. Do not consider {minimumDelay} and
+     * {minimumExpiration} for scheduling.
+     *
+     * IMPORTANT: This function does not clean up scheduled operations. This means operations
+     * could potentially be re-executed if the module is reinstalled later. This is a deliberate
+     * design choice, but module implementations may want to override this behavior to clear
```
benceharomi, line 321: Wouldn't it be more secure to clear scheduled operations by default?

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: Yes also good catch. This should come up better during testing, for now the contracts are subject to substantial changes IMO. I'm still not entirely convinced that we want to follow this approach. 

For now I'd lean towards Aryeh's comment on that we should assume accounts won't call `onInstall` or `onUninstall` direct. Can't comment further because I don't have a strong opinion yet

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: Yes, though it may be really expensive without a clear benefit imo. Clearing them would require to change the `_schedules` mapping for an EnumerableSet and then iterate each `operationId` without an easy way to filter for pending operations.

I think this note is the best balance between security and convenience. I wouldn't expect accounts install and uninstall the same module frequently

contracts/account/modules/ERC7579DelayedExecutor.sol
gonzaotc: Nice, thank you for sharing, in my current opinion, checking for installation in the Account is a reliable, standard way to do it, and the increased gas costs of the external call are only faced during installation and uninstallation of the module, which is absolutely not frequent.

Additionally, I think not every Module implementation will have a (mapping Account => State) reliable to determine if the module is installed. For example in this `DelayerExecutor`, `mapping(address account => OperationConfig) private _config` is all we have to do that, and feels "tricky" to me

gonzaotc: In the meanwhile we define the appropriate way to determine installation, I will push a small utility function so we don't repeat `IERC7579ModuleConfig(msg.sender).isModuleInstalled(MODULE_TYPE_EXECUTOR, address(this), "");` all over again

gonzaotc: In the meanwhile we define the appropriate way to determine installation, I will push a small utility function so we don't repeat `IERC7579ModuleConfig(msg.sender).isModuleInstalled(MODULE_TYPE_EXECUTOR, address(this), "");` all over again

contracts/account/modules/ERC7579MultisigWeighted.sol
```diff
@@ -0,0 +1,210 @@
...
+
+        bytes[] memory allSigners = signers(account);
+        for (uint256 i = 0; i < allSigners.length; i++) {
+            delete _weights[account][allSigners[i]];
+        }
+        delete _totalWeight[account];
+
+        // Call parent implementation which will clear signers and threshold
+        super.onUninstall(data);
+    }
```
gonzaotc, line 79: cache length
```suggestion
    function onUninstall(bytes calldata data) public virtual override {
        address account = msg.sender;

        bytes[] memory allSigners = signers(account).length;
        uint256 allSignersLength = allSigners.length;
        for (uint256 i = 0; i < allSignersLength; i++) {
            delete _weights[account][allSigners[i]];
        }
        delete _totalWeight[account];

        // Call parent implementation which will clear signers and threshold
        super.onUninstall(data);
    }
```

contracts/account/modules/ERC7579MultisigWeighted.sol
```diff
@@ -0,0 +1,210 @@
...
+            bytes memory signer = signers[i];
+            uint256 newWeight = newWeights[i];
+            require(isSigner(account, signer), ERC7579MultisigNonexistentSigner(signer));
+            require(newWeight > 0, ERC7579MultisigInvalidWeight(signer, newWeight));
+        }
+
+        _unsafeSetSignerWeights(account, signers, newWeights);
+        _totalWeight[account] = totalWeight(account) - oldWeight + _weightSigners(account, signers);
+        _validateReachableThreshold(account);
+    }
```
gonzaotc, line 133: cache length
```suggestion
    function _setSignerWeights(address account, bytes[] memory signers, uint256[] memory newWeights) internal virtual {
        uint256 signersLength = signers.length;
        require(signersLength == newWeights.length, ERC7579MultisigMismatchedLength());
        uint256 oldWeight = _weightSigners(account, signers);

        for (uint256 i = 0; i < signers.length; i++) {
            bytes memory signer = signers[i];
            uint256 newWeight = newWeights[i];
            require(isSigner(account, signer), ERC7579MultisigNonexistentSigner(signer));
            require(newWeight > 0, ERC7579MultisigInvalidWeight(signer, newWeight));
        }

        _unsafeSetSignerWeights(account, signers, newWeights);
        _totalWeight[account] = totalWeight(account) - oldWeight + _weightSigners(account, signers);
        _validateReachableThreshold(account);
    }
```

contracts/account/modules/ERC7579MultisigWeighted.sol
```diff
@@ -0,0 +1,210 @@
...
+     *
+     * Emits {ERC7913SignerWeightChanged} for each signer.
+     */
+    function _unsafeSetSignerWeights(address account, bytes[] memory signers, uint256[] memory newWeights) private {
+        for (uint256 i = 0; i < signers.length; i++) {
+            delete _weights[account][signers[i]];
+            emit ERC7913SignerWeightChanged(account, signers[i], newWeights[i]);
+        }
+    }
+}
```
gonzaotc, line 210:
```suggestion
    function _unsafeSetSignerWeights(address account, bytes[] memory signers, uint256[] memory newWeights) private {
        uint256 signersLength = signers.lenght;
        for (uint256 i = 0; i < signersLength; i++) {
            delete _weights[account][signers[i]];
            emit ERC7913SignerWeightChanged(account, signers[i], newWeights[i]);
        }
    }
}
```

contracts/account/modules/ERC7579MultisigWeighted.sol
```diff
@@ -0,0 +1,210 @@
...
+     * @dev Cleans up module's configuration when uninstalled from an account.
+     * Clears all signers, weights, and total weights.
+     *
+     * See {ERC7579Multisig-onUninstall}.
+     */
+    function onUninstall(bytes calldata data) public virtual override {
+        address account = msg.sender;
+
+        bytes[] memory allSigners = signers(account);
+        for (uint256 i = 0; i < allSigners.length; i++) {
```
gonzaotc, line 72:
```suggestion
        bytes[] memory allSigners = signers(account);
        uint256 allSignersLength = allSigners.length;
        for (uint256 i = 0; i < allSignersLength; i++) {
```

contracts/account/modules/ERC7579MultisigWeighted.sol
```diff
@@ -0,0 +1,210 @@
...
+     * * Each weight must be greater than 0. Reverts with {ERC7579MultisigInvalidWeight} if not.
+     * * See {_validateReachableThreshold} for the threshold validation.
+     *
+     * Emits {ERC7913SignerWeightChanged} for each signer.
+     */
+    function _setSignerWeights(address account, bytes[] memory signers, uint256[] memory newWeights) internal virtual {
+        require(signers.length == newWeights.length, ERC7579MultisigMismatchedLength());
+        uint256 oldWeight = _weightSigners(account, signers);
+
+        for (uint256 i = 0; i < signers.length; i++) {
```
gonzaotc, line 123: cache length
```suggestion
        uint256 signersLength = signers.length;
        require(signersLength == newWeights.length, ERC7579MultisigMismatchedLength());
        uint256 oldWeight = _weightSigners(account, signers);

        for (uint256 i = 0; i < signers.length; i++) {
```

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: We discussed this internally and concluded that the most common pattern is that modules behave as initializable contracts. For the AccountERC7579, changing the order will violate checks-effects-interactions.

Initializable:
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/policies/ratelimit/src/RateLimitPolicy.sol#L33
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/policies/gas/src/GasPolicy.sol#L32
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/policies/call-policy/src/CallPolicy.sol#L199
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/signers/webauthn/src/WebAuthnSigner.sol#L122
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/signers/ecdsa/src/ECDSASigner.sol#L52
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/policies/timestamp/src/TimestampPolicy.sol#L48
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/policies/signature-caller/src/SignaturePolicy.sol#L44
- https://github.com/kot2271/Automation_Module_ERC_7579/blob/main/src/ExecutorModule.sol#L50
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/validators/ERC20SessionKeyValidator.sol#L226
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/validators/MultipleOwnerECDSAValidator.sol#L28
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/hooks/HookMultiPlexer.sol#L107

Stateless:

- https://github.com/kroist/flashcall-module-erc-7579/blob/main/src/executors/FlashcallFallbackExecutor.sol#L33
- https://github.com/Destiner/module-frame/blob/main/src/FrameValidator.sol#L47
- https://github.com/Destiner/module-airdrop-claim/blob/main/src/module/src/AirdropModule.sol#L38

Open reinstallation:

- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/validators/CredibleAccountModule.sol#L241-L267
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/validators/SessionKeyValidator.sol#L361
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/hooks/ModuleIsolationHook.sol#L23
- https://github.com/etherspot/etherspot-modular-accounts/blob/master/src/modules/fallbacks/ERC1155FallbackHandler.sol#L61
- https://github.com/zerodevapp/kernel-7579-plugins/blob/master/hooks/onlyEntrypoint/src/OnlyEntryPointHook.sol#L12

contracts/account/modules/ERC7579Multisig.sol
```diff
@@ -0,0 +1,294 @@
...
+     * disabled until they are added later.
+     */
+    function onInstall(bytes calldata initData) public virtual {
+        if (initData.length > 32) {
+            // More than just delay parameter
+            (bytes[] memory signers_, uint256 threshold_) = abi.decode(initData, (bytes[], uint256));
+            _addSigners(msg.sender, signers_);
+            _setThreshold(msg.sender, threshold_);
+        }
+    }
```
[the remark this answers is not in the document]
ernestognw: Fixed by adding a check that signers' length is 0

contracts/account/modules/ERC7579MultisigConfirmation.sol
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * ```solidity
+     * abi.encode(bytes signer, bytes signature)
+     * ```
+     *
+     * * signer: The ERC-7913 signer to add
+     * * signature: The signature from this signer confirming their addition
+     *
+     * The function verifies each signature before adding the signer. If any signature is invalid,
+     * the function reverts with ERC7579MultisigInvalidConfirmationSignature.
```
ernestognw, line 42:
```suggestion
     * the function reverts with {ERC7579MultisigInvalidConfirmationSignature}.
```

contracts/account/modules/ERC7579MultisigConfirmation.sol
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * This module ensures that only willing participants can be added as signers to a
+ * multisig by requiring each new signer to provide a valid signature confirming their
+ * consent to be added. Each signer must sign an EIP-712 message to confirm their addition.
+ *
+ * TIP: Use this module to ensure that all guardians in a social recovery or multisig setup have
+ * explicitly agreed to their roles.
+ */
+abstract contract ERC7579MultisigConfirmation is ERC7579Multisig, EIP712 {
+    bytes32 private constant MULTISIG_CONFIRMATION = keccak256("MultisigConfirmation(address account,address module)");
```
ernestognw, line 20: Would it make sense to add a nonce value? Confirmations could be reused in this way, but I don't see why that's an issue if you want to only validate that the signer does exists and can sign. Perhaps expiration?

ernestognw: Try to come up with a concrete opinion. Based on your comments, it seems that 1) it's no a big deal if the confirmation is perpetual and 2) contracts may change, so I'd argue that you're suggesting a configurable optional expiration (e.g. `type().max == expiration`). I would agree though.

ernestognw: Added expiration encoded in the signature field

ernestognw:
Captura de pantalla 2025-05-12 a la(s) 9 50 58 p m

Some review and they're ready to go. I'll proceed adding them to Wizard once merged, and then I'll write some guides on how to add and use ERC-7579 modules

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,445 @@
...
+        if (_schedules[operationId].scheduledAt == 0) return OperationState.Unknown;
+        if (_schedules[operationId].canceled) return OperationState.Canceled;
+        if (_schedules[operationId].executed) return OperationState.Executed;
+        (, uint48 executableAt, uint48 expiresAt) = getSchedule(operationId);
+        if (block.timestamp < executableAt) return OperationState.Scheduled;
+        if (block.timestamp > expiresAt) return OperationState.Expired;
+        return OperationState.Ready;
+    }
+
+    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's {OperationState-Ready}.
```
gonzaotc, line 115: I think the current design of `execute`, `_execute`, and `canExecute` across `ERC7579Executor` and `ERC7579DelayedExecutor` should be slightly re-designed with the objetive of throwing more specific and consistent errors. 

In the current design, `ERC7579Executor.execute` calls `ERC7579DelayedExecutor.canExecute`, where it throws a very specific `ERC7579UnauthorizedExecution` upon a false return. I think this is very misleading when the reason is that the operation is not ready, where `ERC7579ExecutorUnexpectedOperationState` should be thrown instead. 

Consider that this will not only affect this DelayedExecutor implementation, but also any other adding more requirements to `canExecute`. The actual revert reason would also be shadowed.

One option could be to keep `ERC7579Executor.execute` as it is currently, and refactor `canExecute` to use it only for determining "WHO" is authorized and not extra conditions, in order to match the current error thrown more consistently and avoid shadowing other reasons.

If that direction is followed, `canExecute` on this `ERC7579DelayedExecutor` would just return true, since anyone can execute an operation that is ready. 

Note that this would work since `ERC7579DelayedExecutor._execute` already checks the state correctness via `_validateStateBitmap` and throws `ERC7579ExecutorUnexpectedOperationState` if needed. 

I think this design may be right since the developer has all the freedom to extend `_execute` for custom conditions and throw explicit errors, as we are doing already.

gonzaotc:
Review not completed, some comments added

test/account/modules/ERC7579Executor.test.js
```diff
@@ -0,0 +1,79 @@
...
+  encodeSingle,
+  encodeMode,
+  CALL_TYPE_CALL,
+  EXEC_TYPE_DEFAULT,
+} = require('@openzeppelin/contracts/test/helpers/erc7579');
+const { shouldBehaveLikeERC7579Module } = require('./ERC7579Module.behavior');
+
+async function fixture() {
+  // Deploy ERC-7579 validator module
+  const mock = await ethers.deployContract('$ERC7579MultisigExecutorMock');
```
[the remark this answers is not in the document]
gonzaotc: Update: discussed and agreed that is okay to test the `ERC7579Executor` with the `ERC7579MultiExecutorMock`, since `ERC7579Executor` is abstract and would need a mock that implements it's required functions, getting very near to the currently used `ERC7579MultiExecutorMock`, what would be an unnecessary additional mock.

contracts/account/modules/ERC7579DelayedExecutor.sol
gonzaotc: Update, also note that checking the operation state in `canExecute` breaks it's definition provided in `ERC7579Executor` that specifies that it is intended to check if the caller is authorized:

```
* @dev Check if the caller is authorized to execute operations.
* By default, this checks if the caller is the account itself. Derived contracts can
* override this to implement custom authorization logic.
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,445 @@
...
+    ) public view virtual returns (bool) {
+        return account == msg.sender;
+    }
+
+    /// @dev Minimum delay after which {setDelay} takes effect.
+    function minSetback() public view virtual returns (uint32) {
+        return 1 days; // Up to ~136 years
+    }
+
+    /// @dev Delay for a specific account.
```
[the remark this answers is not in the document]
ernestognw: I like this one but the contracts from the OZ vanilla repo are not referenced correctly. See https://github.com/OpenZeppelin/openzeppelin-community-contracts/issues/30

We could also add a hardcoded link to the vanilla docs but the docs don't have the `Time` library yet

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,445 @@
...
+     * See {AccountERC7579-uninstallModule}.
+     */
+    function onUninstall(bytes calldata) public virtual {
+        _config[msg.sender].installed = false;
+        _setDelay(msg.sender, 0, minSetback()); // Avoids immediate downgrades
+        _setExpiration(msg.sender, 0);
+    }
+
+    /**
+     * @dev Internal implementation for setting an account's delay. See {getDelay}.
```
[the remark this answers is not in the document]
ernestognw: Same as https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121/files#r2087781187

contracts/account/modules/ERC7579DelayedExecutor.sol
gonzaotc: Drafted at https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/152

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,445 @@
...
+    function setExpiration(uint32 newExpiration) public virtual {
+        _setExpiration(msg.sender, newExpiration);
+    }
+
+    /**
+     * @dev Schedules an operation to be executed after the account's delay period (see {getDelay}).
+     * Operations are uniquely identified by the combination of `mode`, `executionCalldata`, and `salt`.
+     * See {canSchedule} for authorization checks.
+     */
+    function schedule(address account, Mode mode, bytes calldata executionCalldata, bytes32 salt) public virtual {
```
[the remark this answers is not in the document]
ernestognw: Yeah this is a good point. I'm thinking a threat would be if I get socially engineered to install a module with a pre-scheduled operation, the contract would just look fine and the only way to spot it would be by looking at the indexed events.

We should note it at least

contracts/account/modules/ERC7579DelayedExecutor.sol
ernestognw: Why not just replace the implementation of ERC7579Executor's `execute` to prioritize the error in `_execute`?

```
function execute(
        address account,
        Mode mode,
        bytes calldata executionCalldata,
        bytes32 salt
    ) public virtual returns (bytes[] memory returnData) {
        bool allowed = canExecute(account, mode, executionCalldata, salt);
        returnData = _execute(account, mode, executionCalldata, salt);
        require(allowed, ERC7579UnauthorizedExecution());
        return returnData;
    }
```

test/account/modules/ERC7579DelayedExecutor.test.js
```diff
@@ -0,0 +1,240 @@
...
+    beforeEach(async function () {
+      await this.mockAccountFromEntrypoint.installModule(this.moduleType, this.mock.target, this.installData);
+      await this.mock.$_schedule(this.mockAccount.address, this.mode, this.calldata, salt);
+    });
+
+    it('reverts with ERC7579UnauthorizedExecution before delay passes with any caller', async function () {
+      await expect(
+        this.mock.execute(this.mockAccount.address, this.mode, this.calldata, salt),
+      ).to.be.revertedWithCustomError(this.mock, 'ERC7579UnauthorizedExecution');
+    });
```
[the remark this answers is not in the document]
ernestognw: Yes, this is a good observation, I think we should change the order in which permissions to schedule, execute and cancel

contracts/account/modules/ERC7579DelayedExecutor.sol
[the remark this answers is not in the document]
gonzaotc: I would prefer the call to `schedule` to revert with a precise error instead tbh

contracts/account/modules/ERC7579DelayedExecutor.sol
gonzaotc: While this fix the returned errors, it wastes a lot of gas for unauthorized callers, since they will be paying for the entire execution before getting reverted. I think we should be able to come up with a design that throws precise errors but reverts as soon as possible

ernestognw: > While this fix the returned errors, it wastes a lot of gas for unauthorized callers, since they will be paying for the entire execution before getting reverted.

They can simulate and not execute. If gas is a concern, can you measure it so we can have concrete numbers to discuss?

ernestognw:
I changed the `can*` functions to internal `_validate*` functions, it keeps an internal interface for validation and is more consistent with the delayed executor.

Build is failing since the MultisigMocks need to be adapted, but it's still unclear how:

```
abstract contract ERC7579MultisigExecutorMock is ERC7579Executor, ERC7579Multisig, EIP712 {
    bytes32 private constant EXECUTE_OPERATION =
        keccak256("ExecuteOperation(address account,bytes32 mode,bytes executionCalldata,bytes32 salt)");

    function _validateExecution(
        address account,
        Mode mode,
        bytes calldata executionCalldata,
        bytes32 salt
    ) internal view virtual returns (bool) {
        // We're missing the `signature` here
        return _validateMultisignature(account, _getExecuteTypeHash(account, mode, executionCalldata, salt), signature);
    }

    function _getExecuteTypeHash(
        address account,
        Mode mode,
        bytes calldata executionCalldata,
        bytes32 salt
    ) internal view virtual returns (bytes32) {
        return
            _hashTypedDataV4(
                keccak256(abi.encode(EXECUTE_OPERATION, account, Mode.unwrap(mode), executionCalldata, salt))
            );
    }
}
```

For the modules to be successful, `_validateExecution` and `_validateMultisignature` must combine seamlessly.

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,459 @@
...
+    function canExecute(
+        address account,
+        Mode mode,
+        bytes calldata executionCalldata,
+        bytes32 salt
+    ) public view virtual override returns (bool) {
+        return
+            state(account, mode, executionCalldata, salt) == OperationState.Ready ||
+            super.canExecute(account, mode, executionCalldata, salt);
+    }
```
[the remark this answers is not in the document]
ernestognw: This was a good code smell. I ended up refactoring and making the Executor's have a `_validateExecution` as you suggested in your alternative PR. The `_validateExecution` now not only returns bool checking the validity of the execution but also processes `data` into `calldata`, so that the authorization comes from the provided bytes.

I think we can solve this

contracts/account/modules/ERC7579DelayedExecutor.sol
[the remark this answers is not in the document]
ernestognw: Fixed in 24dad81e9abc6fc66f964b777720b7b99771dc7c. The idea is that the public `schedule` does revert but use cases where someone can pre-schedule operations on a non-deployed account can be done by exposing the internal `_schedule` through some alternative public function.

ernestognw:
> I changed the `can*` functions to internal `_validate*` functions, it keeps an internal interface for validation and is more consistent with the delayed executor.
> 
> Build is failing since the MultisigMocks need to be adapted, but it's still unclear how:
> 
> ```solidity
> // ...
> ```
> 
> For the modules to be successful, `_validateExecution` and `_validateMultisignature` must combine seamlessly.

Fixed by differentiating the `data` argument from the public `schedule`, `execute` and `cancel` functions and `executeCalldata` on their internal versions. The reasoning is that developers must override the public versions if they want to perform validation on `data`, slice it and extract the `executeCalldata` from there. This provides more flexibility and allows combining use cases as in the [ERC7579MultisigMocks](https://github.com/OpenZeppelin/openzeppelin-contracts-experimental/blob/ec0ae9e69881d6df5eb13af382d09079624bf079/contracts/mocks/account/modules/ERC7579MultisigMocks.sol)

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,458 @@
...
+        _validateStateBitmap(id, _encodeStateBitmap(OperationState.Unknown));
+
+        (uint32 executableAfter, , ) = getDelay(account);
+
+        uint48 timepoint = Time.timestamp();
+        _schedules[id].scheduledAt = timepoint;
+        _schedules[id].executableAfter = executableAfter;
+        _schedules[id].expiresAfter = getExpiration(account);
+
+        emit ERC7579ExecutorOperationScheduled(account, id, salt, mode, executionCalldata, timepoint);
```
ernestognw, line 382: This `timepoint` is wrong. It should be the time at which the operation becomes executable. The timestamp is accessible by indexers in the block itself

```suggestion
        emit ERC7579ExecutorOperationScheduled(account, id, salt, mode, executionCalldata, executableAfter + timepoint);
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,458 @@
...
+    /**
+     * @dev Internal version of {schedule} that takes an `account` address as an argument.
+     *
+     * Requirements:
+     *
+     * * The operation must be `Unknown`.
+     *
+     * Emits an {ERC7579ExecutorOperationScheduled} event.
+     */
+    function _schedule(
```
ernestognw, line 366: I'm thinking this one should instead be `_scheduleAt`. So that the public `schedule` uses `_scheduleAt` and passes in the delay. The idea is that developers would have an internal function that allows them to schedule at specific timepoints if needed. The current `_schedule` does not allow for that.

It should have a note, though, making it clear that exposing the function might allow bypassing the regular delay

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,458 @@
...
+        if (_schedules[operationId].executed) return OperationState.Executed;
+        (, uint48 executableAt, uint48 expiresAt) = getSchedule(operationId);
+        if (block.timestamp < executableAt) return OperationState.Scheduled;
+        if (block.timestamp > expiresAt) return OperationState.Expired;
+        return OperationState.Ready;
+    }
+
+    /// @dev Minimum delay after which {setDelay} takes effect.
+    function minSetback() public view virtual returns (uint32) {
+        return 1 days; // Up to ~136 years
```
ernestognw, line 129: I think this should be the same as the default delay though. Otherwise the minSetback is essentially the lower security bound for the account

```suggestion
        return 5 days; // Up to ~136 years
```

ernestognw: Replaced the `defaultDelay` function with the `minSetback`

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -0,0 +1,460 @@
...
+    }
+
+    /// @dev Same as {state}, but for a specific operation id.
+    function state(bytes32 operationId) public view returns (OperationState) {
+        if (_schedules[operationId].scheduledAt == 0) return OperationState.Unknown;
+        if (_schedules[operationId].canceled) return OperationState.Canceled;
+        if (_schedules[operationId].executed) return OperationState.Executed;
+        (, uint48 executableAt, uint48 expiresAt) = getSchedule(operationId);
+        if (block.timestamp < executableAt) return OperationState.Scheduled;
+        if (block.timestamp > expiresAt) return OperationState.Expired;
```
ernestognw, line 126: Off by one

```suggestion
        if (block.timestamp >= expiresAt) return OperationState.Expired;
```

ernestognw:
Happy with the current design and it's good enough for the community version. Given the schedule, I think we should move on and merge so I can start writing ERC-7579 guides that include the social recovery use case. Also I should start playing around with this on Wizard.

PR #122, opened 2025-04-25
Remove `Arrays` library usage from `EnumerableSetExtended`

ernestognw:
### Description

Unfortunately, https://github.com/OpenZeppelin/openzeppelin-contracts/commit/8a4eadea51028d1d8c6eb4252e3248b08a43a740 didn't make it into the 5.3.0 release, which makes available some changes needed for the `EnumerableSetExtended` library. Concretely, the `unsafeSetLength` functions are not available to users.

As a result, https://github.com/OpenZeppelin/contracts-wizard/pull/527 do not compile. This PR addresses the issue by temporarily porting the missing features in 5.3.0

### DIFF
--- a/contracts/utils/structs/EnumerableSetExtended.sol
+++ b/contracts/utils/structs/EnumerableSetExtended.sol
@@ -3,7 +3,6 @@
 
 pragma solidity ^0.8.20;
 
-import {Arrays} from "@openzeppelin/contracts/utils/Arrays.sol";
 import {Hashes} from "@openzeppelin/contracts/utils/cryptography/Hashes.sol";
 
 /**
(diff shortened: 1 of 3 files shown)


arr00:
Given that each function is called just once, does it not make sense to inline the assembly?

contracts/utils/structs/EnumerableSetExtended.sol
```diff
@@ -121,7 +120,19 @@ library EnumerableSetExtended {
...
+        unsafeSetLength(set._values, 0);
+    }
+
+    /**
+     * @dev Helper to set the length of a dynamic array. Directly writing to `.length` is forbidden.
+     *
+     * WARNING: this does not clear elements if length is reduced, of initialize elements if length is increased.
+     */
+    // Replace when these are available in Arrays.sol
+    function unsafeSetLength(string[] storage array, uint256 len) internal {
```
arr00, line 132: I don't think it makes sense to expose this in the library.

ernestognw:
> Given that each function is called just once, does it not make sense to inline the assembly?

Yeah that makes sense. Updated

scripts/generate/templates/EnumerableSetExtended.js
```diff
@@ -124,7 +123,11 @@ function clear(${name} storage set) internal {
     for (uint256 i = 0; i < len; ++i) {
         delete set._positions[set._values[i]];
     }
-    Arrays.unsafeSetLength(set._values, 0);
+    // Replace when these are available in Arrays.sol
+    ${value.type}[] storage array = set._values;
+    assembly ("memory-safe") {
+        sstore(array.slot, len)
+    }
```
arr00, line 130:
```suggestion
    assembly ("memory-safe") {
        sstore(array.slot, 0)
    }
```

arr00 (approved):

PR #124, opened 2025-04-25
Change ERC-7913 title category in utilities docs

ernestognw:
### Description

This makes the ERC-7913 verifiers to show its own section on the sidebar menu. This is also the correct way to do it since it doesn't belong to the "typed signatures" category


ernestognw:
Docs change. Merging

PR #133, opened 2025-05-02
Add ERC7579Validator

gonzaotc:
Following our common modular & composable approach, I propose to decompose ERC7579SignatureValidator into ERC7579Validator to achieve this inheritance structure:

  * ERC7579Validator
    * ERC7579SignatureValidator
    * ERC7579MultisigValidator
  * ERC7579Executor
    * ERC7579MultisigExecutor
  * ERC7579Hook
  * ERC7579Fallback


contracts/account/modules/ERC7579Module.sol
ernestognw, line 1: I don't think this is needed because the events are emitted by the account. These would be redundant if anyone is indexing accounts

contracts/mocks/account/modules/ERC7579ValidatorMock.sol
```diff
@@ -3,43 +3,29 @@
...
     mapping(address sender => address signer) private _associatedSigners;
 
-    function onInstall(bytes calldata data) public virtual override(IERC7579Module, ERC7579ModuleMock) {
+    function onInstall(bytes calldata data) public virtual override(ERC7579Module, IERC7579Module) {
         _associatedSigners[msg.sender] = address(bytes20(data[0:20]));
         super.onInstall(data);
     }
 
-    function onUninstall(bytes calldata data) public virtual override(IERC7579Module, ERC7579ModuleMock) {
+    function onUninstall(bytes calldata data) public virtual override(ERC7579Module, IERC7579Module) {
```
ernestognw, line 18: You know the order of these doesn't matter, right? 😂

contracts/account/modules/ERC7579Module.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    /// @dev Emitted when a module is installed on an account.
+    event ModuleInstalledReceived(address account, bytes data);
+
+    /// @dev Emitted when a module is uninstalled from an account.
+    event ModuleUninstalledReceived(address account, bytes data);
+
+    constructor(uint256 moduleTypeId) {
+        _moduleTypeId = moduleTypeId;
+    }
```
ernestognw, line 41: I don't like that this is forcing an immutable variable for `_moduleTypeId`, I think it's too opinionated and instead, it should be defined in a constant and overrideable with a virtual function

contracts/account/modules/ERC7579Validator.sol
```diff
@@ -0,0 +1,69 @@
...
+ * Developers must implement the onInstall, onUninstall, and {_isValidSignatureWithSender} function in derived contracts to
+ * define the specific signature validation logic.
+ *
+ * Example usage:
+ *
+ * ```solidity
+ * contract MyValidatorModule is ERC7579Validator {
+ *     function onInstall(bytes calldata data) public override {
+ *         // Install logic here
+ *     }
```
ernestognw, line 23: We should not recommend overriding without calling super, someone doing this will skip the `ModuleInstalledReceived` and `ModuleUninstalledReceived` events

contracts/account/modules/ERC7579SignatureValidator.sol
```diff
@@ -92,31 +87,9 @@ contract ERC7579SignatureValidator is IERC7579Validator {
      * making the account unusable. As an account operator, make sure to uninstall to a predefined path
      * in your account that properly side effects of uninstallation.  See {AccountERC7579-uninstallModule}.
      */
-    function onUninstall(bytes calldata) public virtual {
+    function onUninstall(bytes calldata data) public virtual override(ERC7579Module, IERC7579Module) {
         _setSigner(msg.sender, "");
```
ernestognw, line 91: Missing super

ernestognw (commented):
I would only keep the `ERC7579Validator`, in my opinion is the only abstraction that's worth it, the ERC7579Module can't be generalized cleanly to all module types.

I'd test the module before marking the PR as ready. Changing to draft

contracts/account/modules/ERC7579Module.sol
gonzaotc: Yeah I think there is potential interest on measuring installation/uninstallation of modules, mainly for "Module Stores" or wallets offering modules to the user trying to identify popularity, but it is already doable by the events emitted on the Account

ernestognw: Yes. Also these events are not standard. I would lean towards removing the ERC7579Module abstraction

ernestognw (approved):

PR #134, opened 2025-05-04
Add AccountFactory contract

ernestognw:


ernestognw:
imo this is ready to merge, but we're holding back because we're not convinced there should be a "canonical factory", given the actual inner workings of a factory may differ significantly (e.g. initializable clones vs clones with immutable arguments)

I'd agree with merging if someone else has the same opinion

PR #135, opened 2025-05-05
Add ERC7579 validators to API Reference

ernestognw:
A guide on how to use ERC-7579 modules will come up later


ernestognw:
Merging, only docs.

PR #146, opened 2025-05-07
Update node modules and remove unnecessary npm dependencies

arr00:


ernestognw (approved):
The PR added more dependencies to `package-lock.json`, which seem consistent with recent dependency updates in vanilla contracts.

LGTM

PR #149, opened 2025-05-13
Vendor dependencies in `vendored` branch

ernestognw:
Teams installing the repository in Foundry would need to install dependencies regardless. With this proposal, the idea is that they just point to a different branch where the dependencies are vendored to `lib/`

PR #151, opened 2025-05-13
Vendor dependencies to `vendored` branch

ernestognw:
Second attempt of #149 
Tested in `vendor-test` branch: https://github.com/OpenZeppelin/openzeppelin-community-contracts/actions?query=branch%3Atest-vendor

### DIFF
--- a/.github/workflows/vendor.yml
+++ b/.github/workflows/vendor.yml
@@ -1,0 +1,27 @@
+name: vendor dependencies
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  vendor:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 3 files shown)


zkfriendly:
If I understand correctly, based on this approach, the `remappings.txt` file in the `vendored` branch should not contain any `node_modules`. However, the [remappings.txt](https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/vendored/remappings.txt) file has not been changed and still contains `node_modules`.

ernestognw:
Yes, I reset the `vendored` branch after testing it `test-vendor`, so the remappings will get updated once we merge this PR
See that the job trigger is the `master` branch

zkfriendly:
> Yes, I reset the `vendored` branch after testing it `test-vendor`, so the remappings will get updated once we merge this PR See that the job trigger is the `master` branch

thanks for the clarification

ernestognw:
Closing in favor of #153

PR #152, opened 2025-05-13
ERC7579DelayedExecutor - Refactored validation & executing logic

gonzaotc:
Refactor to https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121 with some objetives:

- Validations were previously divided and duplicated across `canExecute` and `_execute`, which were causing inconsistencies if they didn't match. See https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121#discussion_r2089196202

- The functions `_execute`, `_schedule`, `_cancel` were performing validations, while in this approach they're low-level functions that assume prior validation. I think is good to divide validation from the actual executing logic.

- More specific error throwing, where `_validateExecutionRequest` can determine the priority order of errors to throw for different requirement conditions. See https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121#discussion_r2087753905. Additionally, this approach allows to revert with appropriate errors before executing, which results in gas savings for unauthorized callers. See https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121#discussion_r2089214518

- Scheduling was available even if the module wasn't installed. See https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/121#discussion_r2087665961

In this approach, the purposes of each function are very specific and any of them can be overrided for customization:
- `public execute`: allows external calling, calls internal validation and execution.
- `internal _validateExecutionRequest`: validates execution requirements.
- `internal _execute`: low-level execution.


contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -370,6 +358,16 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
...
     }
 
+    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's {OperationState-Ready}.
+    function canExecute(
+        address /* account */,
+        Mode /* mode */,
+        bytes calldata /* executionCalldata */,
+        bytes32 /* salt */
+    ) public view virtual override returns (bool) {
+        return true;
```
ernestognw, line 368: This would allow anyone to, for example, reexecute an operation. I'm also realizing that the current ERC7579DelayedExecutor is allowing the account to do that!

This would allow anyone to, for example, reexecute an operation. I'm also realizing that the current ERC7579DelayedExecutor is allowing the account to do that!

The issue of the order of errors was solved in the other PR, wdyt?

ernestognw: > I'm also realizing that the current ERC7579DelayedExecutor is allowing the account to do that!

Ah no, false alarm, execution is just not given the state. That means that the current implementation is effectively just `state(account, mode, executionCalldata, salt) == OperationState.Ready` haha, cool.

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -122,21 +125,19 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         return OperationState.Ready;
     }
 
-    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's `Ready`.
+    /// @dev See {ERC7579Executor-canExecute}. This implementation allows anyone to execute an operation.
     function canExecute(
```
ernestognw, line 129: There's no reason to keep `canExecute` if it just returns true

contracts/account/modules/ERC7579Executor.sol
```diff
@@ -57,25 +59,54 @@ abstract contract ERC7579Executor is IERC7579Module {
...
+     *```
+     */
+    function _validateExecutionRequest(
+        address account,
+        Mode mode,
+        bytes calldata executionCalldata,
+        bytes32 salt
+    ) internal view virtual {
+        require(canExecute(account, mode, executionCalldata, salt), ERC7579UnauthorizedExecution());
+    }
```
ernestognw, line 89: These functions are defeating the purpose of `canExecute`

contracts/account/modules/ERC7579Executor.sol
```diff
@@ -30,52 +30,51 @@ abstract contract ERC7579Executor is IERC7579Module {
...
-        returnData = _execute(account, mode, executionCalldata, salt); // Prioritize errors thrown in _execute
-        require(allowed, ERC7579UnauthorizedExecution());
-        return returnData;
+        _validateExecutionRequest(account, mode, executionCalldata, salt);
+        return _execute(account, mode, executionCalldata, salt);
     }
 
     /**
-     * @dev Internal version of {execute}. Emits {ERC7579ExecutorOperationExecuted} event.
+     * @dev Low-level internal function to execute an operation. Does not perform any validation checks.
```
ernestognw, line 75:
```suggestion
     * @dev Low-level internal function to execute an operation. Does not perform any validation checks.
```

What does it mean to be "low-level"? I'd like to avoid using ambiguous terms if possible

contracts/account/modules/ERC7579Executor.sol
```diff
@@ -30,52 +30,51 @@ abstract contract ERC7579Executor is IERC7579Module {
...
-    function canExecute(
+    function _validateExecutionRequest(
         address account,
         Mode /* mode */,
         bytes calldata /* executionCalldata */,
         bytes32 /* salt */
-    ) public view virtual returns (bool) {
-        return msg.sender == account;
+    ) internal view virtual {
+        require(msg.sender == account, ERC7579UnauthorizedExecution());
```
ernestognw, line 57: I would insist the issue with `require` statements is that developers can't remove them! The original `canExecute` intentionally returns `bool` to follow the pattern of `condition || super.canExecute()`, which allows to workaround a `require(canSchedule(...))` error. I would rather keep that pattern in the new `_validateExecutionRequest `

```suggestion
    ) internal view virtual returns (bool) {
        return msg.sender == account;
```

contracts/account/modules/ERC7579Executor.sol
```diff
@@ -30,52 +30,51 @@ abstract contract ERC7579Executor is IERC7579Module {
...
-     *    return isAuthorized || super.canExecute(account, mode, executionCalldata, salt);
+     *  ) internal view virtual override {
+     *    bool conditionMet = ...; // custom logic to check condition
+     *    require(conditionMet, ERC7579ExecutorConditionNotMet());
+     *    super._validateExecutionRequest(account, mode, executionCalldata, salt);
      *  }
      *```
      */
-    function canExecute(
+    function _validateExecutionRequest(
```
ernestognw, line 51: I would rename to just

```suggestion
    function _validateExecution(
```

It may be better not to introduce the concept of a "request"

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -398,26 +415,21 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
...
     }
 
     /**
-     * @dev Internal version of {cancel} that takes an `account` address as an argument.
-     *
-     * Requirements:
+     * @dev Low-level internal function to cancel an operation. Does not perform any validation checks.
      *
-     * * The operation must be `Scheduled` or `Ready`.
+     * Note: Canceled operations can't be rescheduled.
```
ernestognw, line 427: It needs to be `NOTE` so that it's picked up by the docs engine 😄 

```suggestion
     * NOTE: Canceled operations can't be rescheduled.
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -122,72 +125,106 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
...
-        Mode /* mode */,
-        bytes calldata /* executionCalldata */,
-        bytes32 /* salt */
-    ) public view virtual returns (bool) {
-        return account == msg.sender;
+        Mode mode,
+        bytes calldata executionCalldata,
+        bytes32 salt
+    ) internal view virtual {
+        require(_config[account].installed, ERC7579ExecutorModuleNotInstalled());
```
ernestognw, line 222: I would insist on keeping this case as a no-op. It's a valid use case to extend and allow scheduling on contracts that haven't installed the module. For example, a factory contract could deploy accounts with create2 and install this module with pre-scheduled payments (ie. subscriptions). In that case it becomes impossible to build that case on top of this contract

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -122,72 +125,106 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         return OperationState.Ready;
     }
 
-    /// @dev See {ERC7579Executor-canExecute}. Allows anyone to execute an operation if it's `Ready`.
-    function canExecute(
+    /**
+     * @dev See {ERC7579Executor-_validateExecutionRequest}. This implementation only validates
+     * the operation is in {OperationState.Ready}. Any caller can execute the operation if it is
```
ernestognw, line 130: The docs don't render enums. I noticed recently haha

```suggestion
     * the operation is in `Ready`. Any caller can execute the operation if it is
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -122,72 +125,106 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
...
+    function _validateExecutionRequest(
         address account,
         Mode mode,
         bytes calldata executionCalldata,
         bytes32 salt
-    ) public view virtual override returns (bool) {
-        return
-            state(account, mode, executionCalldata, salt) == OperationState.Ready ||
-            super.canExecute(account, mode, executionCalldata, salt);
+    ) internal view virtual override {
```
ernestognw, line 155: The function is not calling super. Depending on the linearization order it would make some combinations to skip important implementations of `_validateExecutionRequest` without the user noticing.

For example, if we have a contract B that implements `_validateExecutionRequest` too (it calls super), perhaps allows a "module admin" to execute on certain accounts, then:

```solidity
contract A is ERC7579DelayedExecutor, B {
  // Will run B's `_validateExecutionRequest` and fallbacks to ERC7579DelayedExecutor's
}

contract A is B, ERC7579DelayedExecutor {
  // Will run ERC7579DelayedExecutor's `_validateExecutionRequest` and stop. Making B unreachable.
}
```

contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -398,26 +415,21 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         bytes32 salt
     ) internal virtual override returns (bytes[] memory returnData) {
         bytes32 id = hashOperation(account, mode, executionCalldata, salt);
-        _validateStateBitmap(id, _encodeStateBitmap(OperationState.Ready));
```
ernestognw, line 401: It's now possible to call `_execute` without being in a valid state. That's dangerous as it would allow any other route to call `_execute` to bypass the whole delay system

gonzaotc: Added clear notes

contracts/account/modules/ERC7579DelayedExecutor.sol
gonzaotc: Right, thank you. I added the super call in `DelayedExecutor` and removed the requirement from the base `Executor` `_validateExecution`

PR #157, opened 2025-05-17
Add ERC-7579 modules docs and EIP-7702 note in accounts docs

ernestognw:
Follow up to #121

### DIFF
--- a/contracts/account/modules/ERC7579DelayedExecutor.sol
+++ b/contracts/account/modules/ERC7579DelayedExecutor.sol
@@ -32,6 +32,12 @@ import {ERC7579Executor} from "./ERC7579Executor.sol";
  * after a transition period defined by the current delay or {minSetback}, whichever
  * is longer.
  *
+ * ==== Authorization
+ *
+ * Authorization for scheduling and canceling operations is controlled through the {_validateSchedule}
+ * and {_validateCancel} functions. These functions can be overridden to implement custom
+ * authorization logic, such as requiring specific signers or roles.
+ *
  * TIP: Use {_scheduleAt} to schedule operations at a specific points in time. This is
  * useful to pre-schedule operations for non-deployed accounts (e.g. subscriptions).
  */
@@ -93,12 +99,6 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         OperationState currentState,
         bytes32 allowedStates
     );
-
-    /// @dev The operation is not authorized to be canceled.
-    error ERC7579ExecutorUnauthorizedCancellation();
-
-    /// @dev The operation is not authorized to be scheduled.
-    error ERC7579ExecutorUnauthorizedSchedule();
 
     /// @dev The module is not installed on the account.
     error ERC7579ExecutorModuleNotInstalled();
@@ -232,10 +232,9 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
      */
     function schedule(address account, bytes32 salt, bytes32 mode, bytes calldata data) public virtual {
         require(_config[account].installed, ERC7579ExecutorModuleNotInstalled());
-        bool allowed = _validateSchedule(account, salt, mode, data);
+        _validateSchedule(account, salt, mode, data);
         (uint32 executableAfter, , ) = getDelay(account);
         _scheduleAt(account, salt, mode, data, Time.timestamp(), executableAfter);
-        require(allowed, ERC7579ExecutorUnauthorizedSchedule());
     }
 
     /**
@@ -243,9 +242,8 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
      * scheduled the operation. See {_cancel}.
      */
     function cancel(address account, bytes32 salt, bytes32 mode, bytes calldata data) public virtual {
-        bool allowed = _validateCancel(account, salt, mode, data);
+        _validateCancel(account, salt, mode, data);
         _cancel(account, mode, data, salt); // Prioritize errors thrown in _cancel
-        require(allowed, ERC7579ExecutorUnauthorizedCancellation());
     }
 
     /**
@@ -267,32 +265,30 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         _setExpiration(msg.sender, 0);
     }
 
-    /// @inheritdoc ERC7579Executor
+    /**
+     * @dev Returns `data` as the execution calldata. See {ERC7579Executor-_execute}.
+     *
+     * NOTE: This function relies on the operation state validation in {_execute} for
+     * authorization. Extensions of this module should override this function to implement
+     * additional validation logic if needed.
+     */
     function _validateExecution(
         address /* account */,
         bytes32 /* salt */,
         bytes32 /* mode */,
         bytes calldata data
-    ) internal virtual override returns (bool valid, bytes calldata executionCalldata) {
-        return (true, data); // Anyone can execute, the state validation of the operation is enough
+    ) internal virtual override returns (bytes calldata) {
+        return data;
     }
 
     /**
-     * @dev Whether the caller is authorized to cancel operations.
-     * By default, this checks if the caller is the account itself. Derived contracts can
-     * override this to implement custom authorization logic.
+     * @dev Validates whether an operation can be canceled.
      *
      * Example extension:
      *
      * ```solidity
-     *  function _validateCancel(
-     *     address account,
-     *     bytes32 salt,
-     *     bytes32 mode,
-     *     bytes calldata data
-     *  ) internal override returns (bool) {
-     *    bool isAuthorized = ...; // custom logic to check authorization
-     *    return isAuthorized || super._validateCancel(account, salt, mode, data);
+     *  function _validateCancel(address account, bytes32 salt, bytes32 mode, bytes calldata data) internal override {
+     *    // e.g. require(msg.sender == account);
      *  }
      *```
      */
@@ -301,26 +297,16 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         bytes32 /* salt */,
         bytes32 /* mode */,
         bytes calldata /* data */
-    ) internal virtual returns (bool) {
-        return account == msg.sender;
-    }
+    ) internal virtual;
 
     /**
-     * @dev Whether the caller is authorized to schedule operations.
-     * By default, this checks if the caller is the account itself. Derived contracts can
-     * override this to implement custom authorization logic.
+     * @dev Validates whether an operation can be scheduled.
      *
      * Example extension:
      *
      * ```solidity
-     *  function _validateSchedule(
-     *     address account,
-     *     bytes32 salt,
-     *     bytes32 mode,
-     *     bytes calldata data
-     *  ) internal override returns (bool) {
-     *    bool isAuthorized = ...; // custom logic to check authorization
-     *    return isAuthorized || super._validateSchedule(account, salt, mode, data);
+     *  function _validateSchedule(address account, bytes32 salt, bytes32 mode, bytes calldata data) internal override {
+     *    // e.g. require(msg.sender == account);
      *  }
      *```
      */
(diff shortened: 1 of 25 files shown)


contracts/account/modules/ERC7579DelayedExecutor.sol
```diff
@@ -273,7 +273,7 @@ abstract contract ERC7579DelayedExecutor is ERC7579Executor {
         bytes32 /* salt */,
         bytes32 /* mode */,
         bytes calldata data
-    ) internal view virtual override returns (bool valid, bytes calldata executionCalldata) {
+    ) internal virtual override returns (bool valid, bytes calldata executionCalldata) {
```
ernestognw, line 276: I realized it is a valid use case to consume a nonce here. In the case a contract extends and perform other storage updates during cancellation, they should make their signatures for cancellations replayable to avoid unexpected side effects of recancelling.

Anyway, it's a long discussion and I think it's just reasonable to remove the `view` modifier in the `_validate*` functions of the ERC7579Executor and ERC7579DelayedExecutor

ernestognw: Also just realized it's unnecessary to return a `boolean valid` then. That pattern applies when the function is view.

ernestognw: Opened #158 to solve these first

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
+= Account Modules
+
+Smart accounts built with https://eips.ethereum.org/EIPS/eip-7579[ERC-7579] provide a standardized way to extend account functionality through modules (i.e. smart contract instances). This architecture allows accounts to support various features that are compatible with a wide variety of account implementations. See https://erc7579.com/accounts[compatible modules].
```
gonzaotc, line 3: The link is pointing to accounts implementations, not modules
```suggestion
Smart accounts built with https://eips.ethereum.org/EIPS/eip-7579[ERC-7579] provide a standardized way to extend account functionality through modules (i.e. smart contract instances). This architecture allows accounts to support various features that are compatible with a wide variety of account implementations. See https://erc7579.com/accounts[compatible accounts].
```

arr00:
We should add the interfaces to the API section of the docs.

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+Modules can implement multiple types simultaneously, which means you could combine an executor module with hooks to enforce behaviors on an account, such as maintaining ERC-20 approvals or preventing the removal of certain permissions.
+
+See https://erc7579.com/modules[popular module implementations].
+
+==== Building Custom Modules
+
+The library provides _standard composable modules_ as building blocks with an internal API for developers. By combining these components, you can create a rich set of variants without including unnecessary features.
+
+A good starting point is the xref:api:account.adoc#ERC7579Executor[ERC7579Executor] or xref:api:account.adoc#ERC7579Validator[ERC7579Validator], which include an opinionated base layer easily combined with other abstract modules. Hooks and fallback handlers are more straightforward to implement directly from interfaces:
```
arr00, line 37:
```suggestion
A good starting point is the xref:api:account.adoc#ERC7579Executor[`ERC7579Executor`] or xref:api:account.adoc#ERC7579Validator[`ERC7579Validator`], which include an opinionated base layer easily combined with other abstract modules. Hooks and fallback handlers are more straightforward to implement directly from interfaces:
```

ernestognw:
> We should add the interfaces to the API section of the docs.

Yeah, the issue with those is a bit more complex since they come from the vanilla repository. #30

contracts/mocks/docs/account/modules/MyERC7579SocialRecovery.sol
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    // Data encoding: [uint16(executionCalldataLength), executionCalldata, signature]
+    function _validateExecution(
+        address account,
+        bytes32 salt,
+        bytes32 mode,
+        bytes calldata data
+    ) internal override returns (bytes calldata) {
+        uint16 executionCalldataLength = uint16(uint256(bytes32(data[0:2]))); // First 2 bytes are the length
```
arr00, line 26:
```suggestion
        uint16 executionCalldataLength = uint16(bytes2(data[0:2])); // First 2 bytes are the length
```
Would this work?

ernestognw: Yes, it does!

docs/modules/ROOT/pages/account-modules.adoc
gonzaotc, line 1: I'm not sure where and exactly how, but I would like to briefly mention the benefits of implicitly using ERC-7913 building blocks on the Social Recovery example, since I think, and this is very opinionated, that it is a valuable feature that makes a compelling reason enough to choose this over other implementations, on top of the great composability and extensibility provided. Wdyt? @ernestognw

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
...
+----
+
+For enhanced security, you can extend this foundation with scheduling, delays, and cancellations using xref:api:account.adoc#ERC7579DelayedExecutor[ERC7579DelayedExecutor]. This allows guardians to schedule recovery operations with a time delay, providing a security window to detect and cancel suspicious recovery attempts before they execute:
+
+[source,solidity]
+----
+include::api:example$account/modules/MyERC7579DelayedSocialRecovery.sol[]
+----
+
+NOTE: The delayed executor's signature validation doesn't require a nonce since operations are uniquely identified by their xref:api:account.account#ERC7579DelayedExecutor-hashOperation-address-bytes32-bytes32-bytes-[operation id] and cannot be scheduled twice.
```
arr00, line 119:
```suggestion
NOTE: The delayed executor's signature validation doesn't require a nonce since operations are uniquely identified by their xref:api:account.adoc#ERC7579DelayedExecutor-hashOperation-address-bytes32-bytes32-bytes-[operation id] and cannot be scheduled twice.
```

docs/modules/ROOT/pages/account-modules.adoc
ernestognw: I don't think standards should be _sold_. I mean that a good standard solves an issue cleanly, and I think this is the case. The recommendation should be **use ERC-7913** because it's simply the best way to approach the issue of supporting arbitrary authorization types.

Instead, maybe users would benefit from understanding how to bring ERC-7913 to their use cases. They should be guided about when to use each variant (SingerERC7913, MultisignerERC7913, ERC7579Multisg, ERC7579Utils, ERC7913RSAVerifier, ERC7913ZKVerifier, ERC7913P256Verifier and beyond)

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
...
+
+These implementations demonstrate how to build progressively more secure social recovery mechanisms, from basic multi-signature recovery to time-delayed recovery with cancellation capabilities.
+
+For additional functionality, developers can use:
+
+* xref:api:account.adoc#ERC7579MultisigWeighted[ERC7579MultisigWeighted] to assign different weights to signers
```
arr00, line 125:
```suggestion
* xref:api:account.adoc#ERC7579MultisigWeighted[`ERC7579MultisigWeighted`] to assign different weights to signers
```

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
...
+* xref:api:account.adoc#ERC7579MultisigConfirmation[ERC7579MultisigConfirmation] to implement a confirmation system that verifies signatures when adding signers
```
arr00, line 126:
```suggestion
* xref:api:account.adoc#ERC7579MultisigConfirmation[`ERC7579MultisigConfirmation`] to implement a confirmation system that verifies signatures when adding signers
```

docs/modules/ROOT/pages/account-modules.adoc
```diff
@@ -0,0 +1,126 @@
...
+* Confirmation requirements
+
+A simple first approach combines multiple signatures from guardians with execution capabilities. This provides a basic foundation that can be extended with more sophisticated features:
+
+[source,solidity]
+----
+include::api:example$account/modules/MyERC7579SocialRecovery.sol[]
+----
+
+For enhanced security, you can extend this foundation with scheduling, delays, and cancellations using xref:api:account.adoc#ERC7579DelayedExecutor[ERC7579DelayedExecutor]. This allows guardians to schedule recovery operations with a time delay, providing a security window to detect and cancel suspicious recovery attempts before they execute:
```
arr00, line 112:
```suggestion
For enhanced security, you can extend this foundation with scheduling, delays, and cancellations using xref:api:account.adoc#ERC7579DelayedExecutor[`ERC7579DelayedExecutor`]. This allows guardians to schedule recovery operations with a time delay, providing a security window to detect and cancel suspicious recovery attempts before they execute:
```

docs/modules/ROOT/pages/account-modules.adoc
ernestognw: Oh good catch, it's actually the other way around

contracts/mocks/docs/account/modules/MyERC7579DelayedSocialRecovery.sol
```diff
@@ -0,0 +1,57 @@
...
+
+    function isModuleType(uint256 moduleTypeId) public pure override(ERC7579Executor, ERC7579Validator) returns (bool) {
+        return ERC7579Executor.isModuleType(moduleTypeId) || ERC7579Executor.isModuleType(moduleTypeId);
+    }
+
+    // Data encoding: [uint16(executorArgsLength), executorArgs, uint16(multisigArgsLength), multisigArgs]
+    function onInstall(bytes calldata data) public override(ERC7579DelayedExecutor, ERC7579Multisig) {
+        uint16 executorArgsLength = uint16(bytes2(data[0:2])); // First 2 bytes are the length
+        bytes calldata executorArgs = data[2:2 + executorArgsLength]; // Next bytes are the args
+        uint16 multisigArgsLength = uint16(bytes2(data[2 + executorArgsLength:])); // Next 2 bytes are the length
```
arr00, line 24:
```suggestion
        uint16 multisigArgsLength = uint16(bytes2(data[2 + executorArgsLength:4 + executorArgsLength])); // Next 2 bytes are the length
```
End seems to have gotten lost here

contracts/mocks/docs/account/modules/MyERC7579DelayedSocialRecovery.sol
```diff
@@ -0,0 +1,57 @@
...
+        bytes calldata multisigArgs = data[2 + executorArgsLength + 2:2 + executorArgsLength + 2 + multisigArgsLength]; // Next bytes are the args
```
arr00, line 25:
```suggestion
        bytes calldata multisigArgs = data[4 + executorArgsLength:4 + executorArgsLength + multisigArgsLength]; // Next bytes are the args
```

arr00: Took out the supercomputer and evaluated 2 + 2 = 4

docs/modules/ROOT/pages/account-modules.adoc
ernestognw: I think I get what you say. Is it that the reasons of why using an ERC7579Multisig aren't explained? That'd be true.

I just updated with a bit more context in https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/157/commits/8980e0c2a04fc6dd0fef49fab3284f7c62762958

arr00 (approved):

ernestognw:
Thanks for reviewing both! Merging since the previous change was minimal

PR #159, opened 2025-05-17
Simplify ERC-7579 validator modules

ernestognw:
While working on the documentation (#157) and a PoC in Wizard for ERC-7579 modules. I noticed that the `_rawSignatureValidationWithSender` was receiving `sender` when that's only relevant in the `isValidSignatureWithSender` specified in ERC-7579.

Also, I noticed that if converted into a basic `_rawERC7579Validation(address account, bytes32 hash, bytes calldata signature)`, then it was possible to reuse that function in ERC7579Multisig (and all derived multisig modules). Given that module types are not mutually exclusive, then it makes sense to convert multisigs into validators.


ernestognw:
Looks good. Merging to unblock a PR on Wizard

PR #160, opened 2025-05-24
Add husky to devDependencies

ernestognw:


ernestognw:
Required for https://github.com/OpenZeppelin/contracts-wizard/pull/553

arr00:
@ernestognw why isn't husky pulled from the vanilla dependencies like everything else?

PR #171, opened 2025-06-09
Update Crosschain contract: transition from CAIP-10 to ERC-7930 

Amxx:
Based on #169

### DIFF
--- a/contracts/crosschain/ERC7786Aggregator.sol
+++ b/contracts/crosschain/ERC7786Aggregator.sol
@@ -5,19 +5,17 @@ pragma solidity ^0.8.27;
 import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
 import {EnumerableSet} from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 import {Address} from "@openzeppelin/contracts/utils/Address.sol";
-import {CAIP2} from "@openzeppelin/contracts/utils/CAIP2.sol";
-import {CAIP10} from "@openzeppelin/contracts/utils/CAIP10.sol";
 import {Pausable} from "@openzeppelin/contracts/utils/Pausable.sol";
-import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
 import {IERC7786GatewaySource, IERC7786Receiver} from "../interfaces/IERC7786.sol";
+import {ERC7930} from "../utils/ERC7930.sol";
 
 /**
  * @dev N of M gateway: Sends your message through M independent gateways. It will be delivered to the receiver by an
  * equivalent aggregator on the destination chain if N of the M gateways agree.
  */
 contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, Pausable {
-    using EnumerableSet for *;
-    using Strings for *;
+    using EnumerableSet for EnumerableSet.AddressSet;
+    using ERC7930 for bytes;
 
     struct Outbox {
         address gateway;
@@ -30,7 +28,7 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, 
         bool executed;
     }
 
-    event OutboxDetails(bytes32 indexed outboxId, Outbox[] outbox);
+    event OutboxDetails(bytes32 indexed sendId, Outbox[] outbox);
     event Received(bytes32 indexed receiveId, address gateway);
     event ExecutionSuccess(bytes32 indexed receiveId);
     event ExecutionFailed(bytes32 indexed receiveId);
@@ -41,7 +39,7 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, 
     error ERC7786AggregatorValueNotSupported();
     error ERC7786AggregatorInvalidCrosschainSender();
     error ERC7786AggregatorAlreadyExecuted();
-    error ERC7786AggregatorRemoteNotRegistered(string caip2);
+    error ERC7786AggregatorRemoteNotRegistered(bytes2 chainType, bytes chainReference);
     error ERC7786AggregatorGatewayAlreadyRegistered(address gateway);
     error ERC7786AggregatorGatewayNotRegistered(address gateway);
     error ERC7786AggregatorThresholdViolation();
@@ -52,7 +50,7 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, 
      ****************************************************************************************************************/
 
     /// @dev address of the matching aggregator for a given CAIP2 chain
-    mapping(string caip2 => string) private _remotes;
+    mapping(bytes2 chainType => mapping(bytes chainReference => bytes addr)) private _remotes;
 
     /// @dev Tracking of the received message pending final delivery
     mapping(bytes32 id => Tracker) private _trackers;
@@ -69,8 +67,8 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, 
     /****************************************************************************************************************
      *                                        E V E N T S   &   E R R O R S                                         *
      ****************************************************************************************************************/
-    event RemoteRegistered(string chainId, string aggregator);
-    error RemoteAlreadyRegistered(string chainId);
+    event RemoteRegistered(bytes remote);
+    error RemoteAlreadyRegistered(bytes remote);
 
     /****************************************************************************************************************
      *                                              F U N C T I O N S                                               *
@@ -92,18 +90,19 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable, 
     /// @inheritdoc IERC7786GatewaySource
     /// @dev Using memory instead of calldata avoids stack too deep errors
     function sendMessage(
-        string calldata destinationChain,
-        string memory receiver,
-        bytes memory payload,
-        bytes[] memory attributes
-    ) public payable virtual whenNotPaused returns (bytes32 outboxId) {
+        bytes calldata recipient, // Binary Interoperable Address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) public payable virtual whenNotPaused returns (bytes32 sendId) {
         if (attributes.length > 0) revert UnsupportedAttribute(bytes4(attributes[0]));
         if (msg.value > 0) revert ERC7786AggregatorValueNotSupported();
+
         // address of the remote aggregator, revert if not registered
-        string memory aggregator = getRemoteAggregator(destinationChain);
+        bytes memory aggregator = getRemoteAggregator(recipient);
+        bytes memory sender = ERC7930.formatEvmV1(block.chainid, msg.sender);
 
         // wrapping the payload
-        bytes memory wrappedPayload = abi.encode(++_nonce, msg.sender.toChecksumHexString(), receiver, payload);
+        bytes memory wrappedPayload = abi.encode(++_nonce, sender, recipient, payload);
 
         // Post on all gateways
         Outbox[] memory outbox = new Outbox[](_gateways.length());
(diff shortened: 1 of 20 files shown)


contracts/utils/ERC7930.sol
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    function formatEvmV1(uint256 chainid, address addr) internal pure returns (bytes memory) {
+        if (chainid <= type(uint8).max)
+            return abi.encodePacked(bytes2(0x0001), bytes2(0x0000), uint8(1), uint8(chainid), uint8(20), addr);
+        if (chainid <= type(uint16).max)
+            return abi.encodePacked(bytes2(0x0001), bytes2(0x0000), uint8(2), uint16(chainid), uint8(20), addr);
+        if (chainid <= type(uint24).max)
+            return abi.encodePacked(bytes2(0x0001), bytes2(0x0000), uint8(3), uint24(chainid), uint8(20), addr);
+        else revert();
```
luiz-lvj, line 44: If we want to enable all EVM chains in this function, there are chainids higher than`2ˆ24-1 = 16777215`, as we can see [here](https://github.com/DefiLlama/chainlist/blob/b75b9610ebeb86db9167cd41c533ed82d7bb8917/constants/extraRpcs.js#L5269).

I understand that in these cases it's possible to use formatV1 instead, so should we add some natspec?

contracts/utils/ERC7930.sol
```diff
@@ -0,0 +1,127 @@
...
+
+import {Bytes} from "@openzeppelin/contracts/utils/Bytes.sol";
+import {Calldata} from "@openzeppelin/contracts/utils/Calldata.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+
+// 0x00010000010114D8DA6BF26964AF9D7EED9E03E53415D37AA96045
+// 0x000100022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef02005333498d5aea4ae009585c43f7b8c30df8e70187d4a713d134f977fc8dfe0b5
+// 0x000100000014D8DA6BF26964AF9D7EED9E03E53415D37AA96045
+// 0x000100022045296998a6f8e2a784db5d9f95e18fc23f70441a1039446801089879b08c7ef000
+// 0x0001000002A4B114D8DA6BF26964AF9D7EED9E03E53415D37AA96045
```
luiz-lvj, line 13:
```suggestion

```

contracts/utils/ERC7930.sol
```diff
@@ -0,0 +1,127 @@
...
+
```
luiz-lvj, line 14:
```suggestion

/**
* @dev Collection of functions related to ERC7930, Interoperable Addresses
*/
```

contracts/utils/ERC7930.sol
```diff
@@ -0,0 +1,127 @@
...
+library ERC7930 {
+    using SafeCast for uint256;
+    using Bytes for bytes;
+
+    error ERC7930ParsingError(bytes);
+
```
luiz-lvj, line 20:
```suggestion
/**
* @dev Format a `bytes2` chain type, `bytes` chain reference and a `bytes` address (in memory) into an interoperable address
*/
```

contracts/utils/ERC7930.sol
```diff
@@ -0,0 +1,127 @@
...
+            abi.encodePacked(
+                bytes2(0x0001),
+                chainType,
+                chainReference.length.toUint8(),
+                chainReference,
+                addr.length.toUint8(),
+                addr
+            );
+    }
+
```
luiz-lvj, line 36:
```suggestion
/** 
* @dev Format a `uint256` chain id and an `address` `addr`into an interoperable address

* NOTE: for chainid > uint24, use `formatV1` instead
*/
```

test/helpers/chains.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+  try {
+    return ethers.toBeHex(ethers.decodeBase64(value));
+  } catch {
+    /* Do nothing */
+  }
+  throw new Error(`Unable to decode ${value}`);
+};
+
+const formatERC7913v1 = ({ type, reference, address }) => {
```
luiz-lvj, line 59: Shouldn't it be ERC7930?
```suggestion
const formatERC7930v1 = ({ type, reference, address }) => {
```

test/helpers/chains.js
```diff
@@ -0,0 +1,136 @@
...
+      addressHex,
+    ],
+  );
+  const checksum = ethers.keccak256(ethers.getBytes(binary).slice(2)).slice(2, 10).toUpperCase();
+  const name = [address, '@', type, reference && `:${reference}`, '#', checksum].filter(Boolean).join('');
+
+  return { binary, name, fields: { type, reference, address, checksum } };
+};
+
+const parseERC7913v1 = input => {
```
luiz-lvj, line 86: Same as above

Amxx:
Note: 
Please don't look at 7930.sol contract and review this PR instead https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5736. The goal is to have a clear/documented version in the vanilla repo and import it here.

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -182,18 +170,21 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
      * some value for unknown reason. In that case we want to register this gateway having delivered the message and
      * not revert. Any value accrued that way can be recovered by the admin using the {sweep} function.
      */
+    // slither-disable-next-line reentrancy-no-eth
     function executeMessage(
-        string calldata /*messageId*/, // gateway specific, empty or unique
-        string calldata sourceChain, // CAIP-2 chain identifier
-        string calldata sender, // CAIP-10 account address (does not include the chain identifier)
+        bytes32 /*receiveId*/,
+        bytes calldata remote, // Binary Interoperable Address
```
luiz-lvj, line 176: why change from `sender` to `remote`?

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -253,10 +245,18 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
...
+        (bytes2 chainType, bytes memory chainReference, ) = chain.parseV1();
+        return getRemoteAggregator(chainType, chainReference);
+    }
+
+    function getRemoteAggregator(
+        bytes2 chainType,
+        bytes memory chainReference
+    ) public view virtual returns (bytes memory) {
+        bytes memory addr = _remotes[chainType][chainReference];
+        if (bytes(addr).length == 0) revert ERC7786AggregatorRemoteNotRegistered(chainType, chainReference);
```
luiz-lvj, line 258:
```suggestion
        require(bytes(addr).length != 0, ERC7786AggregatorRemoteNotRegistered(chainType, chainReference));
```

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -92,31 +90,27 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
...
-        string calldata destinationChain,
-        string memory receiver,
-        bytes memory payload,
-        bytes[] memory attributes
-    ) public payable virtual whenNotPaused returns (bytes32 outboxId) {
+        bytes calldata recipient, // Binary Interoperable Address
+        bytes calldata payload,
+        bytes[] calldata attributes
+    ) public payable virtual whenNotPaused returns (bytes32 sendId) {
         if (attributes.length > 0) revert UnsupportedAttribute(bytes4(attributes[0]));
```
luiz-lvj, line 97: here and in a few other places:
```suggestion
        require(attributes.length == 0, UnsupportedAttribute(bytes4(attributes[0]));
```

luiz-lvj (approved):

contracts/crosschain/ERC7786Aggregator.sol
Amxx: if we do `require(<condition>, <error>);`, the error is encoded regardless of the value of the condition. So basically, if there are no attributes (and thus no error) the system will still try to the `attributes[0]` to build the error. This causes a panic.

contracts/crosschain/ERC7786Aggregator.sol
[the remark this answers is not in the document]
Amxx: The logic of the aggregator is to have a 7786 interface, both on the sending, and on the receiving end. The idea is also that the sub-gateways are 7786.

Maybe if the sub-gateways are more than just 7786 (7965?), then it would be possible to leverage that to internal things differently ... but in the end, the external part of the adaptor should be at least 7786.

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -30,18 +28,18 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
...
+    event OutboxDetails(bytes32 indexed sendId, Outbox[] outbox);
     event Received(bytes32 indexed receiveId, address gateway);
     event ExecutionSuccess(bytes32 indexed receiveId);
     event ExecutionFailed(bytes32 indexed receiveId);
     event GatewayAdded(address indexed gateway);
     event GatewayRemoved(address indexed gateway);
     event ThresholdUpdated(uint8 threshold);
 
-    error ERC7786AggregatorValueNotSupported();
+    error UnsupportedNativeTransfer();
```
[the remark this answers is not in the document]
Amxx: We can rename that error, but the aggregator is an independant contract that does calls to gateways. The aggregator should NOT inherit from another gateway implementation.

I think the analogy is ERC-4626 <> ERC-20. 
ERC-4626 implements the ERC-20 interface, and all the corresponding logic, but you would not do `contract A is ERC4626, WETH, ... {}`

contracts/crosschain/ERC7786Aggregator.sol
```diff
@@ -92,31 +90,29 @@ contract ERC7786Aggregator is IERC7786GatewaySource, IERC7786Receiver, Ownable,
...
+            revert UnsupportedAttribute(attributes[0].length < 0x04 ? bytes4(0) : bytes4(attributes[0][0:4]));
+
         // address of the remote aggregator, revert if not registered
-        string memory aggregator = getRemoteAggregator(destinationChain);
+        bytes memory aggregator = getRemoteAggregator(recipient);
+        bytes memory sender = InteroperableAddress.formatEvmV1(block.chainid, msg.sender);
 
         // wrapping the payload
-        bytes memory wrappedPayload = abi.encode(++_nonce, msg.sender.toChecksumHexString(), receiver, payload);
+        bytes memory wrappedPayload = abi.encode(++_nonce, sender, recipient, payload);
```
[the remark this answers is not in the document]
ernestognw: > not attached to an account

ah right, that's true

> Should we rename "nonce" to "counter"

haha noo, let's leave it as "nonce", that's a deep rabbit hole 😂

contracts/crosschain/ERC7786Aggregator.sol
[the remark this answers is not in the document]
Amxx: > I think it assumes all messages come from the same source chain

The design of **this** aggregator is all about having a message come from one source on one chain, and getting to one destination on one other chain through multiple (indenpandant) gateways.

Someone wanting to send one message to many chains is a different usecase, that will require a different aggregator. That is another contract ... and we will eventually have both.

contracts/crosschain/ERC7786Aggregator.sol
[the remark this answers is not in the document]
Amxx: Its going to be renamed anyway. The current marking name is "ERC7786OpenBridge". Makes little sens in my opinion.

I think "multibridge" might be a good option.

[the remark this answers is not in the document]
ernestognw: More in favor of MultiBridge after past discussions but happy with OpenBridge since marketing has gone with that one.

PR #174, opened 2025-06-12
Add an ERC7802Bridge

Amxx:
Based on #171


contracts/crosschain/ERC7802Bridge.sol
ernestognw, line 1: I believe the bridge should be nativel ERC-7786

Amxx: I strongly disagree.

IMO we should be able to put that bridge on top of any protocol without having to modify/reconfigure it. That means separating the concerns between the token handling layer, and the gateway layer. In particular, this separation is necessary to build 7802 token bridges on top of 7786 aggregators, or to build 7802 token bridges on top of protocols that natively implement 7786.

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,156 @@
...
+    function getRemoteToken(address token, bytes memory chain) public view virtual returns (bytes memory remote) {
+        remote = _remoteTokens[token][chain];
+        require(remote.length > 0, ERC7802BridgeMissingRemote(token, chain));
+    }
+
+    // TODO: check that bridge includes an address ? prevent override ?
+    function registerRemote(address token, bytes memory bridge, bytes memory remote) public virtual onlyOwner {
+        (bytes2 chainType, bytes memory chainReference, ) = bridge.parseV1();
+        bytes memory chain = InteroperableAddress.formatV1(chainType, chainReference, "");
+
```
Amxx, line 65:
```suggestion

// check that there is a gateway registered for that chain
getGateway(chain);

```

contracts/crosschain/ERC7802Bridge.sol
ernestognw: Yeah I see what you mean. The main separation of concerns we should keep is data vs value. 7802 token bridges should indeed be built on top of 7786 protocols

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.27;
```
ernestognw, line 3:
```suggestion
pragma solidity ^0.8.26;
```

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
...
+
+    error ERC7802BridgeMissingGateway(bytes chain);
+    error ERC7802BridgeMissingRemote(address token, bytes chain);
+    error ERC7802BridgeImproperChainIdentifier(bytes chain);
+    error ERC7802BridgeDuplicate();
+    error ERC7802BridgeInvalidGateway();
+    error ERC7802BridgeInvalidSender();
+
+    mapping(address token => mapping(bytes chain => bytes bridge)) private _remoteBridges;
+    mapping(address token => mapping(bytes chain => bytes remote)) private _remoteTokens;
```
ernestognw, line 32:
```suggestion
    mapping(address token => mapping(bytes chain => bytes)) private _remoteBridges;
    mapping(address token => mapping(bytes chain => bytes)) private _remoteTokens;
```

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
...
+        bridge = _remoteBridges[token][chain];
+        require(bridge.length > 0, ERC7802BridgeMissingRemote(token, chain));
+    }
+
+    function getRemoteToken(address token, bytes memory chain) public view virtual returns (bytes memory remote) {
+        remote = _remoteTokens[token][chain];
+        require(remote.length > 0, ERC7802BridgeMissingRemote(token, chain));
+    }
+
+    // TODO: check that gateway is not address(0) ? prevent  override ?
```
ernestognw, line 51: If we're checking that gateway is not `address(0)` here I don't think we should check it on the `getGateway`. I'm generally unsure whether to check `address(0)` in the first place.

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
...
+            (bytes, bytes, bytes, uint256)
+        );
+
+        // identify source chain and validate corresponding gateway
+        (bytes2 chainType, bytes memory chainReference, ) = from.parseV1();
+        bytes memory srcChain = InteroperableAddress.formatV1(chainType, chainReference, "");
+        require(msg.sender == getGateway(srcChain), ERC7802BridgeInvalidGateway());
+
+        // identify local token (requested for mint) and validate sender is the correct bridge
+        (, address token) = local.parseEvmV1(); // todo: check chainid ?
```
ernestognw, line 131: I think we should indeed check `chainId`. One can register a bridge that may have the same address in another chain (perhaps different code), so `getRemoteBridge` will return true.

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
...
+
+        return IERC7786Receiver.executeMessage.selector;
+    }
+
+    function _fetchTokens(address token, address from, uint256 amount) internal virtual {
+        IERC7802(token).crosschainBurn(from, amount);
+    }
+
+    function _distributeTokens(address token, address to, uint256 amount) internal virtual {
+        IERC7802(token).crosschainMint(to, amount);
```
ernestognw, line 149: Shall we validate that `to` is not the gateway itself? Otherwise we should add a sweep function (not sure if I like the sweep function though)

contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,161 @@
...
+    }
+}
+
+abstract contract ERC7802BridgeCustody is ERC7802Bridge {
```
ernestognw, line 153: I don't think this version is necessary if the main bridge can receive tokens itself. Otherwise, I would call this one an `ERC7802Passive` since the main difference is whether tokens are **actively sent** to a recipient or **passively claimed** by the recipient

PR #182, opened 2025-06-24
Add ZKJWTUtils

ernestognw:
Similar to https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/96 but for ZK JWT


contracts/utils/cryptography/ZKJWTUtils.sol
```diff
@@ -0,0 +1,147 @@
...
+        } else if (!_commandMatch(jwtProof, template, templateParams, stringCase)) {
+            return JWTProofError.MismatchedCommand;
+        } else if (!jwtRegistry.isDKIMPublicKeyHashValid(jwtProof.domainName, jwtProof.publicKeyHash)) {
+            // Validate JWT public key and authorized party through registry
+            // The domainName contains "kid|iss|azp" format for JWT validation
+            return JWTProofError.JWTPublicKeyHash;
+        }
+
+        // Verify the zero-knowledge proof of JWT signature
+        // TODO: Is `verifyEmailProof` supposed to be non-view?
```
ernestognw, line 115:
```suggestion
```

ernestognw:
It seems ready imo. I couldn't get the `testInvalidJWTProof` to pass. It'll always revert with fuzzing. It seems that this comes from differences between the [JwtGroth16Verifier](https://github.com/zkemail/zk-jwt/blob/6cfd29c8155fd173bdce7c201860672b996b2ba9/packages/contracts/src/utils/JwtGroth16Verifier.sol) used under the hood vs the [Groth16Verifier](https://github.com/zkemail/email-tx-builder/blob/c97004116b8ae8fac6e99d5a1f819784ed49f6dd/packages/contracts/src/utils/Groth16Verifier.sol) used for ZKEmail.

I think we should discuss whether how to remove the `try/catch` block in `isValidZKJWT` or also add it to `isValidZKEmail`

ernestognw:
I think we should discuss https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/210 first. I'd like to port those changes to ZKJWTUtils as well but first we need to validate that dropping `templateId` is actually fine

PR #186, opened 2025-07-10
Add ERC20Freezable and ERC20uRWA

ernestognw:


contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @dev Returns whether a user is allowed to interact with the token.
+     *
+     * Default implementation only disallows explicitly RESTRICTED users (i.e. a blocklist).
+     *
+     * To convert into an allowlist, override as:
+     *
+     * ```solidity
+     * function isUserAllowed(address user) public view virtual override returns (bool) {
+     *     return getRestriction(user) != Restriction.UNRESTRICTED; // i.e. DEFAULT && UNRESTRICTED
```
ernestognw, line 44:
```suggestion
     *     return getRestriction(user) != Restriction.UNRESTRICTED; // i.e. DEFAULT && RESTRICTED
```

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * @dev Returns whether a user is allowed to interact with the token.
+     *
+     * Default implementation only disallows explicitly RESTRICTED users (i.e. a blocklist).
+     *
+     * To convert into an allowlist, override as:
+     *
+     * ```solidity
+     * function isUserAllowed(address user) public view virtual override returns (bool) {
+     *     return getRestriction(user) != Restriction.UNRESTRICTED; // i.e. DEFAULT && RESTRICTED
```
ernestognw, line 44:
```suggestion
     *     return getRestriction(user) == Restriction.UNRESTRICTED;
```

contracts/token/ERC20/extensions/ERC20Freezable.sol
```diff
@@ -0,0 +1,60 @@
...
+ * (e.g. a DAO or a well-configured multisig) that a specific amount
+ * of tokens held by an account won't be transferable until those
+ * tokens are unfrozen using {_unfreezeTokens}.
+ */
+abstract contract ERC20Freezable is ERC20 {
+    /// @dev Frozen amount of tokens per address.
+    mapping(address account => uint256) private _frozenBalances;
+
+    /// @dev The operation failed because the user has insufficient unfrozen balance.
+    error ERC20InsufficientUnfrozenBalance(address user, uint256 needed, uint256 available);
```
xaler5, line 24: What about `ERC7943InsufficientUnfrozenBalance` ? Is not mandatory but just curious of your thoughts

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,104 @@
...
+
+    /// @inheritdoc IERC7943
+    function getFrozen(address user, uint256) public view virtual returns (uint256 amount) {
+        return frozen(user);
+    }
+
+    /// @inheritdoc IERC7943
+    function setFrozen(address user, uint256, uint256 amount) public virtual {
+        require(amount <= balanceOf(user), ERC20InsufficientBalance(user, balanceOf(user), amount));
+        _checkFreezer(user, amount);
```
james-toussaint, line 41:
```suggestion
        _checkFreezer(user, amount);
        require(amount <= balanceOf(user), ERC20InsufficientBalance(user, balanceOf(user), amount));
```
Maybe more a style preference, but checking ACL looks to me the first thing to do, like a modifier would do.
https://eips.ethereum.org/EIPS/eip-7943#security-considerations
(Same applies inside `forceTransfer(...)`)

contracts/mocks/token/ERC20uRWAMock.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.26;
+
+import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {uRWA20} from "../../token/ERC20/extensions/ERC20uRWA.sol";
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+
+// solhint-disable-next-line contract-name-capwords
+abstract contract uRWA20Mock is uRWA20, AccessControl {
```
james-toussaint, line 9:
```suggestion
contract uRWA20Mock is uRWA20, AccessControl {
```
?
Do we need the mock to be abstract?

ernestognw: It's not needed but it helps because hardhat-exposed exposes the constructors automatically 😄

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+    /// @dev Updates the restriction of a user.
+    function _setRestriction(address user, Restriction restriction) internal virtual {
+        if (getRestriction(user) != restriction) {
+            _restrictions[user] = restriction;
+            emit UserRestrictionsUpdated(user, restriction);
+        } // no-op if restriction is unchanged
+    }
+
+    /// @dev Convenience function to restrict a user (set to RESTRICTED).
+    function _allowUser(address user) internal virtual {
```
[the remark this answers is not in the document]
ernestognw: I wanted to make it consistent with `isUserAllowed` but these still make sense. It makes sense anyway but no strong opinion and I prefer the consistency with `isUserAllowed`. It's also true that "allowed" overlaps with the "allowance" concept in tokens

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+     * To convert into an allowlist, override as:
+     *
+     * ```solidity
+     * function isUserAllowed(address user) public view virtual override returns (bool) {
+     *     return getRestriction(user) == Restriction.UNRESTRICTED;
+     * }
+     * ```
+     */
+    function isUserAllowed(address user) public view virtual returns (bool) {
+        return getRestriction(user) != Restriction.RESTRICTED; // i.e. DEFAULT && UNRESTRICTED
```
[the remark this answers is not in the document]
ernestognw: Yeah, good idea. Though, these functions come in ERC20Allowlist and ERC20Blocklist (i.e. `allowed` and `blocked` respectively). Initially ERC20Restricted was not a thing and this was the default recommendation (see [here](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/186/commits/2072d11ee5087ee9f883a6c2fdaaf94984415ae7#diff-48acb3ef31e4d54fb8da7f91b87a9d508b8fa23e7fd81ade81ea74e99fc8824aR21)):

```solidity
// Using ERC20Allowlist
contract MyuRWA20 is uRWA20, ERC20Allowlist {
    function isUserAllowed(address user) public view virtual override returns (bool) {
        return allowed(user);
    }
}

// Using ERC20Blocklist
contract MyuRWA20 is uRWA20, ERC20Blocklist {
    function isUserAllowed(address user) public view virtual override returns (bool) {
        return !blocked(user);
    }
}
```

I currently prefer the ERC20Freezable with a default "blocklist" behavior, since I believe is the least restrictive option (i.e. an allowlist is just a stricter blocklist)

contracts/token/ERC20/extensions/ERC20Restricted.sol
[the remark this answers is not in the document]
ernestognw: Updated 😄

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * @dev Extension of {ERC20} that allows to implement user access restrictions
+ * through the {isUserAllowed} function. Inspired by https://eips.ethereum.org/EIPS/eip-7943[EIP-7943].
+ *
+ * By default, each user has no explicit restriction. The {isUserAllowed} function acts as
+ * an blocklist. Developers can override {isUserAllowed} to check that `restriction == UNRESTRICTED`
+ * to implement an allowlist.
+ */
+abstract contract ERC20Restricted is ERC20 {
+    enum Restriction {
```
arr00, line 16: I feel like something like status or role would fit better here.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,104 @@
...
+import {ERC20Restricted} from "./ERC20Restricted.sol";
+
+/**
+ * @dev Extension of {ERC20} according to https://eips.ethereum.org/EIPS/eip-7943[EIP-7943].
+ *
+ * Combines standard ERC-20 functionality with RWA-specific features like user restrictions,
+ * asset freezing, and forced asset transfers.
+ */
+// solhint-disable-next-line contract-name-capwords
+abstract contract uRWA20 is ERC20, ERC165, ERC20Freezable, ERC20Restricted, IERC7943 {
```
arr00, line 17: `ERC20uRWA` or `uEWA20`?

contracts/token/ERC20/extensions/ERC20Freezable.sol
```diff
@@ -0,0 +1,60 @@
...
+    function _update(address from, address to, uint256 value) internal virtual override {
+        if (from != address(0)) {
+            uint256 unfrozen = available(from);
+            require(unfrozen >= value, ERC20InsufficientUnfrozenBalance(from, value, unfrozen));
+        }
+        super._update(from, to, value);
+    }
+
+    // We don't check frozen balance for approvals since the actual transfer
+    // will be checked in _update. This allows for more flexible approval patterns.
```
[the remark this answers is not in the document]
ernestognw: For clarity, it is not possible to bypass the restriction since every `transferFrom` would go through `_update`

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+     *
+     * To convert into an allowlist, override as:
+     *
+     * ```solidity
+     * function isUserAllowed(address user) public view virtual override returns (bool) {
+     *     return getRestriction(user) == Restriction.UNRESTRICTED;
+     * }
+     * ```
+     */
+    function isUserAllowed(address user) public view virtual returns (bool) {
```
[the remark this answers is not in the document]
ernestognw: Ah, that comment is related to how the ERC20uRWA was set up in the first place. See [this comment](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/186#discussion_r2223036122).

I agree that a default is opinionated, but I think we don want to be opinionated here to avoid users to make a decision at first. I prefer a non-restrictive default rather than a virtual function

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,103 @@
...
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override(ERC20, ERC20Freezable, ERC20Restricted) {
+        if (from == address(0)) {
+            // Minting
+            require(isUserAllowed(to), ERC7943NotAllowedUser(to));
+        } else if (to != address(0)) {
+            // Transfer
+            require(isTransferAllowed(from, to, 0, amount), ERC7943NotAllowedTransfer(from, to, 0, amount));
```
ernestognw, line 77: I think I'll remove this check. Essentially, it duplicates the check that `super._update` already does given it inherits from ERC20Freezable.

In terms of compliance, the spec states:

> Public transfers (transfer, transferFrom, safeTransferFrom, etc.) MUST NOT succeed in cases in which isTransferAllowed or isUserAllowed would return false for either one or both from and to addresses.

So this is still true even if we don't call `isTransferAllowed` here.

> The isTransferAllowed MUST validate that the amount being transferred doesn't exceed the unfrozen amount (which is the difference between the current balance and the frozen balance). Additionally it MUST perform an isUserAllowed check on the from and to parameters.

1. `ERC20Freezable._update` checks:
    - `if (from != address(0))` → validates `amount <= available(from)` ✅

2. `ERC20Restricted._update` checks:
    - `if (from != address(0)) _checkRestricted(from)` → calls `isUserAllowed(from)` ✅ 
    - `if (to != address(0)) _checkRestricted(to)` → calls `isUserAllowed(to)` ✅

This change would stop emitting the ERC7943NotAllowedTransfer error, but I think that's fine since it's not very expressive and the `super._update` call will revert with more appropriate errors. The ERC allows to skip reverting with ERC7943NotAllowedTransfer.

As a side effect, I think any override to `isTransferAllowed` would not be enforced in `_update` automatically, so I guess we can note it.

contracts/token/ERC20/extensions/ERC20Freezable.sol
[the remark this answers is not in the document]
arr00: agree with not checking

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -0,0 +1,96 @@
...
+        _setRestriction(account, Restriction.ALLOWED);
+    }
+
+    /// @dev Convenience function to reset a user account to default restriction.
+    function _resetUser(address account) internal virtual {
+        _setRestriction(account, Restriction.DEFAULT);
+    }
+
+    /// @dev Checks if a user account is restricted. Reverts with {ERC20Restricted} if so.
+    function _checkRestricted(address account) internal view virtual {
```
arr00, line 93:
```suggestion
    function _checkRestriction(address account) internal view virtual {
```
Seems more consistent.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,109 @@
...
+
+    /// @inheritdoc IERC7943
+    function getFrozen(address user, uint256) public view virtual returns (uint256 amount) {
+        return frozen(user);
+    }
+
+    /// @inheritdoc IERC7943
+    function setFrozen(address user, uint256, uint256 amount) public virtual {
+        _checkFreezer(user, amount);
+        require(amount <= balanceOf(user), ERC20InsufficientBalance(user, balanceOf(user), amount));
```
arr00, line 45: Is it actually an issue to allow setting more than the balance of a user to frozen? Couldn't the user monitor the mempool and ddos freeze attempts?

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,109 @@
...
+        _checkEnforcer(from, to, amount);
+        require(isUserAllowed(to), ERC7943NotAllowedUser(to));
+
+        // Update frozen balance if needed
+        uint256 currentFrozen = frozen(from);
+        uint256 currentBalance = balanceOf(from);
+        uint256 newAmount = currentBalance - amount;
+        if (currentFrozen > newAmount) {
+            _setFrozen(from, newAmount);
+        }
```
arr00, line 60: Why is this necessary, given we are bypassing the available check?

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,109 @@
...
+        address to,
+        uint256 amount
+    ) internal virtual override(ERC20, ERC20Freezable, ERC20Restricted) {
+        if (from == address(0)) {
+            // Minting
+            require(isUserAllowed(to), ERC7943NotAllowedUser(to));
+        }
+        // Note that `isTransferAllowed` duplicates the `available` check made by `super._update` in ERC20Freezable,
+        // so, the following line is not needed but left for reference:
+        // require(isTransferAllowed(from, to, 0, amount), ERC7943NotAllowedTransfer(from, to, 0, amount));
```
arr00, line 84: given that its just an external utility, I think its a bit misleading to bring up `isTransferAllowed` here

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,109 @@
...
+     */
+    function _update(
+        address from,
+        address to,
+        uint256 amount
+    ) internal virtual override(ERC20, ERC20Freezable, ERC20Restricted) {
+        if (from == address(0)) {
+            // Minting
+            require(isUserAllowed(to), ERC7943NotAllowedUser(to));
+        }
```
arr00, line 81: This is just to emit a different event here?

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,109 @@
...
+     * Example usage with {AccessControl-onlyRole}:
+     *
+     * ```solidity
+     * function _checkEnforcer(address from, address to, uint256 amount) internal view override onlyRole(ENFORCER_ROLE) {}
+     * ```
+     */
+    function _checkEnforcer(address from, address to, uint256 amount) internal view virtual;
+
+    /**
+     * @dev Internal function to check if the user has sufficient unfrozen balance.
```
arr00, line 100: Isn't this an access control function?

ernestognw: Yes! Updated

contracts/token/ERC20/extensions/ERC20uRWA.sol
ernestognw: Yeah makes sense. I'll cap it to `Math.min(amount, balanceOf(user))`. I think it's better to follow the "no op" policy here

contracts/token/ERC20/extensions/ERC20uRWA.sol
ernestognw: Right. The `if` check is a gas optimization.

1) No if (`setFrozen(Math.min(currentFrozen, newBalance))`): Always executes SSTORE (warm) + emits Frozen event (100 gas + ~1000 gas)
2) With if: Just a JUMPI when no update needed (~10 gas)

arr00: I'm saying we never really need to call `_setFrozen`

contracts/token/ERC20/extensions/ERC20uRWA.sol
arr00: The specific error is not required by the standard specifically (just to revert, which would happen downstream). Using the ERC error is fine by me though.

contracts/token/ERC20/extensions/ERC20uRWA.sol
arr00: going further here, what if the freezer intends to freeze more assets than is currently in the wallet (maybe they have an incoming transfer or possible future tokens). Is there an issue with this usecase?

contracts/token/ERC20/extensions/ERC20uRWA.sol
ernestognw: Ah got it! Great catch. I think if the error is not mandated it's better to save up the `if` condition.

ernestognw: Removed it in [f5b2008](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/186/commits/f5b2008116a9b75db5e57baa8e3f70bdce5c71fc)

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    /**
+     * @dev See {IERC7943-setFrozen}.
+     *
+     * NOTE: The `amount` is capped to the balance of the `user` to ensure the {IERC7943-Frozen} event
+     * emits values that consistently reflect the actual amount of tokens that are frozen.
+     */
+    function setFrozen(address user, uint256, uint256 amount) public virtual {
+        uint256 actualAmount = Math.min(amount, balanceOf(user));
```
arr00, line 50: https://github.com/ethereum/ERCs/blob/9f76a19660ffec0e57df3ed2d660ec1d03993f6c/ERCS/erc-7943.md?plain=1#L147

The ERC does specify this, but I'm not sure I agree.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -0,0 +1,113 @@
...
+        uint256 newBalance;
+        unchecked {
+            // Safe because ERC20._update will check that balanceOf(from) >= amount
+            newBalance = balanceOf(from) - amount;
+        }
+        if (currentFrozen > newBalance) {
+            _setFrozen(from, newBalance);
+        }
+
+        ERC20._update(from, to, amount); // Explicit raw update to bypass all restrictions
```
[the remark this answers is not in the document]
ernestognw: Added a CAUTION note as we agreed in our 1:1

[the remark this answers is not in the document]
ernestognw: Actually I updated the `forceTransfer` function to remove the restrictions from storage momentarily before calling `_update`. I think it's more important to preserve side effects down the inheritance chain than saving gas in that function. Also, calling `forceTransfer` should be really occasional, so it's justifiable imo

arr00 (dismissed):

PR #191, opened 2025-07-16
Decode WebAuthn object from calldata buffers, with out-of-bound detection

Amxx:
Inspired by https://github.com/OpenZeppelin/openzeppelin-contracts/pull/5400
and https://github.com/Vectorized/solady/blob/main/src/utils/WebAuthn.sol#L180-L211


ernestognw (approved):
Small nit, but I like it overall. Thanks!

PR #194, opened 2025-07-20
Refactor WebAuthn

Amxx:
Observations

- Flags UP doesn't appear to be optional: enable it always
- Flags BE/BS are eaither not set (succes) or should be checks for inconsistency

Proposition:
- Create a variant of `verify` with a `requireUV` boolean
- Create a "default" variant of ` verify` that enables the UV check (equivalent to the old `verifyStrict`)
- Remove `verifyMinimal` with no way to reproduce it (UP check always on)
- Remove old `verify`, using the new `verify` (with bool) with false reproduces the "no UV check" part. BS-BE is still active
- Remove `verifyStrict`, the new `verify` (without boolean) has the same behavior.
- Use `verify` with UV in the signer/verifier. This is eqauivalent to using verifyStrict / requireUV 
  - this follows the recommandations of the M-04 issue.

Side changes:
- Update the helper to set the UP and UV flags
- Refactor the fuzzing tests

### DIFF
--- a/contracts/utils/cryptography/WebAuthn.sol
+++ b/contracts/utils/cryptography/WebAuthn.sol
@@ -62,10 +62,11 @@ library WebAuthn {
 
     /**
      * @dev Performs the absolute minimal verification of a WebAuthn Authentication Assertion.
+     *
      * This function includes only the essential checks required for basic WebAuthn security:
      *
-     * 1. Type is "webauthn.get" (see {validateExpectedTypeHash})
-     * 2. Challenge matches the expected value (see {validateChallenge})
+     * 1. Type is "webauthn.get" (see {_validateExpectedTypeHash})
+     * 2. Challenge matches the expected value (see {_validateChallenge})
      * 3. Cryptographic signature is valid for the given public key
      *
      * For most applications, use {verify} or {verifyStrict} instead.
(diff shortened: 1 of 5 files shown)


contracts/utils/cryptography/WebAuthn.sol
```diff
@@ -104,69 +119,63 @@ library WebAuthn {
...
     /**
      * @dev Validates that the https://www.w3.org/TR/webauthn-2/#up[User Present (UP)] bit is set.
+     *
      * Step 16 in https://www.w3.org/TR/webauthn-2/#sctn-verifying-assertion[verifying an assertion].
      *
      * NOTE: Required by WebAuthn spec but may be skipped for platform authenticators
      * (Touch ID, Windows Hello) in controlled environments. Enforce for public-facing apps.
      */
-    function validateUserPresentBitSet(bytes1 flags) internal pure returns (bool) {
+    function _validateUserPresentBitSet(bytes1 flags) private pure returns (bool) {
```
ernestognw, line 172: Not opposed to make them private, but is there any strong reason?

test/utils/cryptography/WebAuthn.t.sol
```diff
@@ -4,348 +4,181 @@ pragma solidity ^0.8.20;
...
-        // Create authenticator data with minimum required length (37 bytes)
-        bytes memory authenticatorData = new bytes(37);
-
-        // Create client data JSON with invalid type
-        string memory clientDataJSON = string.concat(
-            // solhint-disable-next-line quotes
-            '{"type":"webauthn.create","challenge":"',
-            Base64.encodeURL(challenge),
-            SUFFIX
+    function testFuzzVerify(bytes memory challenge, uint256 seed) public view {
```
ernestognw, line 13: Let's drop the `Fuzz` part. Same for others

contracts/utils/cryptography/WebAuthn.sol
Amxx: My argument is to reduce the ABI of the library. I would avoid exposing functions that the users don't need so that they don't get confused by "should I use that?". If its private, it gives a clear signal that its internal helpers and that users should not call them (verify does that).

Which is all motivated by: I don't expect anyone to call that directly. 

If someone really wants, they would need to know for you extract the flags from the auth already so I'm not sure they need that helper

Amxx: Note, once we go to the vanila repo, we cannot move something from internal to private (for backward compatibility reasons). But we can move something from private to internal. So I'd have as much private as possible, and eventually make it internal if we ever got motivated requests

ernestognw: > I don't expect anyone to call that directly.

Yeah I'm realizing they only make sense if you want to do something like `verifyMinimal() && validateUserPresentBit()`

PR #195, opened 2025-07-21
Add an ERC-7802 bridge (v2)

Amxx:
Trustless alternative to #174


contracts/crosschain/ERC7802Bridge.sol
```diff
@@ -0,0 +1,256 @@
...
+
+        return bridgeId;
+    }
+
+    function setPaused(bytes32 bridgeId, bool isPaused) public bridgeAdminRestricted(bridgeId) {
+        _bridges[bridgeId].isPaused = isPaused;
+        emit BridgePaused(bridgeId, isPaused);
+    }
+
+    function updateGateway(
```
Amxx, line 133: currently, it is possible to add new chains to a bridgeId. This should not be allowed unless there is a mechanism to create the bridgeId on that new chain that is not part of the initial config.

PR #199, opened 2025-07-25
Replace submodule with npm dependency

Amxx:
Do we have any reason to favor submodules over npm packages ? IMO npm versionned packages are a better choice when available. I find them easier to track and update.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -12,10 +12,6 @@
 [submodule "lib/email-tx-builder"]
 	path = lib/email-tx-builder
 	url = https://github.com/zkemail/email-tx-builder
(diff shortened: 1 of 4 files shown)


ernestognw:
Packages were moved to submodules in [PR153](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/153), it was while you were on vacations. The reasoning is that other teams like ZKEmail using Foundry are already running `forge install`, but may not even have a `package.json`, so the less disruptive version for them is to keep dependencies as submodules.

I think that's reasonable, and we shouldn't force community-contracts users to run `npm install`

PR #200, opened 2025-07-25
Implement requestRelay in Axelar gateway

Amxx:


contracts/crosschain/axelar/AxelarGatewayBase.sol
```diff
@@ -41,8 +43,19 @@ abstract contract AxelarGatewayBase is Ownable {
...
+        _axelarGasService = _gasService;
+    }
+
+    // This is already exposed by AxelarExecutable which AxelarDestinationGateway inherit from. Because its not
+    // virtual, resolution is not possible. Therefore, we should not expose it.
+    // function gateway() public view virtual returns (IAxelarGateway) {
+    //     return _axelarGateway;
+    // }
+
+    function gasService() public view virtual returns (IAxelarGasService) {
```
frangio, line 57: Is this function required by Axelar?

contracts/crosschain/axelar/AxelarGatewayBase.sol
```diff
@@ -41,8 +43,19 @@ abstract contract AxelarGatewayBase is Ownable {
...
+    constructor(IAxelarGateway _gateway, IAxelarGasService _gasService) {
         _axelarGateway = _gateway;
+        _axelarGasService = _gasService;
+    }
+
+    // This is already exposed by AxelarExecutable which AxelarDestinationGateway inherit from. Because its not
+    // virtual, resolution is not possible. Therefore, we should not expose it.
+    // function gateway() public view virtual returns (IAxelarGateway) {
+    //     return _axelarGateway;
+    // }
```
frangio, line 55: Any reason not to remove this instead?

contracts/crosschain/utils/ERC7786Attributes.sol
```diff
@@ -0,0 +1,31 @@
...
+
+import {IERC7786Attributes} from "../../interfaces/IERC7786Attributes.sol";
+
+library ERC7786Attributes {
+    function tryDecodeRequestRelay(
+        bytes memory attribute
+    ) internal pure returns (bool success, uint256 value, uint256 gasLimit, address refundRecipient) {
+        success = bytes4(attribute) == IERC7786Attributes.requestRelay.selector && attribute.length >= 0x64;
+
+        assembly ("memory-safe") {
```
frangio, line 13: Isn't this technically unsafe as we're potentially reading out of bounds? Although we are multiplying by 0 in that case... so I think this is probably ok.

contracts/crosschain/utils/ERC7786Attributes.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.27;
+
+import {IERC7786Attributes} from "../../interfaces/IERC7786Attributes.sol";
+
+library ERC7786Attributes {
+    function tryDecodeRequestRelay(
```
frangio, line 8: I think we should include the attribute signature in documentation for this function.

contracts/crosschain/axelar/AxelarGatewaySource.sol
frangio, line 1: We should document somewhere that the `gasLimit` parameter is ignored, that instead `msg.value` determines the gas limit.

Does this behavior make sense?

contracts/crosschain/axelar/AxelarGatewaySource.sol
```diff
@@ -17,11 +19,20 @@ abstract contract AxelarGatewaySource is IERC7786GatewaySource, AxelarGatewayBas
     using InteroperableAddress for bytes;
     using Strings for address;
 
+    struct MessageDetails {
+        string destination;
+        string target;
+        bytes payload;
+    }
+
+    uint256 private _sendId;
```
frangio, line 28: Rename to `_nextSendId`?

contracts/crosschain/axelar/AxelarGatewaySource.sol
```diff
@@ -30,29 +41,58 @@ abstract contract AxelarGatewaySource is IERC7786GatewaySource, AxelarGatewayBas
...
         (bytes2 chainType, bytes calldata chainReference, ) = recipient.parseV1Calldata();
-        string memory axelarDestination = getAxelarChain(InteroperableAddress.formatV1(chainType, chainReference, ""));
         bytes memory remoteGateway = getRemoteGateway(chainType, chainReference);
-        _axelarGateway.callContract(
-            axelarDestination,
-            address(bytes20(remoteGateway)).toChecksumHexString(), // TODO non-evm chains?
-            adapterPayload
-        );
+        string memory axelarDestination = getAxelarChain(InteroperableAddress.formatV1(chainType, chainReference, ""));
+        string memory axelarTarget = address(bytes20(remoteGateway)).toChecksumHexString(); // TODO non-evm chains?
```
frangio, line 69: What do we need to resolve this TODO?

contracts/crosschain/axelar/AxelarGatewaySource.sol
```diff
@@ -30,29 +41,58 @@ abstract contract AxelarGatewaySource is IERC7786GatewaySource, AxelarGatewayBas
...
+                axelarTarget,
+                adapterPayload,
+                refundRecipient
+            );
+        } else {
+            _details[sendId] = MessageDetails(axelarDestination, axelarTarget, adapterPayload);
+        }
+    }
+
+    function requestRelay(bytes32 sendId, uint256 /*gasLimit*/, address refundRecipient) external payable {
```
frangio, line 86: Should we delete details from storage at the end of this function?

contracts/crosschain/axelar/AxelarGatewaySource.sol
```diff
@@ -30,29 +41,58 @@ abstract contract AxelarGatewaySource is IERC7786GatewaySource, AxelarGatewayBas
...
+        MessageDetails storage details = _details[sendId];
+        require(details.payload.length > 0);
 
-        return sendId;
+        _axelarGasService.payNativeGasForContractCall{value: msg.value}(
```
frangio, line 90: Is it okay to call this function _after_ `callContract` has been invoked? I have a feeling that it needs to be called beforehand...

frangio: Source:

https://github.com/axelarnetwork/axelar-cgp-solidity/blob/d030522e97463061a0c5f4c17bfa50c11cf23b59/contracts/gas-service/AxelarGasService.sol#L156

contracts/crosschain/axelar/AxelarGatewayBase.sol
Amxx: No, I thought of that as "discovery". Just like `gateway()` is not required.

Amxx: should we not have it ?

contracts/crosschain/utils/ERC7786Attributes.sol
Amxx: Yes, multiply by zero is here for security. Any "unsafe" access should be nullified. If the content being read is irrelevant, I don't see how that could be a problem.

contracts/crosschain/axelar/AxelarGatewaySource.sol
Amxx: Good observation, I'll check that documentation

Amxx: Note: that means the implementation of `sendMessage` -> `requestRelay` should not work. @cjcobb23 what is the interface to pay gas AFTER a message is sent ?

contracts/interfaces/IERC7786Attributes.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.4;
+
+/**
+ * @dev Standard attributes for ERC-7786. These attributes may be standardized in different ERC.
```
Amxx, line 6:
```suggestion
 * @dev Standard attributes for ERC-7786. These attributes may be standardized in different ERCs.
```

contracts/crosschain/utils/ERC7786Attributes.sol
frangio: It would definitely be a problem with traditional undefined behavior, but I don't think that's what happens with unsafe memory accesses in solc...

contracts/crosschain/axelar/AxelarGatewaySource.sol
frangio: As far as I know it's the `addGas` functions and those take tx hash and log index!

It really seems like there's no way to implement `requestRelay` for this adapter other than delaying `callContract` until `requestRelay` is called...

PR #210, opened 2025-08-10
Remove `templateId` and simplify EmailProof validation in ZKEmailUtils

ernestognw:
This PR streamlines the ZKEmailUtils by receiving an EmailProof directly instead of an EmailAuthMsg struct. From the original struct, the `templateId` was unused and not included in the proof so anyone can construct the expected value, so it was removed.

On the `SignerZKEmail` and `ERC7913ZKEmailVerifier` side, their implementation was simplified following changes in ZKEmailUtils.

**Key Changes:**
- Remove `templateId` and `skippedCommandPrefix` functionality across all ZKEmail contracts
- Add robust `tryDecodeEmailProof` function for safe calldata decoding with comprehensive bounds checking
- Update signature encoding in tests to use direct `EmailProof` field encoding
- Maintain security through existing DKIM verification and zero-knowledge proof validation

The simplified implementation reduces complexity while preserving all essential security guarantees.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## 10-08-2025
+## 14-08-2025
 
 - `ZKEmailUtils`: Add `tryDecodeEmailProof` function for safe calldata decoding with comprehensive bounds checking and validation for `EmailProof` struct.
 - `ZKEmailUtils`: Update `isValidZKEmail` to receive `EmailProof` struct directly instead of `EmailAuthMsg` struct.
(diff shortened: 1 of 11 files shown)


zkfriendly (dismissed):
Thanks for the simplification. Removing `templateId` and `skippedCommandPrefix` makes sense to me here. Other changes also look good to me.

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 10-08-2025
```
ernestognw, line 1:
```suggestion
## 14-08-2025
```

PR #228, opened 2026-04-09
Add abstract `ERC7540` and specific implementations

ernestognw:
## Release Notes

* **New Features**
  - Implemented full ERC-7540 standard support for asynchronous deposit and redemption request management
  - Added operator authorization system for vault operation controls
  - Enabled admin-fulfilled deposit request processing with asset claim functionality
  - Enabled admin-fulfilled redemption request processing with asset claim functionality
  - Introduced time-delayed deposit and redemption mechanisms with configurable processing delays


contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,536 @@
...
+        revert NotImplemented();
+    }
+
+    function _asyncMaxWithdraw(address /*owner*/) internal view virtual returns (uint256) {
+        revert NotImplemented();
+    }
+
+    function _asyncMaxRedeem(address /*owner*/) internal view virtual returns (uint256) {
+        revert NotImplemented();
+    }
```
ernestognw: Would it be better to leave all these functions as unimplemented virtuals rather than reverting?

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,536 @@
...
+            address(asset_),
+            abi.encodeCall(IERC20Metadata.decimals, ())
+        );
+        Memory.unsafeSetFreeMemoryPointer(ptr);
+
+        return
+            (success && LowLevelCall.returnDataSize() >= 32 && uint256(returnedDecimals) <= type(uint8).max)
+                ? (true, uint8(uint256(returnedDecimals)))
+                : (false, 0);
+    }
```
ernestognw, line 133: I believe we should add to vanilla's SafeERC20 the following function:

```solidity
function tryGetDecimals(address token) internal view returns (bool success, uint8 decimals) {
    bytes4 selector = IERC20Metadata.decimals.selector;
    assembly ("memory-safe") {
        mstore(0x00, selector)
        success := staticcall(gas(), token, 0x00, 4, 0x00, 0x20)
        success := and(and(success, gt(returndatasize(), 0x1f)), lt(mload(0x00), 0x100))
        decimals := mul(success, mload(0x00))
    }
}
```

This way we can reuse it here and in ERC4626

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
```diff
@@ -0,0 +1,83 @@
...
+        _deposits[controller].claimableAssets += assets;
+        _deposits[controller].claimableShares += shares;
+
+        emit DepositClaimable(controller, 0, assets, shares);
+    }
+
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = Math.mulDiv(assets, maxMint(controller), maxDeposit(controller), Math.Rounding.Floor);
+        _deposits[controller].claimableAssets = Math.saturatingSub(_deposits[controller].claimableAssets, assets);
+        _deposits[controller].claimableShares = Math.saturatingSub(_deposits[controller].claimableShares, shares);
```
ernestognw, line 90:
```suggestion
        _deposits[controller].claimableAssets = Math.saturatingSub(_asyncMaxDeposit(controller), assets);
        _deposits[controller].claimableShares = Math.saturatingSub(_asyncMaxMint(controller), shares);
```

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
```diff
@@ -0,0 +1,83 @@
...
+        return shares;
+    }
+
+    function _consumeClaimableMint(uint256 shares, address controller) internal virtual override returns (uint256) {
+        uint256 assets = Math.mulDiv(shares, maxDeposit(controller), maxMint(controller), Math.Rounding.Ceil);
+        _deposits[controller].claimableAssets = Math.saturatingSub(_deposits[controller].claimableAssets, assets);
+        _deposits[controller].claimableShares = Math.saturatingSub(_deposits[controller].claimableShares, shares);
```
ernestognw, line 98:
```suggestion
        _deposits[controller].claimableAssets = Math.saturatingSub(_asyncMaxDeposit(controller), assets);
        _deposits[controller].claimableShares = Math.saturatingSub(_asyncMaxMint(controller), shares);
```

contracts/token/ERC20/extensions/ERC7540AdminRedeem.sol
```diff
@@ -0,0 +1,82 @@
...
+        _redeems[controller].claimableShares += shares;
+        _redeems[controller].claimableAssets += assets;
+
+        emit RedeemClaimable(controller, 0, assets, shares);
+    }
+
+    function _consumeClaimableWithdraw(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = Math.mulDiv(assets, maxRedeem(controller), maxWithdraw(controller), Math.Rounding.Ceil);
+        _redeems[controller].claimableAssets = Math.saturatingSub(_redeems[controller].claimableAssets, assets);
+        _redeems[controller].claimableShares = Math.saturatingSub(_redeems[controller].claimableShares, shares);
```
ernestognw, line 91:
```suggestion
        _redeems[controller].claimableAssets = Math.saturatingSub(_asyncMaxWithdraw(controller), assets);
        _redeems[controller].claimableShares = Math.saturatingSub(_asyncMaxRedeem(controller), shares);
```

contracts/token/ERC20/extensions/ERC7540AdminRedeem.sol
```diff
@@ -0,0 +1,82 @@
...
+        return shares;
+    }
+
+    function _consumeClaimableRedeem(uint256 shares, address controller) internal virtual override returns (uint256) {
+        uint256 assets = Math.mulDiv(shares, maxWithdraw(controller), maxRedeem(controller), Math.Rounding.Floor);
+        _redeems[controller].claimableAssets = Math.saturatingSub(_redeems[controller].claimableAssets, assets);
+        _redeems[controller].claimableShares = Math.saturatingSub(_redeems[controller].claimableShares, shares);
```
ernestognw, line 99:
```suggestion
        _redeems[controller].claimableAssets = Math.saturatingSub(_asyncMaxWithdraw(controller), assets);
        _redeems[controller].claimableShares = Math.saturatingSub(_asyncMaxRedeem(controller), shares);
```

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,87 @@
...
+import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import {Checkpoints} from "@openzeppelin/contracts/utils/structs/Checkpoints.sol";
+import {ERC7540} from "./ERC7540.sol";
+
+abstract contract ERC7540DelayDeposit is ERC7540 {
+    using SafeCast for uint256;
+    using Checkpoints for Checkpoints.Trace208;
+
+    mapping(address controller => Checkpoints.Trace208 trace) private _deposits;
```
ernestognw:
```suggestion
    mapping(address controller => Checkpoints.Trace208) private _deposits;
```

ernestognw: Perhaps we should have a getter for this internal variable (i.e. `function deposits(address controller) ... returns (Checkpoints.Trace208)`

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,87 @@
...
+    mapping(address controller => uint256) private _claimedDeposits;
```
ernestognw, line 37: Same comment about a getter

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    function clock() public view virtual returns (uint48) {
+        return uint48(block.timestamp);
+    }
+
+    function depositDelay(address /*controller*/) public view virtual returns (uint48) {
+        return 1 hours;
```
ernestognw, line 59: Is this value arbitrary or based on current implementations?

contracts/token/ERC20/extensions/ERC7540DelayRedeem.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    mapping(address controller => Checkpoints.Trace208) private _redeems;
+    mapping(address controller => uint256) private _claimedRedeems;
+
+    function clock() public view virtual returns (uint48) {
+        return uint48(block.timestamp);
+    }
+
+    function redeemDelay(address /*controller*/) public view virtual returns (uint48) {
+        return 1 hours;
```
ernestognw, line 59: Same as in the deposit version

contracts/token/ERC20/extensions/ERC7540DelayRedeem.sol
```diff
@@ -0,0 +1,86 @@
...
+import {SafeCast} from "@openzeppelin/contracts/utils/math/SafeCast.sol";
+import {Checkpoints} from "@openzeppelin/contracts/utils/structs/Checkpoints.sol";
+import {ERC7540} from "./ERC7540.sol";
+
+abstract contract ERC7540DelayRedeem is ERC7540 {
+    using SafeCast for uint256;
+    using Checkpoints for Checkpoints.Trace208;
+
+    mapping(address controller => Checkpoints.Trace208) private _redeems;
+    mapping(address controller => uint256) private _claimedRedeems;
```
ernestognw, line 37: Same comment about adding getters

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,164 @@
...
+    using DoubleEndedQueue for DoubleEndedQueue.Bytes32Deque;
+
+    struct EpochDepositMetadata {
+        uint256 totalAssets;
+        uint256 totalShares;
+        mapping(address account => uint256) requests;
+    }
+
+    mapping(uint256 epochId => EpochDepositMetadata) private _epochs;
+    mapping(address account => DoubleEndedQueue.Bytes32Deque) private _memberOf;
```
ernestognw: We should add getters for these 2 as well

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,164 @@
...
+        require(details.totalAssets > 0 && details.totalShares == 0); // TODO: invalid resolve
+
+        details.totalShares = totalShares;
+        // TODO: emit event
+    }
+
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = 0;
+
+        while (assets > 0) {
```
ernestognw: Perhaps by creating too many small requests, it's possible to DoS this function (i.e. make the epoch have too many loop iterations). I'm wondering if we should provide a way to defend against that

contracts/token/ERC20/extensions/ERC7540DelayRedeem.sol
```diff
@@ -0,0 +1,123 @@
...
+
+    /// @dev Returns the current clock value. Defaults to `block.timestamp`.
+    function clock() public view virtual returns (uint48) {
+        return uint48(block.timestamp);
+    }
```
ernestognw, line 55: We're missing `CLOCK_MODE` and inheriting from IERC6372

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,500 @@
...
+        }
+
+        uint256 shares = _isDepositAsync()
+            ? Math.mulDiv(assets, maxMint(controller), maxDeposit(controller), Math.Rounding.Floor)
+            : previewDeposit(assets);
+
+        _deposit(_msgSender(), receiver, assets, shares);
+
+        return shares;
+    }
```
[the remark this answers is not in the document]
ernestognw: This has been addressed in the current code. The inline `Math.mulDiv(assets, maxMint(controller), maxDeposit(controller), ...)` was replaced with a delegation to `_consumeClaimableDeposit(assets, controller)` (and `_consumeClaimableMint` for mint), which are virtual hooks implemented by each fulfillment strategy. The strategy extensions handle the conversion internally, so the base `deposit`/`mint` functions no longer perform division themselves and cannot hit a division-by-zero.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,500 @@
...
+        }
+
+        uint256 assets = _isDepositAsync()
+            ? Math.mulDiv(shares, maxDeposit(controller), maxMint(controller), Math.Rounding.Ceil)
+            : previewMint(shares);
+
+        _deposit(_msgSender(), receiver, assets, shares);
+
+        return assets;
+    }
```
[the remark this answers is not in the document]
ernestognw: Fixed in the current code. The `mint` three-argument overload now uses `_isDepositAsync() ? controller : receiver` for the `maxMint` check, matching `deposit`'s behavior. The inline `Math.mulDiv` was also replaced with `_consumeClaimableMint(shares, controller)`, so the controller is consistently used for all async state lookups.

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
```diff
@@ -0,0 +1,74 @@
...
+        _deposits[controller].claimableShares += shares;
+
+        emit DepositClaimable(controller, 0, assets, shares);
+    }
+
+    function _deposit(address caller, address receiver, uint256 assets, uint256 shares) internal virtual override {
+        _deposits[receiver].claimableAssets = Math.saturatingSub(_deposits[receiver].claimableAssets, assets);
+        _deposits[receiver].claimableShares = Math.saturatingSub(_deposits[receiver].claimableShares, shares);
+        super._deposit(caller, receiver, assets, shares);
+    }
```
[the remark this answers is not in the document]
ernestognw: Addressed by redesigning the Admin strategy. The current `ERC7540AdminDeposit` no longer overrides `_deposit`. Instead, the base `ERC7540.deposit(assets, receiver, controller)` calls `_consumeClaimableDeposit(assets, controller)`, which is overridden in `ERC7540AdminDeposit` to decrement `_deposits[controller]` — using the correct `controller` key throughout. The `receiver` is only used for the share transfer destination in the base `_deposit`, never for state tracking.

contracts/token/ERC20/extensions/ERC7540AdminRedeem.sol
```diff
@@ -0,0 +1,79 @@
...
+        address caller,
+        address receiver,
+        address owner,
+        uint256 assets,
+        uint256 shares
+    ) internal virtual override {
+        _redeems[receiver].claimableAssets = Math.saturatingSub(_redeems[receiver].claimableAssets, assets);
+        _redeems[receiver].claimableShares = Math.saturatingSub(_redeems[receiver].claimableShares, shares);
+        super._withdraw(caller, receiver, owner, assets, shares);
+    }
```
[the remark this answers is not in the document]
ernestognw: Same as the deposit side — addressed by the redesign. `ERC7540AdminRedeem` no longer overrides `_withdraw`. The base `ERC7540.redeem(shares, receiver, ownerOrController)` calls `_consumeClaimableRedeem(shares, ownerOrController)`, which is overridden to decrement `_redeems[controller]`. The `receiver` only controls where assets are transferred, not state tracking.

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,82 @@
...
+
+        _deposits[controller].push(timepoint, (assets + latest).toUint208());
+
+        return timepoint;
+    }
+
+    function _deposit(address caller, address receiver, uint256 assets, uint256 shares) internal virtual override {
+        _claimedDeposits[receiver] += assets;
+        super._deposit(caller, receiver, assets, shares);
+    }
```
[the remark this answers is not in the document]
ernestognw: Addressed in the current code. The `ERC7540DelayDeposit` no longer overrides `_deposit`. Instead, the base `ERC7540.deposit(assets, receiver, controller)` calls `_consumeClaimableDeposit(assets, controller)`, which is overridden in `ERC7540DelayDeposit` to increment `_claimedDeposits[controller]` — using the correct `controller` key. The `receiver` is only used for the share transfer destination, never for claim tracking.

contracts/token/ERC20/extensions/ERC7540DelayRedeem.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    function delay(address /*controller*/) internal view virtual returns (uint48) {
+        return 1 hours;
+    }
+
+    function _isDepositAsync() internal pure virtual override returns (bool) {
+        return true;
+    }
```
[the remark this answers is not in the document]
ernestognw: This was indeed a bug in the original commit but it's fixed in the current code. `ERC7540DelayRedeem` now correctly overrides `_isRedeemAsync()` (not `_isDepositAsync()`).

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: If we leave them as unimplemented, the devs will necessarily have to provide an implementation, even if they don't need one. For example, if you want to build a vault that is sync deposit & async redeem, and if there functions are not implemented, you'll have to provide an implementation for all the deposit & mint related function in this, even though you done need them.

An option could be to have `ERC7540SyncDeposit` and `ERC7540SyncRedeem` that you use to "fill the hole" when you want a sync behavior.

Amxx: Implemented in https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/228/commits/438bab68071c6d3690186dc064b33d888077aa87

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: Lets add an issue for that in the vanilla repo !

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
Amxx: LEts discuss getters in our one-on-one !

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
Amxx: arbitrary, easily overridable. If you feel like the default value should be different we can change it.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
Amxx: There is a limit to the number of request (which can be customized by overriding the `_requestQueueLimit` function). This should limit the number of iteration of this loop. Not that this contract is not tested at all, the design might have serious flaws.

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
Amxx: I think this is a common discussion that we've been facing many time:

When we have some state, a getter for that state, and an update function that modifies the state, should the modify function update what is storage, or what is returned by the getter ?

---

[In the case of ERC20](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/ERC20.sol#L176-L204), the balances are updated from storage, and not from `balanceOf` (which may be overriden). 

One of the reason we do things that way, is that if the override on the getter does some change like
```solidity
function balanceOf(address account) public virtaul override return (uint256) {
    return super.balanceOf(account) - _frozen[account];
}
```
then doing the update using the getter function as a reference would "burn" the frozen amount, and would do that each update.

AFAIK, the only "solution" would be to have an internal setter that could be overriden alongside the getter, but that would mean having an internal function that directly sets the storage, which is something that goes against our guidelines.

contracts/token/ERC20/extensions/ERC7540.sol
ernestognw: I went ahead and implemented here: https://github.com/OpenZeppelin/openzeppelin-contracts/pull/6482

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
ernestognw: Nah, 1 hour sounds like a good default to me, I was just curious. Solving.

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
ernestognw: Good point, this has indeed come up before. Here's where I land on the getter vs setter question:

**Getter (virtual read function):** I think providing one is worthwhile. Without it, a developer who wants to change how the value is computed must override every function that reads the private variable, often duplicating logic (e.g. copy, paste, modify). A virtual getter gives a single override point for consistent behavior across all read sites. Not providing one hurts the developer experience more than providing one.

**Setter (virtual write function):** I'm less convinced this is needed. In practice, a developer overriding the getter is almost certainly introducing their own storage variable. At that point, the original setter writes to a variable the getter no longer reads — it becomes dead code. You *could* override the setter too to write the new variable, but that feels like unnecessary indirection when the new variable is already directly accessible in the subcontract.

That said, I acknowledge your ERC20 `balanceOf` example: if a getter override subtracts a frozen amount, writing back the getter's return value double-counts the subtraction on each update. However, I think in this case it shouldn't be an issue because we're using internal getters, and I'd prefer the consistency of having a single read override function

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
ernestognw: Correct, I didn't consider the limit function. We should add a `NOTE:` for this.

ernestognw: Explained in [37a10d7](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/228/commits/37a10d7f9660e53264ce46f4cf860467e854b55b)

ernestognw:
Just extracted the epoch based version to a separate PR: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/230

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
Amxx: In this particular case, the return cannot be in memory, it will have to be a storage pointer. That means we would providing direct access to the storage, defeating the purpose of having it private.

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
Amxx: I'm not seeing the difference between this and the ERC20 example. IMO, the internal getter is just "sugar syntax" for handling the 7540 sync/async mess, and its basically the internal equivalent of `maxDeposit`/`maxMint` ?

Are you saying that if someone wanted to replicate the `balanceOf` override that I described just above, they would do it on the public `maxDeposit` and not on the internal `_asyncMaxDeposit` ? I'd say both overrides are equally liquelly, and would result in very different behavior. Advanced users would liquelly spot that, but I'm not sure put them in a situation where they have to worry about that.

Do you have an concrete example where using the internal getter (instead of the sotrage slot) would improve the override experience ? I can't think of any.

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
Amxx: following our discussion, should be resolved by the addition of `_readyDepositAt`

contracts/token/ERC20/extensions/ERC7540DelayRedeem.sol
Amxx: following our discussion, should be resolved by the addition of _readyRedeemAt

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -769,6 +769,10 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
      *   deposit assets are tracked via {totalPendingDepositAssets} and decremented in {_deposit}.
      * * Non-zero address: shares are pre-minted to this address during fulfillment (via
      *   {_mintSharesOnDepositFulfill}) and transferred to the receiver on claim.
+     *
+     * NOTE: before using a non-zero address used here, it is essential to check that that address
+     * will not be able to transfer shares. We recommand using addresses such as address(1) or
```
ernestognw: I wouldn't recommend using `address(1)` since it risks being a precompile and there could be a way to transfer shares out.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,878 @@
...
+            _transferIn(caller, assets);
+            _mint(receiver, shares);
+        } else if (_depositShareOrigin() == address(0)) {
+            _totalPendingDepositAssets -= assets;
+            _mint(receiver, shares);
+        } else {
+            _transfer(_depositShareOrigin(), receiver, shares);
+        }
+
+        emit Deposit(caller, receiver, assets, shares);
```
ernestognw: The spec states the following:

> When the `Deposit` event is emitted, the first parameter MUST be the `controller`, and the second parameter MUST be the `receiver`.

Here we're emitting with the `caller`, which is `_msgSender()` in both calls to this internal `_deposit` function.

Maybe we should pass in the `controller` as an argument? @Amxx

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,878 @@
...
+    function deposit(
+        uint256 assets,
+        address receiver,
+        address controller
+    ) public virtual onlyOperatorOrController(_isDepositAsync(), controller, _msgSender()) returns (uint256) {
+        // Note: if _isDepositAsync is false, controller is ignored.
+        uint256 maxAssets = maxDeposit(_isDepositAsync() ? controller : receiver);
+        if (assets > maxAssets) {
+            revert ERC4626ExceededMaxDeposit(_isDepositAsync() ? controller : receiver, assets, maxAssets);
+        }
```
ernestognw, line 423: If `maxDeposit(...)` returns 0, then this check may pass if `assets` is 0 (i.e. `0 > 0` is false). Later on, if `_isDepositAsync` is true, the `_consumeClaimableDeposit` function will panic with division by zero.

I believe that's ok but should be added to the requirements. wdyt? @Amxx Alternatively we can change this check to be `if (assets > maxAssets || (_isDepositAsync() && maxAssets == 0)) { ... }` so it reverts with `ERC4626ExceededMaxDeposit`. I prefer the panic, though

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
```diff
@@ -0,0 +1,131 @@
...
+        if (_depositShareOrigin() != address(0)) {
+            _mintSharesOnDepositFulfill(assets, shares);
+        }
+
+        emit DepositClaimable(controller, 0, assets, shares);
+    }
+
+    /// @dev Consumes `assets` from the claimable deposit and returns the proportional shares (rounded down).
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = Math.mulDiv(assets, maxMint(controller), maxDeposit(controller), Math.Rounding.Floor);
```
ernestognw, line 104: If `maxMint(controller)` returns 0, this will panic. Should be documented in "Requirements" on this function's NatSpec

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,144 @@
...
+        uint256 latest = _deposits[controller].latest();
+
+        _deposits[controller].push(timepoint, (assets + latest).toUint208());
+
+        return super._requestDeposit(assets, controller, owner, timepoint);
+    }
+
+    /// @dev Consumes `assets` from claimable deposits, returns proportional shares (rounded down).
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = Math.mulDiv(assets, maxMint(controller), maxDeposit(controller), Math.Rounding.Floor);
```
ernestognw, line 93: Same comment about panic if `maxDeposit(controller)` is 0

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,878 @@
...
+
+        uint256 shares = _isDepositAsync() ? _consumeClaimableDeposit(assets, controller) : previewDeposit(assets);
```
ernestognw, line 425: On the async path, the exchange rate is computed when the shares are claimed, meaning that the deposit is susceptible to an inflation attack:

A frontrunner could:
1. Observe a pending deposit about to become claimable
2. Donate assets to the vault (inflating `totalAssets`)
3. The claimer receives fewer shares than expected
4. The attacker redeems their shares at the inflated rate

The `_decimalsOffset` mitigates this, which is especially important for the sync redeem case. It should be documented imo.

contracts/token/ERC20/extensions/ERC7540AdminDeposit.sol
```diff
@@ -0,0 +1,131 @@
...
+     * The fulfiller should ensure the vault's `totalAssets()` already reflects the deposited assets
+     * (e.g. after deploying them to a yield source) to avoid diluting existing holders.
+     *
+     * Emits a {DepositClaimable} event.
+     *
+     * Requirements:
+     *
+     * * `assets` must not exceed the pending deposit amount for the `controller`.
+     */
+    function _fulfillDeposit(uint256 assets, uint256 shares, address controller) internal virtual {
```
ernestognw, line 81: We should document that multiple partial fulfillments with different exchange rates will average. For example:

- `_fulfillDeposit(50, 100, controller)` → 2 shares/asset
- `_fulfillDeposit(50, 25, controller)` → 0.5 shares/asset
- claimableAssets=100, claimableShares=125

A partial claim of 50 assets gives `mulDiv(50, 125, 100) = 62` shares. Developers might expect FIFO ordering if not documented properly.

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -0,0 +1,878 @@
...
+        _asset = asset_;
+    }
+
+    /**
+     * @dev See {IERC165-supportsInterface}.
+     *
+     * Reports support for {IERC7540Operator} unconditionally. Support for {IERC7540Deposit} and
+     * {IERC7540Redeem} is conditional on the corresponding async selector returning `true`.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
```
ernestognw: Shall this report `IERC4626` too?

contracts/token/ERC20/extensions/ERC7540SyncDeposit.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: MIT
+
+pragma solidity ^0.8.27;
+
+import {ERC7540} from "./ERC7540.sol";
+
+/**
+ * @dev Module for enabling synchronous behavior (ERC-4626) for the deposit flow of an ERC-7540 vault.
+ *
+ * Note that an ERC-7540 vault is required to have at least on flow operating in asynchronous mode, so this module
```
ernestognw:
```suggestion
 * Note that an ERC-7540 vault is required to have at least one flow operating in asynchronous mode, so this module
```

contracts/token/ERC20/extensions/ERC7540SyncDeposit.sol
```diff
@@ -0,0 +1,60 @@
...
+ * cannot combined with {ERC7540SyncRedeem}.
```
ernestognw:
```suggestion
 * cannot be combined with {ERC7540SyncRedeem}.
```

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,144 @@
...
+     * A request is claimable once its timepoint has elapsed and the assets haven't been claimed yet.
+     */
+    function _claimableDepositRequest(
+        uint256 requestId,
+        address controller
+    ) internal view virtual override returns (uint256) {
+        unchecked {
+            uint48 timepoint = requestId.toUint48();
+            return
+                requestId > clock()
```
ernestognw: Maybe?

```suggestion
                timepoint > clock()
```

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
```diff
@@ -0,0 +1,144 @@
...
+                    ? 0
+                    : _readyDepositAt(controller, timepoint) - _readyDepositAt(controller, timepoint - 1);
```
ernestognw, line 135: The `requestId` is user input. If `timepoint == 0`, this becomes:

```solidity
_readyDepositAt(controller, 0) - _readyDepositAt(controller, type(uint48).max)
```

`_readyDepositAt(controller, type(uint48).max)` returns the cumulative of all deposits minus claimed, so the unchecked subtraction wraps to a huge value when there are unclaimed deposits. Same issue exists in `_claimableRedeemRequest`.

While `requestId = 0` is never produced by the Delay strategy, the spec says these functions "MUST NOT revert unless due to integer overflow caused by an unreasonably large input". In this case, we're silently returning a wrong value, which is arguably worse.

I think this validates what I was originally arguing about having more than one getter. The root cause is the refactoring in f144026 that unified `_pendingDepositRequest` and `_claimableDepositRequest` under the same `_readyDepositAt` internal getter. Before that commit, `_pendingDepositRequest` used **raw** checkpoint diffs (claim-unaware) while `_claimableDepositRequest` had its own claim-aware formula. Now they're the same expression, just swapping which branch is zero.

It works numerically because `_claimedDeposits[controller] <= cumulative[clock()]` is maintained by `maxDeposit`, so claims never spill into un-matured timepoints and the deduction cancels in the subtraction. But pending amounts *can't* have been claimed by definition, so `_pendingDepositRequest` shouldn't depend on claim state at all. The `requestId = 0` bug is a direct symptom: the invariant doesn't hold for `type(uint48).max`, the claims don't cancel, and the formula breaks.

This is exactly why I think the storage variables need two getters: one raw, one claim-aware:

```solidity
/// @dev Raw cumulative deposits at `timepoint` (claim-unaware).
function _cumulativeDepositAt(address owner, uint48 timepoint) internal view returns (uint256) {
    return _deposits[owner].upperLookupRecent(timepoint);
}

/// @dev Net claimable deposits at `timepoint` (claim-aware).
function _readyDepositAt(address owner, uint48 timepoint) internal view returns (uint256) {
    return Math.saturatingSub(_cumulativeDepositAt(owner, timepoint), _claimedDeposits[owner]);
}
```

Then `_pendingDepositRequest` uses `_cumulativeDepositAt` (raw diff, no claim dependency) and `_claimableDepositRequest` uses `_readyDepositAt` (claim-aware diff). This fixes the `requestId = 0` edge case on the pending side, makes the semantic distinction explicit, and removes the hidden dependency on the `claimed <= cumulative[clock()]` invariant for pending correctness. They represent genuinely different concepts, so a single getter conflating them was forcing a false unification.

Same applies to `ERC7540DelayRedeem`.

contracts/token/ERC20/extensions/ERC7540.sol
ernestognw: Added a note to `_consumeClaimableDeposit `

contracts/token/ERC20/extensions/ERC7540.sol
ernestognw: Added to requirements in the NatSpec

contracts/token/ERC20/extensions/ERC7540DelayDeposit.sol
Amxx: fixed that border condition in #e39484dd314905133828103c35ac945863d036a6

Note that we would have add it also with `_cumulativeDepositAt(ower, 0) - _cumulativeDepositAt(owner, type(uint256).max)`

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: Vanila's ERC4626 does not, so I wouldn't

(its not part of ERC-4626 specs)

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: I'm good with the panic + requirement for now.

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: If `_isDepositAsync`, then `caller` is not used. We could refactor `caller` into `callerOrController`

Then the calls on line 423 and 459 would become
```solidity
        _deposit(_isDepositAsync() ? controller, _msgSender(), receiver, assets, shares);
```

I honestly though I had done that already. I may have failled to push that, or maybe it was ovewritten by some commit. Anyway, this would be an option that doesn't introduce an extra variable.

docs/modules/ROOT/pages/erc7540.adoc
```diff
@@ -0,0 +1,477 @@
...
+| `withdraw(assets, receiver, owner)`
+| Burns shares, transfers assets out immediately
+| *Claims* a previously fulfilled redeem request
+
+| `redeem(shares, receiver, owner)`
+| Same as withdraw, but specifying shares
+| *Claims* by specifying shares
+|===
+
+ERC-7540 adds only two new entry points: `requestDeposit` and `requestRedeem`. Everything else builds on top of the existing ERC-4626 interface.
```
Amxx: 7540 also adds getters for observing the pending/claimable state of the requestId. Should we mention that here ?

docs/modules/ROOT/pages/erc7540.adoc
```diff
@@ -0,0 +1,477 @@
...
+   ┌─────────┐                 ┌─────────┐             ┌───────────┐                  ┌─────────┐
+   │  (none) │ ──────────────► │ Pending │ ──────────► │ Claimable │ ───────────────► │ Claimed │
+   └─────────┘                 └─────────┘             └───────────┘                  └─────────┘
+                                Assets/shares           Exchange rate                  Shares/assets
+                                locked in vault         is locked                      delivered to
+                                                                                       receiver
+....
+
+* **Pending**: Assets (for deposits) or shares (for redeems) are locked in the vault. The exchange rate is not yet known.
+* **Claimable**: The request has been fulfilled. The exchange rate is now locked. The controller can claim at any time.
```
Amxx: In the case of ERC7540Delay, the exchange rate is not frozen when the requestId becomes mature. Instead. assets just become ready, and the exchange rate in fixed when the claim happens. 

I don't think its a big deal, but maybe it should be mentionned ?

docs/modules/ROOT/pages/erc7540.adoc
```diff
@@ -0,0 +1,477 @@
...
+|===
+| Deposit Strategy | Redeem Strategy | Use Case
+
+| `ERC7540AdminDeposit`
+| `ERC7540AdminRedeem`
+| RWA vault with manual settlement on both sides
+
+| `ERC7540AdminDeposit`
+| `ERC7540SyncRedeem`
+| KYC-gated deposits, instant redemptions
```
Amxx: I'm worried about advertising that for KYC gatting. If KYC is not ok, we don't have any solution to unlock the funds without minting shares.

docs/modules/ROOT/pages/erc7540.adoc
ernestognw: Yes, we should mention them too

docs/modules/ROOT/pages/erc7540.adoc
ernestognw: Yeah, that's a good point. Instead of being specific about the delay case, I believe we should mention that depending on the vault, the exchange rate could be locked at fulfilling time or might be dependant on when the user claims

docs/modules/ROOT/pages/erc7540.adoc
ernestognw: True, let's just keep it as "gated"

Amxx (approved):

PR #230, opened 2026-04-22
ERC-7540: Epoch-based fulfillment strategy

ernestognw:
## Summary

- Adds `ERC7540EpochDeposit` and `ERC7540EpochRedeem` — epoch-based batch fulfillment strategies for ERC-7540 async vaults
- Adds `ERC7540EpochMock` for testing
- Extracted from #228 to reduce complexity of the base PR

Production equivalents: [Cove](https://github.com/Storm-Labs-Inc/cove-contracts-core/blob/master/src/BasketToken.sol), [Amphor](https://github.com/AmphorProtocol/asynchronous-vault/tree/main), [Lagoon](https://github.com/hopperlabsxyz/lagoon-v0/blob/main/src/v0.5.0/ERC7540.sol).

## Known pending work

- [x] Replace bare `require` statements in `_fulfillDeposit`/`_fulfillRedeem` with custom errors — typed errors `*TooEarly`, `*EmptyEpoch`, `*AlreadyFulfilled` are now thrown
- [x] Add fulfillment events to `_fulfillDeposit`/`_fulfillRedeem` — `ERC7540Epoch{Deposit,Redeem}Fulfilled` emitted with the locked rate
- [x] Verify whether `saturatingSub` is needed for the subtraction operations in `_consumeClaimable*` loops — resolved via `saturatingSub` + drained-epoch handling; residual-dust bound is documented as a NOTE in both contracts
- [x] Public getters for `_epochs` totals — `totalDepositAssets`/`totalDepositShares` and `totalRedeemAssets`/`totalRedeemShares` expose the per-epoch state
- [x] Public getter for `_memberOf` (per-controller epoch queue) — depends on OpenZeppelin/openzeppelin-contracts#6522 for paginated `DoubleEndedQueue` access
- [x] Add tests — `ERC7540EpochDeposit.test.js` (435 lines) and `ERC7540EpochRedeem.test.js` (410 lines)

## Test plan

- [x] Test epoch lifecycle: request → fulfill → claim across deposit and redeem
- [x] Test multi-epoch claims (controller with requests in multiple fulfilled epochs)
- [x] Test `_requestQueueLimit` enforcement (32 default)
- [x] Test partial claims within an epoch
- [x] Test that DoS via small requests is bounded per-account by queue limit
- [x] ERC-165 interface support validation — covered via `shouldSupportInterfaces` in the included `shouldBehaveLikeERC7540{Operator,Deposit,Redeem}` helpers

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Implemented ERC-7540 epoch-based asynchronous deposit protocol with automatic request batching, queue management, and pro-rata share distribution upon epoch fulfillment
  * Implemented ERC-7540 epoch-based asynchronous redemption protocol with automatic request batching, queue management, and pro-rata asset distribution upon epoch fulfillment
  * Added comprehensive test infrastructure for validating epoch-driven deposit and redemption request behavior


contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,231 @@
...
+            uint256 requested = details.requests[controller];
+            if (requested <= assets) _memberOf[controller].popFront();
+
+            uint256 batchAssets = requested.min(assets);
+            details.requests[controller] -= batchAssets; // May need saturatingSub for rounding handling
+            details.totalAssets -= batchAssets; // May need saturatingSub for rounding handling
+            assets -= batchAssets; // May need saturatingSub for rounding handling
+
+            uint256 batchShares = batchAssets.mulDiv(details.totalShares, details.totalAssets, Math.Rounding.Floor);
+            details.totalShares -= batchShares; // May need saturatingSub for rounding handling
```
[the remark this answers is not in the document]
ernestognw: Fixed in 2f3cc7d. `_consumeClaimableDeposit` and `_consumeClaimableMint` now compute `batchShares`/`batchAssets` against the pre-mutation totals before any decrements. Also added a contract-level NOTE clarifying that the rate is recomputed against remaining totals at each claim — the pro-rata invariant still holds exactly (sum of distributed shares equals the fulfilled `totalShares`), with floor-rounding residue absorbed by the final claim in the epoch.

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
```diff
@@ -0,0 +1,232 @@
...
+            );
+            if (requested <= assets) _memberOf[controller].popFront();
+
+            uint256 batchAssets = requested.min(assets);
+            details.totalAssets -= batchAssets; // May need saturatingSub for rounding handling
+            assets -= batchAssets; // May need saturatingSub for rounding handling
+
+            uint256 batchShares = batchAssets.mulDiv(details.totalShares, details.totalAssets, Math.Rounding.Floor);
+            details.requests[controller] -= batchShares; // May need saturatingSub for rounding handling
+            details.totalShares -= batchShares; // May need saturatingSub for rounding handling
```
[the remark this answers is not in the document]
ernestognw: Fixed in 2f3cc7d. `_consumeClaimableWithdraw` and `_consumeClaimableRedeem` now compute the batch conversion against the pre-mutation epoch totals before decrementing. Same NOTE applied at the contract level: the pro-rata invariant holds exactly (sum of distributed assets equals fulfilled `totalAssets`), with floor-rounding residue absorbed by the final claim.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /// @dev A request is claimable if its epoch has been fulfilled (`totalShares > 0`).
+    function _claimableDepositRequest(
+        uint256 requestId,
+        address controller
+    ) internal view virtual override returns (uint256) {
+        EpochDepositMetadata storage details = _epochs[requestId];
+        return details.totalShares == 0 ? 0 : details.requests[controller];
+    }
```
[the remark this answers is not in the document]
ernestognw: Resolved in 2f3cc7d by documenting the assumption rather than adding a `fulfilled` flag. A fulfillment with `totalShares == 0` is a confiscation event with no economic purpose (depositors would receive no shares for their deposited assets), so we assume admins won't pass 0. If 0 is passed accidentally, the call is a no-op and the admin can re-fulfill with the correct value. Added a NOTE in `_fulfillDeposit` natspec spelling out the assumption and noting that the recovery path is only valid as long as derived contracts preserve the no-side-effect semantics of this function — derived contracts that add side effects on fulfillment should restrict `totalShares != 0`.

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    /// @dev A request is claimable if its epoch has been fulfilled (`totalAssets > 0`).
+    function _claimableRedeemRequest(
+        uint256 requestId,
+        address controller
+    ) internal view virtual override returns (uint256) {
+        EpochRedeemMetadata storage details = _epochs[requestId];
+        return details.totalAssets == 0 ? 0 : details.requests[controller];
+    }
```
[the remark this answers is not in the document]
ernestognw: Resolved in 2f3cc7d via the same approach as the deposit sentinel: documented the assumption in `_fulfillRedeem` natspec instead of adding a `fulfilled` flag. Fulfillment at zero is a confiscation event with no economic purpose, so we assume admins won't pass 0; if 0 is passed accidentally the call is a no-op and the admin can re-fulfill. Recovery only holds as long as derived contracts preserve the no-side-effect semantics — derived contracts that add side effects on fulfillment should restrict `totalAssets != 0`.

ernestognw:
Worth flagging that we still don't have a way to observe the epochs a user has pending requests in — the `_epochs` / `_memberOf` getter item in the "Known pending work" list above. Toward that, I just opened OpenZeppelin/openzeppelin-contracts#6522, which adds a `values(start, end)` slice accessor to `DoubleEndedQueue`. Once it lands, we can expose a paginated getter here that returns the queue of pending requests per account without copying the whole deque.

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
```diff
@@ -0,0 +1,324 @@
...
+        uint256 epochId = currentRedeemEpoch();
+        _epochs[epochId].totalShares += shares;
+        _epochs[epochId].requests[controller] += shares;
+
+        (bool success, bytes32 lastEpochId) = _memberOf[controller].tryBack();
+        if (!success || lastEpochId != bytes32(epochId)) {
+            // Limit the number of pending epochs per account to avoid O(n) loop in
+            // _asyncMaxWithdraw and _asyncMaxRedeem being a concern. Users that have reached
+            // the limit should claim fulfilled requests to clean up the queue.
+            require(_memberOf[controller].length() < _requestQueueLimit());
```
luiz-lvj: we could add a custom error here
```suggestion
            require(_memberOf[controller].length() < _requestQueueLimit(), ERC7540EpochQueueLimitExceeded(controller));
```

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,324 @@
...
+        uint256 epochId = currentDepositEpoch();
+        _epochs[epochId].totalAssets += assets;
+        _epochs[epochId].requests[controller] += assets;
+
+        (bool success, bytes32 lastEpochId) = _memberOf[controller].tryBack();
+        if (!success || lastEpochId != bytes32(epochId)) {
+            // Limit the number of pending epochs per account to avoid O(n) loop in
+            // _asyncMaxDeposit and _asyncMaxMint being a concern. Users that have reached
+            // the limit should claim fulfilled requests to clean up the queue.
+            require(_memberOf[controller].length() < _requestQueueLimit());
```
luiz-lvj:
```suggestion
            require(_memberOf[controller].length() < _requestQueueLimit(), ERC7540EpochQueueLimitExceeded(controller));
```

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * @dev Iterates through the controller's epoch queue front-to-back, consuming assets
+     * and converting them to shares at each epoch's locked rate. Fully consumed epochs
+     * are dequeued.
+     */
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = 0;
+
+        while (assets > 0) {
+            uint256 epochId = uint256(_memberOf[controller].front());
```
luiz-lvj, line 316:
```suggestion
        while (assets > 0) {
            uint256 epochId = uint256(_memberOf[controller].front());
            if (totalDepositShares(epochId) == 0) break;
```

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,324 @@
...
+     * is O(1) (it sets `totalShares` for the entire epoch in a single write).
+     */
+    function _asyncMaxDeposit(address owner) internal view virtual override returns (uint256 assets) {
+        uint256 result = 0;
+        for (uint256 i = 0; i < _memberOf[owner].length(); ++i) {
+            uint256 epochId = uint256(_memberOf[owner].at(i));
+            result += _claimableDepositRequest(epochId, owner);
+        }
+        return result;
+    }
```
luiz-lvj, line 174:
```suggestion
    function _asyncMaxDeposit(address owner) internal view virtual override returns (uint256 assets) {
        uint256 result = 0;
        for (uint256 i = 0; i < _memberOf[owner].length(); ++i) {
            uint256 epochId = uint256(_memberOf[owner].at(i));
            if (totalDepositShares(epochId) == 0) break; // pending front → epochs behind it are blocked
            result += _claimableDepositRequest(epochId, owner);
        }
        return result;
    }
```

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
```diff
@@ -0,0 +1,324 @@
...
+    /**
+     * @dev Iterates through the controller's epoch queue front-to-back, consuming assets
+     * and converting them to shares at each epoch's locked rate. Fully consumed epochs
+     * are dequeued.
+     */
+    function _consumeClaimableWithdraw(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = 0;
+
+        while (assets > 0) {
+            uint256 epochId = uint256(_memberOf[controller].front());
```
luiz-lvj, line 316:
```suggestion
        while (assets > 0) {
            uint256 epochId = uint256(_memberOf[controller].front());
            if (totalDepositShares(epochId) == 0) break;
```

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,324 @@
...
+
+    /// @dev Sums claimable shares across all fulfilled epochs the `owner` participates in.
+    function _asyncMaxMint(address owner) internal view virtual override returns (uint256 shares) {
+        uint256 result = 0;
+        for (uint256 i = 0; i < _memberOf[owner].length(); ++i) {
+            uint256 epochId = uint256(_memberOf[owner].at(i));
```
luiz-lvj, line 180:
```suggestion
    function _asyncMaxMint(address owner) internal view virtual override returns (uint256 shares) {
        uint256 result = 0;
        for (uint256 i = 0; i < _memberOf[owner].length(); ++i) {
            uint256 epochId = uint256(_memberOf[owner].at(i));
            if (totalDepositShares(epochId) == 0) break; // pending front → epochs behind it are blocked
```

luiz-lvj (commented):
Hey @ernestognw, what is the expected behavior if there are two epochs in the queue and the admin, by mistake or something else, calls `_fulfillDeposit` on the newer epoch first? As of now, it would result in losses of the oldest epoch, which is not desired. This happens because:
- _consumeClaimable* consumes .front() (oldest) without checking it's fulfilled.
- _asyncMax* sums _claimableDepositRequest across the whole queue, skipping pending epochs but continuing to count fulfilled ones behind them.

Imo we could either revert if the admin tries to do that or only ever look at the contiguous fulfilled prefix (which I prefer). If the current behavior is intended we should at least document it.

ernestognw:
> Hey @ernestognw, what is the expected behavior if there are two epochs in the queue and the admin, by mistake or something else, calls `_fulfillDeposit` on the newer epoch first? 

Hey @luiz-lvj, thanks for the review, this is a great finding. I believe we just didn't consider an out of order fulfillment because it's arguably on the integrator side but it's true that we can make it safe from the base layer.

I believe the right behavior for the `_fulfill*` functions is to allow out-of-order fulfillment, but then the corresponding `_consume*` functions should stop at the first pending epoch. This way we still keep it as a "no op" (ie. meaning it won't force a revert).

I updated accordingly and added tests for these cases.

contracts/account/modules/ERC7579MultisigStorage.sol
Amxx, line 1: Why is this file modified in the PR ? Looks unrelated.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,367 @@
...
+ *
+ * The `requestId` returned by {requestDeposit} is the epoch ID. By default, epochs are weekly
+ * (`block.timestamp / 1 weeks`); override {currentDepositEpoch} to change the cadence or use
+ * manually-bumped epoch counters.
+ *
+ * Each account tracks its epoch memberships via a {DoubleEndedQueue} capped at
+ * {_depositRequestQueueLimit} entries (default: 32) to bound the O(n) loops in {_asyncMaxDeposit}
+ * and {_asyncMaxMint}. Users that hit the limit should claim fulfilled epochs to free up space.
+ *
+ * NOTE: Claims pay each controller's pro-rata share floor-rounded against the remaining epoch
```
Amxx, line 30: @arr00 can you please check if that reads well in english. I find it difficult, but maybe its just me.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    /// @inheritdoc ERC7540
+    function _isDepositAsync() internal pure virtual override returns (bool) {
+        return true;
+    }
+
+    /// @dev Returns the current epoch ID. Defaults to `block.timestamp / 1 weeks + 1`.
+    function currentDepositEpoch() public view virtual returns (uint256) {
+        // +1 to keep requestId != 0 so the strategy is never mistaken for ERC-7540's controller-only (requestId == 0) accounting mode.
+        return block.timestamp / 1 weeks + 1;
```
Amxx: This is correct, in the send that it increasses every week, but because january 1st 1970 was a thrusday, this will "update" at midnight on the wednesday to thursday night. 

I think it would be more natural to start the new epoch during the "sunday -> monday" night.

Also, we don't need the +1 since we are already more than 1 week after the "genesis" of block.timestamp.

Amxx:
```suggestion
        return (block.timestamp + 3 days) / 1 weeks;
```

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
```diff
@@ -0,0 +1,367 @@
...
+     * without minting shares to the caller. At realistic ERC-20 decimals the per-call drift is
+     * sub-cent. Deployers with non-standard decimals or a zero-drift requirement SHOULD override
+     * to revert when `batchAssets > 0 && batchShares == 0`.
+     */
+    function _consumeClaimableDeposit(uint256 assets, address controller) internal virtual override returns (uint256) {
+        uint256 shares = 0;
+
+        while (assets > 0) {
+            uint256 epochId = uint256(_memberOf[controller].front());
+            if (totalDepositShares(epochId) == 0) break; // oldest queued epoch is still Pending
```
Amxx, line 317: I think this is the correct behavior, particularly since we cannot remove an entry in the middle of a Dequeue ... but it also has a side effect that maybe should be documented.

- If a user participates in multiple epoch (A, B, C, D, ...)
- If the oldest one (epoch A) is not resolved.
- If the other epochs (B, C, D, ...) are resolved.

Then the user cannot claim. This means he cannot retrieve the funds for the other epochs. He also cannot "cleanup" the dequeue to be able to participate in other epoch if he has reached the threashold.

Its important that old epochs are resolved. Resolution doesn't need to be ordered, but it should be "close to it"

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    /// @inheritdoc ERC7540
+    function _isRedeemAsync() internal pure virtual override returns (bool) {
+        return true;
+    }
+
+    /// @dev Returns the current epoch ID. Defaults to `block.timestamp / 1 weeks + 1`.
+    function currentRedeemEpoch() public view virtual returns (uint256) {
+        // +1 to keep requestId != 0 so the strategy is never mistaken for ERC-7540's controller-only (requestId == 0) accounting mode.
+        return block.timestamp / 1 weeks + 1;
```
Amxx: similar to https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/230/changes#r3690447042

contracts/account/modules/ERC7579MultisigStorage.sol
ernestognw: Right, that was an accidentally-committed local change. Reverted to master on the branch — no longer in the PR.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
ernestognw: Applied — `(block.timestamp + 3 days) / 1 weeks`, dropped the `+ 1`. Symmetric on the redeem side. Tests, docs, and NatSpec updated to match.

contracts/token/ERC20/extensions/ERC7540EpochRedeem.sol
ernestognw: Applied, symmetric with the deposit side.

contracts/token/ERC20/extensions/ERC7540EpochDeposit.sol
ernestognw: Agreed on the design and on the concern. The side effect you describe is broader than what the existing `_fulfillDeposit` / `_fulfillRedeem` NOTE currently spells out — beyond the delayed-claim window, a controller whose oldest queued epoch stays Pending also (a) cannot claim any later-fulfilled epochs behind it, (b) cannot free queue slots by claiming, and (c) once at `_depositRequestQueueLimit` / `_redeemRequestQueueLimit`, cannot submit new requests until admin resolves the front. Will expand the NatSpec on the fulfill hooks (and the per-controller-queue section in `erc7540.adoc`) to state explicitly that admins are expected to resolve epochs close to FIFO order — strict ordering not required, but leaving old epochs Pending indefinitely locks affected controllers out of the vault.

PR #232, opened 2026-04-29
Document (and check) that ERC7540Delay modules do no use temporary share custody

Amxx:
## Summary by CodeRabbit

* **New Features**
  * Added deployment-time validation to ERC7540 delay modules to prevent invalid share origin and destination configurations. Contracts now enforce that certain address mechanisms must resolve to zero, preventing misconfiguration-related issues.

* **Tests**
  * Expanded test coverage to verify validation enforcement during contract deployment.


ernestognw (approved):
LGTM. Before merging I'd like to consider making `_redeemShareDestination` and `_depositShareOrigin` only `pure` instead of `view` so that they can't use storage. In that case the only way to mess things up would be through an upgrade

ernestognw:
lol

Screenshot 2026-04-29 at 8 19 39 p m

PR #234, opened 2026-04-29
ERC7540: add suport for ERC7575 share-to-vault lookup

Amxx:
Note: this is an optional interface. We can chose to not implement it.

## Summary by CodeRabbit

* **New Features**
  * Introduced asset-to-vault lookup capability with a new interface standard.
  * ERC7540 now supports vault asset discovery functionality.
  * Enhanced ERC-165 interface detection for better protocol compatibility.

* **Tests**
  * Updated test suite to verify new interface implementations.

### DIFF
--- a/contracts/token/ERC20/extensions/ERC7540.sol
+++ b/contracts/token/ERC20/extensions/ERC7540.sol
@@ -10,7 +10,7 @@ import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 import {LowLevelCall} from "@openzeppelin/contracts/utils/LowLevelCall.sol";
 import {Memory} from "@openzeppelin/contracts/utils/Memory.sol";
 import {IERC7540, IERC7540Operator, IERC7540Deposit, IERC7540Redeem} from "../../../interfaces/IERC7540.sol";
-import {IERC7575} from "../../../interfaces/IERC7575.sol";
+import {IERC7575, IERC7575Share} from "../../../interfaces/IERC7575.sol";
 
 /**
  * @dev Implementation of the ERC-7540 "Asynchronous ERC-4626 Tokenized Vaults" as defined in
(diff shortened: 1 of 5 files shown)


contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -202,6 +203,11 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
         return address(this);
     }
 
+    /// @inheritdoc IERC7575Share
+    function vault(address asset_) public view virtual returns (address) {
+        return asset_ == asset() ? address(this) : address(0);
```
ernestognw: Maybe?

```suggestion
        return share() == address(this) && asset_ == asset() ? address(this) : address(0);
```

Otherwise we might need a note in this function mentioning that if it's overridden, also the `share()` and `supportsInterface()` function should be overridden

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -131,6 +131,7 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
     function supportsInterface(bytes4 interfaceId) public view virtual override(IERC165, ERC165) returns (bool) {
         return
             interfaceId == (type(IERC4626).interfaceId ^ type(IERC7575).interfaceId) ||
+            interfaceId == type(IERC7575Share).interfaceId ||
```
Amxx:
```suggestion
            (interfaceId == type(IERC7575Share).interfaceId && share() == address(this)) ||
```

contracts/token/ERC20/extensions/ERC7540.sol
Amxx: is that waht you mean: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/234/changes#r3163832089?

ernestognw: Yes, pretty much 👍🏻

ernestognw (approved):

PR #235, opened 2026-04-29
Add note about totalAssets consistency in ERC7540

ernestognw:
## Summary by CodeRabbit

* **Documentation**
  * Enhanced documentation clarifying assumptions about ERC-20 token behavior and potential limitations when using non-standard tokens (transfer-fee, rebasing, etc.)
  * Added cautionary notes explaining edge cases where `totalAssets()` calculations may be affected, which could impact claim-path strategies

### DIFF
--- a/contracts/token/ERC20/extensions/ERC7540.sol
+++ b/contracts/token/ERC20/extensions/ERC7540.sol
@@ -56,6 +56,14 @@ import {IERC7575} from "../../../interfaces/IERC7575.sol";
  * using the controller's shares, and claim assets or shares on behalf of the controller. Users should only
  * approve operators they fully trust with both their assets and shares.
  * ====
+ *
+ * [CAUTION]
+ * ====
+ * This contract assumes the underlying `asset` is a well-behaved ERC-20: transfers move exactly the requested
+ * amount, balances do not change without explicit transfers, and `balanceOf` reports faithfully. Fee-on-transfer,
+ * rebasing, and similar non-standard asset behaviors are out of scope. When the asset misbehaves, internal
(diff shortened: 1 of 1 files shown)


Amxx (approved):

contracts/token/ERC20/extensions/ERC7540.sol
```diff
@@ -207,6 +215,12 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
...
      * converted into shares and must not be treated as yield for outstanding shareholders.
+     *
+     * NOTE: Internal flows preserve the invariant `balanceOf(asset, vault) >= totalPendingDepositAssets()` for
+     * any well-behaved ERC-20. Assets with transfer fees, negative rebases, or externally-mutable balances can
+     * violate it and cause this function to revert with an underflow. Strategies that read {totalAssets} on the
+     * claim path become uncallable in that state. Strategies that lock the rate at fulfillment time
+     * are unaffected.
      */
     function totalAssets() public view virtual override returns (uint256) {
         return IERC20(asset()).balanceOf(address(this)) - totalPendingDepositAssets();
```
Amxx, line 226: I'm wondering, should we do a `Math.saturatingSub` here ?

Basically, is it better to revert or to report zero ?

ernestognw: I think it's better to revert. Otherwise `_convertToShares(x) = x * (totalSupply + 10^offset) / 1 = x * (totalSupply + 10^offset)`. This means that minting shares will price assets at `totalSupply + 10^offset` which I believe is worse; the first who claim would do it at an inflated rate while the last to claim will be affected.

PR #236, opened 2026-04-29
Warn users about dead address controller in ERC-7540

ernestognw:
## Summary by CodeRabbit

* **Documentation**
  * Enhanced documentation for deposit and redemption request functions to clarify controller authorization requirements and the implications of misconfiguring controllers without proper claim execution capabilities, including warnings about potential asset and share locking scenarios.

### DIFF
--- a/contracts/token/ERC20/extensions/ERC7540.sol
+++ b/contracts/token/ERC20/extensions/ERC7540.sol
@@ -394,6 +394,12 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
      * * {_isDepositAsync} must return `true`.
      * * `owner` must be `msg.sender` or `msg.sender` must be an approved operator of `owner`.
      * * `owner` must have approved the vault for at least `assets` of the underlying token.
+     *
+     * NOTE: The `controller` is the only address authorized to claim the resulting Request. Passing an address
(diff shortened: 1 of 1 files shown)


Amxx (approved):

Amxx:
It feels a bit overkill. I don't think we have a warning in ERC20 that says "be carefull that if you transfer assets to address 0xdead your funds are likelly lost forever. 

But the doc is technically correct, and since this standard is not very well known, it might be usefull to document it.

This all makes me think of ERC7821 usage of address(0) for self reference. Maybe it could be replicated here:
```solidity
    function requestDeposit(
        uint256 assets,
        address controller,
        address owner
    ) public virtual onlyOperatorOrController(_isDepositAsync(), owner, _msgSender()) returns (uint256) {
        return _requestDeposit(assets, controller == address(0) ? _msgSender() : controller, owner, 0);
    }
```

ernestognw:
Yes. It also feels overkill to me. A similar issue was flagged in open accounts framework and I documented it this way. Imo, this issue is now relevant because it'll be flagged by AI auditors everytime.

The 7821 case is interesting, but there, the ERC does specify the usage of address(0) to indicate the sender. This is not the case, though, it could be a custom choice we implement deliberately. No strong opinion

PR #238, opened 2026-04-29
Compute claimable consumption rates from internal strategy state

ernestognw:
## Summary by CodeRabbit

* **Refactor**
  * Enhanced pro-rata accounting for async deposit and redemption claims to improve consistency and accuracy in exchange rate calculations.
  * Optimized claim consumption logic to better reflect strategy state in locked asset calculations, reducing edge-case errors.
  * Updated documentation for async strategy hooks to clarify their role in claim-rate computations.

### DIFF
--- a/contracts/token/ERC20/extensions/ERC7540.sol
+++ b/contracts/token/ERC20/extensions/ERC7540.sol
@@ -887,15 +887,24 @@ abstract contract ERC7540 is ERC165, ERC20, IERC4626, IERC7540 {
      */
     function _consumeClaimableRedeem(uint256 /*shares*/, address /*controller*/) internal virtual returns (uint256);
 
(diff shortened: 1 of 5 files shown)


Amxx:
I'm quite confuse by what this change implies.

I designed the `_asyncMaxXxx` functions to be a direct implementation of the corresponding `maxXxx` function is the async case. I never imagined that the external and internal function would behave differently.

My understanding is that this PR considers a case where "the external function is modified (overriden) but the internal is not, and so using the internal is the thing to do". I just can't wrap my head around what it means to override the external instead of the internal, and having two behavior cohexist in the same contract.

When there is no override the propose change are fine. They don't modify the behavior at all. What I'm trying to get to is, why would someone override the external instead of the internal, and what would be the best behavior in that case.

ernestognw:
We agreed not to pursue this

PR #248, opened 2026-07-06
Add AccessManager role-based account factory

Amxx:
## Summary

Adds a role-scoped signer primitive and a companion account + factory for it. Authority is delegated to whoever currently holds a given role in an `IAccessManager`, so granting/revoking role membership immediately grants/revokes control without touching the signer or the account.

- `SignerRole` (`contracts/utils/cryptography/signers/SignerRole.sol`): an `AbstractSigner` whose `_rawSignatureValidation` accepts a signature only when the recovered signer currently holds the bound `roleId()` in `accessManager()` with a 0 execution delay. Signature layout is `[20-byte signer address][inner signature]`; the inner signature is checked with `SignatureChecker` so both EOAs and ERC-1271 signers are supported. `accessManager()` and `roleId()` resolution is left to concrete implementations.
- `RoleAccount` (`contracts/account/RoleAccount.sol`): a concrete signer/executor composing `SignerRole` + `ERC7739` (replay-safe ERC-1271 typed-data wrapping) + `ERC7821` (batched execution). It reads `accessManager` (20 bytes) and `roleId` (8 bytes) from the clone's immutable args, and authorizes `ERC7821` execution for any current unrestricted role member in addition to the default self-call path. NOTE: This is *not* an ERC-4337 `Account` — it does not inherit `Account`/`AccountCore` and does not implement `validateUserOp`. It is an ERC-1271 signer with ERC-7821 batch execution.
- `RoleAccountFactory` (`contracts/account/RoleAccountFactory.sol`): permissionless factory that deploys a deterministic `RoleAccount` clone per `(accessManager, roleId)` pair via `Clones.cloneDeterministicWithImmutableArgs` with `salt = bytes32(0)`. Address is fully derived from the immutable-args pair, so it is predictable off-chain and can be funded / authorized before deployment. Emits `RoleAccountDeployed(accessManager, roleId, account)`.

## Notes

- Direction change from earlier revisions: the previous `AccessManagerWithRoleAccounts` extension was dropped in favor of an external factory, so already-deployed access managers can be used unchanged and the manager stays agnostic to consumers (Governor, Account, etc.).
- Formatted with prettier and passes solhint via the pre-commit hooks.

## Summary by CodeRabbit

* **New Features**
  * Added role-based account handlers for more secure signature validation.
  * Enabled deterministic creation and lookup of handler addresses for specific roles.
  * Prevented public-role use when generating role-specific handler addresses.


arr00:
Thoughts on allowing execution from the handler by anyone with the role? Maybe it is overkill but it makes the usecase more generic. Then we don't need to use the phrasing `ConfidentialHandler`--it would just be a shared account across all role holders.

Amxx:
> Thoughts on allowing execution from the handler by anyone with the role?

The current version supports execution through the ERC-7821 inheritance

contracts/access/README.adoc
```diff
@@ -6,7 +6,14 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/community-
...
+ * {AccessManagerWithRoleAccounts}: An `AccessManager` extension that exposes a deterministic account ({RoleAccount}) for each role, so the current members of a role can produce signatures and execute calls on its behalf.
 
 == AccessManager
 
 {{AccessManagerLight}}
+
+== AccessManager with role accounts
+
+This variant of `AccessManager` associates a deterministic account with each role. The account (a {RoleAccount}, built on top of the {RoleSigner} abstract signer) acts on behalf of whoever currently holds the role: any member can produce ERC-1271 signatures for it or trigger batched ERC-7821 executions through it. Because authorization is resolved live against the manager, granting or revoking the role immediately grants or revokes control of the account.
+
```
arr00:
```suggestion

```

contracts/access/README.adoc
```diff
@@ -6,7 +6,9 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/community-
 This directory contains utility contracts to restrict access control in smart contracts. These include:
 
  * {AccessManagerLight}: A simpler version of an AccessManager that uses `bytes8` roles to allow function calls identified by their 4-bytes selector.
+ * {AccessManagerWithRoleAccounts}: An `AccessManager` extension that exposes a deterministic account ({RoleAccount}) for each role, so the current members of a role can produce signatures and execute calls on its behalf.
 
 == AccessManager
 
 {{AccessManagerLight}}
+{{AccessManagerWithRoleAccounts}}
```
arr00:
```suggestion

{{AccessManagerWithRoleAccounts}}
```

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,78 @@
...
+    /**
+     * @dev Returns the role id this signer is bound to, decoded from the clone's immutable arguments.
+     *
+     * Reverts with {InvalidCloneArgs} when the immutable arguments are not exactly a `uint64`, which
+     * happens when the contract is not deployed as a {Clones}-with-immutable-args proxy.
+     */
+    function roleId() public view virtual returns (uint64) {
+        bytes memory cloneArgs = Clones.fetchCloneArgs(address(this));
+        if (cloneArgs.length != 8) revert InvalidCloneArgs();
+        return uint64(bytes8(cloneArgs));
```
Amxx: suggestion: if the args are invalid, don't revert. Instead return 0 so that the admin of the access manager can control it (and no fund are eventualy lost).
```suggestion
        return cloneArgs.length >= 8 ? uint64(bytes8(cloneArgs)) : 0;
```

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * A `signature` shorter than the 20-byte address prefix is rejected without reverting.
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        if (signature.length < 20) return false;
+        address signer = address(bytes20(signature));
+        return SignatureChecker.isValidSignatureNow(signer, hash, signature[20:]) && _isUnrestrictedMember(signer);
```
Amxx:
```suggestion
        address signer = address(bytes20(signature));
        return 
            signature.length >= 20 
                && SignatureChecker.isValidSignatureNow(signer, hash, signature[20:])
                && _isUnrestrictedMember(signer);
```

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,78 @@
...
+     * @dev Returns the role id this signer is bound to, decoded from the clone's immutable arguments.
+     *
+     * Returns 0 (the {IAccessManager} admin role) when the immutable arguments are not at least 8 bytes,
+     * which happens when the contract is not deployed as a {Clones}-with-immutable-args proxy. Rather than
+     * reverting, this falls back to the admin role so the access manager's admin retains control over the
+     * signer and no funds are permanently lost.
+     */
+    function roleId() public view virtual returns (uint64) {
+        bytes memory cloneArgs = Clones.fetchCloneArgs(address(this));
+        return cloneArgs.length >= 8 ? uint64(bytes8(cloneArgs)) : 0;
```
arr00: Note the docs from `fetchCloneArgs`
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/b4f9524d87bdb9c192cf1b0fcdcc9d688c3ba543/contracts/proxy/Clones.sol#L258-L259

If this is not deployed via clones, it returns nonsense. I think its fine but should be explicitly documented.

Amxx: What I'm thinking is that some contract that inherit from this signer could put other stuff in the immutable args. In that case the first 8 bytes are the role, and the following ones are something else, and it's fine to have more than 8 bytes.

Note maybe we should have an immutable arg, set at construction (of the implementation) that defines the offset in the immutable data (if immutable data contains more than just the role. In our case the offset would be 0)

arr00: I think what you are saying is fine. But the issue is related to if someone deploys via create1. Then we just get nonsense returned (part of the contract bytecode).

Amxx: usign either CREATE or CREATE2 are both safe as long as the bytecode being deployed is the correct one (the one returned by `Clones._cloneCodeWithImmutableArgs`)

arr00: If you just naively deploy this contract via create1 then try to utilize it, the role id will be essentially random bytes. We should warn users against doing this.

Amxx: That has nothing to do with create1 (which BTW is not a thing, its CREATE or CREATE2 -- or CREATE3).

And the contract has to be deployed at some point, otherwize clones cannot point to it. The current mechanism does such a deployment in `AccessManagerWithRoleAccounts`:

```solidity
    address private immutable _template = address(new RoleAccount(this));
```

Yes, the template should not be used as a live instance.

Do you want to add something like 
```solidity
contract RoleSigner {
    address private immutable _self = address(this);
    
    // ...
    
    function roleId() public view virtual returns (uint64) {
        if (_self == address(this)) {
            return 0; // direct call to the implementation.
        } else {
            bytes memory cloneArgs = Clones.fetchCloneArgs(address(this)).splice(_roleIdOffset);
            return cloneArgs.length >= 8 ? uint64(bytes8(cloneArgs)) : 0;
        }
    }
}
```

arr00: Yes exactly. I think this is necessary. And yes I was getting at using the template as a live instance.

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,83 @@
...
+ * Instead of holding its own key material, this signer is bound to a single `roleId` and accepts a
+ * signature only when it was produced by an address that currently holds that role in the associated
+ * {accessManager} with 0 delay. This lets a role behave as a shared signer: membership can be granted
+ * or revoked through the access manager without redeploying or reconfiguring the signer.
+ *
+ * The `roleId` is not stored in a regular state variable. This contract is meant to be deployed
+ * behind a {Clones}-with-immutable-args proxy that carries the target role id as its immutable
+ * arguments (see {AccessManagerWithRoleAccounts}, which deploys one clone per role).
+ */
+contract RoleSigner is AbstractSigner {
```
arr00:
```suggestion
abstract contract RoleSigner is AbstractSigner {
```

contracts/utils/cryptography/signers/RoleSigner.sol
Amxx: Implemented in 76003e4.

Proxies that are not clone are still a possible issue, but direct calls to the implementation are now safe.

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,46 @@
...
+ * the role immediately grants or revokes control of the account, without touching the account itself.
+ *
+ * It composes:
+ *
+ * * {RoleSigner}: gates signature validation on role membership.
+ * * {ERC7739}: wraps signatures as ERC-7739 nested typed data / personal-sign messages to provide
+ *   replay-safe ERC-1271 validation on top of {RoleSigner}.
+ * * {ERC7821}: minimal batch executor.
+ *
+ * These accounts are intended to be deployed as {Clones}-with-immutable-args proxies, one per role, by
```
arr00:
```suggestion
 * These accounts are intended to be deployed as `Clones.cloneWithImmutableArgs`, one per role, by
```

arr00 (dismissed):
@ernestognw please review when you get a chance

contracts/access/manager/AccessManagerWithRoleAccounts.sol
```diff
@@ -0,0 +1,62 @@
...
+ * the clone at that address when needed.
+ *
+ * WARNING: A role account grants control to *every current member* of its role. For the special
+ * `PUBLIC_ROLE` (`type(uint64).max`), which every address belongs to, this means the account is
+ * controllable by anyone.
+ *
+ * NOTE: {deployRoleAccount} is permissionless. Because the deployment is deterministic and behaviorally
+ * fixed, this is harmless (front-running it only produces the same account).
+ */
+contract AccessManagerWithRoleAccounts is AccessManager {
```
james-toussaint: I would consider moving this factory outside the `AccessManager` since there is no **required** need for having it here beyond having a few nice-to-have.
Having it externally keeps the complex `AccessManager` clean & allows to plug the role account system later if needed.

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,73 @@
...
+ * Instead of holding its own key material, this signer is bound to a single `roleId` and accepts a
+ * signature only when it was produced by an address that currently holds that role in the associated
+ * {accessManager} with 0 delay. This lets a role behave as a shared signer: membership can be granted
+ * or revoked through the access manager without redeploying or reconfiguring the signer.
+ *
+ * How the `roleId` is resolved is left to implementations (see {roleId}). {RoleAccount} decodes it
+ * from the immutable arguments of a `Clones.cloneWithImmutableArgs` proxy (see
+ * {AccessManagerWithRoleAccounts}, which deploys one clone per role).
+ */
+abstract contract RoleSigner is AbstractSigner {
```
james-toussaint: I'm considering having a multi-sig `RoleMultiSigner` instead, where `RoleSigner` is `RoleMultiSigner` with `threshold=1`. (`RoleSigner` would be added in natspec examples).

```
abstract contract RoleMultiSigner is AbstractSigner {
    ...
    function threshold() public view virtual returns (uint64);
    
        bytes32 hash,
        bytes calldata signature
        ...
        (bytes[] memory signers, bytes[] memory signatures) = abi.decode(signature, (bytes[], bytes[]));
        ...
    }
}

abstract contract RoleSigner is RoleMultiSigner {
    ...
    function threshold() public view virtual override returns (uint64) {
        return 1;
    }
}
```

contracts/utils/cryptography/signers/RoleSigner.sol
```diff
@@ -0,0 +1,73 @@
...
+     * A `signature` shorter than the 20-byte address prefix is rejected without reverting.
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        address signer = address(bytes20(signature));
+        return
+            signature.length >= 20 &&
+            SignatureChecker.isValidSignatureNow(signer, hash, signature[20:]) &&
```
james-toussaint:
```suggestion
            SignatureChecker.isValidSignatureNowCalldata(signer, hash, signature[20:]) &&
```

contracts/utils/cryptography/README.adoc
```diff
@@ -8,6 +8,7 @@ A collection of contracts and libraries that implement various signature validat
  * {ZKEmailUtils}: Library for ZKEmail signature validation utilities, enabling email-based authentication through zero-knowledge proofs.
  * {DKIMRegistry}: Implementation of https://eips.ethereum.org/EIPS/eip-7969[ERC-7969] to enable onchain verification of DomainKeys Identified Mail (DKIM) signatures.
  * {SignerZKEmail}: Implementation of an https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#AbstractSigner[AbstractSigner] that enables email-based authentication through zero-knowledge proofs.
+ * {RoleSigner}: Implementation of an https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#AbstractSigner[AbstractSigner] whose authority is delegated to the members of a role tracked by an `AccessManager`.
```
james-toussaint: nit: what is the ordering rule(s) here and there, a-z, add date?

contracts/utils/cryptography/signers/RoleSigner.sol
arr00: I think this is an interesting independent signer. We need the `RoleSigner` as is for the application we have in mind. But having a multisig signer based on the role makes a lot of sense as well.

arr00: I guess we could just have a multisig signer with 1 threshold.

Amxx: We are on community. We can iterate, replace contracts, do breaking changes, ...

No need to perfectly fine tune things like on vanilla in case it gets released early.

I would recommend merging a simple version and then we can do more complexe signers/constructions in follow-up PRs

arr00 (dismissed):

contracts/utils/cryptography/README.adoc
arr00: Believe its append only per section. Inserted above `ERC7913ZKEmailVerifier` since the verifiers section is below the signers section.

contracts/access/manager/AccessManagerWithRoleAccounts.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// @dev Implementation cloned (with the role id as immutable args) to produce each {RoleAccount}.
+    address private immutable _template = address(new RoleAccount(this));
+
+    constructor(address initialAdmin) AccessManager(initialAdmin) {}
+
+    /**
+     * @dev Returns the deterministic address of the {RoleAccount} for `roleId`, whether or not it has
+     * already been deployed.
+     */
+    function getRoleAccount(uint64 roleId) public view returns (address) {
```
gonzaotc:
```suggestion
    function getRoleAccount(uint64 roleId) public view virtual returns (address) {
```

contracts/access/manager/AccessManagerWithRoleAccounts.sol
```diff
@@ -0,0 +1,62 @@
...
+                abi.encodePacked(roleId),
+                _roleToSalt(roleId)
+            );
+    }
+
+    /**
+     * @dev Deploys the {RoleAccount} clone for `roleId` at its deterministic address and returns it.
+     * Reverts if the account for `roleId` has already been deployed.
+     */
+    function deployRoleAccount(uint64 roleId) public returns (address) {
```
gonzaotc:
```suggestion
    function deployRoleAccount(uint64 roleId) public virtual returns (address) {
```

contracts/access/manager/AccessManagerWithRoleAccounts.sol
```diff
@@ -0,0 +1,62 @@
...
+        return Clones.cloneDeterministicWithImmutableArgs(_template, abi.encodePacked(roleId), _roleToSalt(roleId));
+    }
```
gonzaotc: If anyone wants to change the role encoding, they must override both `deployRoleAccount` and `getRoleAccount`. With the following, they can just override `_roleToArgs`

```suggestion
    /**
     * @dev Deploys the {RoleAccount} clone for `roleId` at its deterministic address and returns it.
     * Reverts if the account for `roleId` has already been deployed.
     */
    function deployRoleAccount(uint64 roleId) public returns (address) {
        return Clones.cloneDeterministicWithImmutableArgs(_template, abi.encodePacked(roleId), _roleToSalt(roleId));
    }

    /**
     * @dev Encodes the role id as the immutable args for the {RoleAccount} clone.
     */
    function _roleToArgs(uint64 roleId) internal view virtual returns (bytes memory) {
        return abi.encodePacked(roleId);
    }
```

contracts/access/manager/AccessManagerWithRoleAccounts.sol
gonzaotc: Add an event for indexing role accounts creation?; `RoleAccountDeployed(uint64 indexed roleId, address 
  account)`

contracts/account/RoleAccount.sol
gonzaotc, line 1: I don't think we need to expose this contract. Instead, the account construction should be achieved using our Accounts framework; which consists of extending OpenZeppelin's `Account`, choosing a `Signer`, and the desired optional extensions.

I'd recommend to do that leveraging by the Accounts Wizard (https://wizard.openzeppelin.com/#account), which let's you choose exactly what you need (rather than forcing ERC-7821 as this contract does)

The Wizard output for a feature-rich RoleAccount would look something like this:
```
contract MyAccount is Account, EIP712, ERC7739, RoleSigner, ERC7821, ERC721Holder, ERC1155Holder {
      uint64 private immutable _roleId;

      constructor(IAccessManager accessManager, uint64 roleId_)
          EIP712("MyAccount", "1")
          RoleSigner(accessManager)
      {
          _roleId = roleId_;
      }

      function roleId() public view virtual override returns (uint64) {
          return _roleId;
      }

      function _erc7821AuthorizedExecutor(
          address caller,
          bytes32 mode,
          bytes calldata executionData
          return
              caller == address(entryPoint()) || _isUnrestrictedMember(caller) ||
              super._erc7821AuthorizedExecutor(caller, mode, executionData);
      }
}
```

contracts/access/manager/AccessManagerWithRoleAccounts.sol
gonzaotc: I strongly agree with James here. 

1. This extension is hardcoding the shape of the contract that consumes the roles (in this case, the shape of the account), which I don't think it's necessary. To me, the AccessManager should only be the source of truth for roles, but it doesn't need to be aware which contracts consume that state.
i.e: someone may want to have a single AccessManager permissionning Accounts of different shapes, OR even permissionning different contracts such as Accounts AND a `GovernorRoleVotes` that only lets role holders to vote. 

2. I like James proposal to avoid the extension (https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/260) via an external factory, since it makes any already deployed AccessManager compatible with a RoleSigner. Technically, both proposals could live together; the extension for new Access Managers and the factory for already deployed ones, but I think it the external factory is superior since the AccessManager is agnostic to it's consumers.

contracts/access/manager/AccessManagerWithRoleAccounts.sol
arr00: sure

contracts/account/RoleAccount.sol
arr00: I highly doubt we'd add this to wizard. We need this feature in its entirety for a project we are building related to Zama. Its generalized enough to fit here but certainly not general enough to add to wizard.

gonzaotc:
Following these comments: 

1. [Decoupling the AccessManager from the Account](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248#discussion_r3691931659)
2.  [Moving the factory outside the AccessManager](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248#discussion_r3673265517)

Considering a lot of projects still use other less powerful Access Control mechanisms, what do you think of extending the idea to other authorization mechanisms:

- **Abstract SignerRole**
```
/**
 * @dev Implementation of {AbstractSigner} whose authority is delegated to the members of a role
 * of a role-based authorization system.
 */
abstract contract SignerRole is AbstractSigner {
    /**
     * @dev Returns whether `account` currently holds the role.
     */
    function _hasRole(address account) internal view virtual returns (bool);

    /**
     * @inheritdoc AbstractSigner
     */
        bytes32 hash,
        bytes calldata signature
        address signer = address(bytes20(signature));
        return
            signature.length >= 20 &&
            SignatureChecker.isValidSignatureNowCalldata(signer, hash, signature[20:]) &&
            _hasRole(signer);
    }
}
```


- **SignerAccessControlRole**
> I have a contract controlled by `AccessControl` and I want accounts per role that can hold assets.
```
/**
 * @dev Implementation of {SignerRole} for an {IAccessControl}.
 */
abstract contract SignerAccessControlRole is SignerRole {
    /**
     * @dev The {IAccessControl} system this signer is bound to.
     */
    IAccessControl public immutable accessControl;
    
    /**
     * @dev Thrown when the {IAccessControl} is the zero address.
     */
    error InvalidAccessControl();

    constructor(IAccessControl accessControl_) {
        accessControl = accessControl_;
        require(address(accessControl_) != address(0), InvalidAccessControl());
    }

    /**
     * @dev The role id this signer is bound to in the {IAccessControl} contract.
     */
    function roleId() public view virtual returns (bytes32);

    /**
     * @inheritdoc SignerRole
     */
    function _hasRole(address account) internal view virtual override returns (bool) {
        return accessControl.hasRole(roleId(), account);
    }
}
```

- **SignerAccessManagerRole**
> I have one or multiple contracts controlled by `AccessManager` and I want accounts per role that can hold assets.
```
abstract contract SignerAccessManagerRole is SignerRole {
    /**
     * @dev The {IAccessManager} system this signer is bound to.
     */
    IAccessManager public immutable accessManager;

    /**
     * @dev Thrown when the {IAccessManager} is the zero address.
     */
    error InvalidAccessManager();

    constructor(IAccessManager accessManager_) {
        require(address(accessManager_) != address(0), InvalidAccessManager());
        accessManager = accessManager_;
    }

    /**
     * @dev The role id this signer is bound to in the {IAccessManager} contract.
     */
    function roleId() public view virtual returns (uint64);

    /**
     * @inheritdoc SignerRole
     */
    function _hasRole(address account) internal view virtual override returns (bool) {
        (bool isMember, uint32 executionDelay) = accessManager.hasRole(roleId(), account);
        return isMember && executionDelay == 0;
    }
}
```

- **SignerOwnable**
> I have a contract controlled by `Ownable` and I want an Account that can hold assets. i.e: my protocol is Ownable and has a Treasury. I want to synchronize their ownership, so that if the Owner changes into a Governor, the Governor owns the Treasury automatically. 
```
abstract contract SignerOwnable is SignerRole {
    /**
     * @dev The {Ownable} system this signer is bound to.
     */
    Ownable public immutable ownable;
    
    /**
     * @dev Thrown when the {Ownable} is the zero address.
     */
    error InvalidOwnable();

    constructor(Ownable ownable_) {
        ownable = ownable_;
        require(address(ownable_) != address(0), InvalidOwnable());
    }

    /**
     * @inheritdoc SignerRole
     */
    function _hasRole(address account) internal view virtual override returns (bool) {
        return ownable.owner() == account;
    }
}
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+     */
+    function roleId() public view virtual override returns (uint64 roleId_) {
+        require(_self != address(this), DirectCallNotAllowed());
+        (, roleId_) = _fetchArgs();
+    }
+
+    /// @dev Decodes the access manager and role id packed in the clone's immutable arguments.
+    function _fetchArgs() private view returns (IAccessManager, uint64) {
+        bytes memory cloneArgs = Clones.fetchCloneArgs(address(this));
+        require(cloneArgs.length >= 28, MissingImmutableArgs());
```
Amxx: Note, this could be 
```suggestion
        require(cloneArgs.length == 28, MissingImmutableArgs());
```

contracts/access/manager/AccessManagerWithRoleAccounts.sol
Amxx: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248/changes/1d44b02672d50e4eebcd01bcddb11d5c2ecdfc12

contracts/account/RoleAccount.sol
gonzaotc: I can be biased but I'd argue the concept of managing an Account via roles is general enough to fit even the wizard, the only missing piece is the new `AbstractSigner` (RoleSigner) being introduced in this PR and a contract that holds the roles state. (AccessManager or other role based authorization statefull contract)

See [comment](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248#issuecomment-5146125405) about generalizing `RoleSigner` further to also support other role based authorization mechanisms. 

I would like to be able to use this feature with any Account, but I'm ok if it gets initially shipped with a default Account template that is overridable with my custom Account

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 28-07-2026
+
+- `SignerRole`: Add an `AbstractSigner` whose authority is delegated to the members of a role tracked by an `AccessManager`. A signature is accepted only when its `[signer][inner signature]` payload is valid and the signer currently holds the bound role.
```
arr00:
```suggestion
- `SignerRole`: Add an `AbstractSigner` whose authority is delegated to the members of a role tracked by an `IAccessManager`. A signature is accepted only when its `[signer][inner signature]` payload is valid and the signer currently holds the bound role.
```

contracts/utils/cryptography/README.adoc
```diff
@@ -8,6 +8,7 @@ A collection of contracts and libraries that implement various signature validat
  * {ZKEmailUtils}: Library for ZKEmail signature validation utilities, enabling email-based authentication through zero-knowledge proofs.
  * {DKIMRegistry}: Implementation of https://eips.ethereum.org/EIPS/eip-7969[ERC-7969] to enable onchain verification of DomainKeys Identified Mail (DKIM) signatures.
  * {SignerZKEmail}: Implementation of an https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#AbstractSigner[AbstractSigner] that enables email-based authentication through zero-knowledge proofs.
+ * {SignerRole}: Implementation of an https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#AbstractSigner[AbstractSigner] whose authority is delegated to the members of a role tracked by an `AccessManager`.
```
arr00:
```suggestion
 * {SignerRole}: Implementation of an https://docs.openzeppelin.com/contracts/5.x/api/utils/cryptography#AbstractSigner[AbstractSigner] whose authority is delegated to the members of a role tracked by an `IAccessManager`.
```

contracts/account/README.adoc
```diff
@@ -4,6 +4,8 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/community-
 
 This directory includes contracts to build accounts for ERC-4337. These include:
 
+ * {RoleAccount}: An account bound to a role of an `AccessManager` (through {SignerRole}), acting on behalf of the current members of that role for both ERC-1271 signatures and ERC-7821 batched execution.
```
arr00:
```suggestion
 * {RoleAccount}: An account bound to a role of an `IAccessManager` (through {SignerRole}), acting on behalf of the current members of that role for both ERC-1271 signatures and ERC-7821 batched execution.
```

contracts/account/README.adoc
```diff
@@ -4,6 +4,8 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/community-
...
+ * {RoleAccountFactory}: A factory that deploys a deterministic account ({RoleAccount}) for any role of any `AccessManager`, so the current members of a role can produce signatures and execute calls on its behalf.
```
arr00:
```suggestion
 * {RoleAccountFactory}: A factory that deploys a deterministic account ({RoleAccount}) for any role of any `IAccessManager`, so the current members of a role can produce signatures and execute calls on its behalf.
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * It composes:
+ *
+ * * {SignerRole}: gates signature validation on role membership.
+ * * {ERC7739}: wraps signatures as ERC-7739 nested typed data / personal-sign messages to provide
+ *   replay-safe ERC-1271 validation on top of {SignerRole}.
+ * * {ERC7821}: minimal batch executor.
+ *
+ * These accounts are intended to be deployed as `Clones.cloneWithImmutableArgs`, one per (access manager,
+ * role) pair, by {RoleAccountFactory}. The access manager and role id are encoded in the clone's immutable
```
arr00:
```suggestion
 * These accounts are intended to be deployed as `Clones.cloneWithImmutableArgs`, once per (access manager,
 * role) pair, by {RoleAccountFactory}. The access manager and role id are encoded in the clone's immutable
```

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,61 @@
...
+            Clones.predictDeterministicAddressWithImmutableArgs(
+                address(_template),
+                abi.encodePacked(accessManager, roleId),
+                bytes32(0)
+            );
+    }
+
+    /**
+     * @dev Deploys the {RoleAccount} clone for `roleId` on `accessManager` at its deterministic address and
+     * returns it. Reverts if that account has already been deployed.
```
arr00:
```suggestion
     * returns it. Reverts if the account has already been deployed.
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,69 @@
...
+     * which decodes it from the clone's immutable arguments).
+     */
+    function roleId() public view virtual returns (uint64);
+
+    /**
+     * @dev Returns whether `account` currently holds {roleId} in the {accessManager} and has no execution delay.
+     * This signer does not allow roles with execution delays to interact with it.
+     */
+    function _isUnrestrictedMember(address account) internal view virtual returns (bool) {
+        (bool hasRole, uint32 executionDelay) = accessManager().hasRole(roleId(), account);
```
james-toussaint, line 55: Minor, considering `accessManager` & `roleId` are used together I would retrieve them by fetching once:
```suggestion
        (IAccessManager accessManager_, uint64 roleId_) = _fetchArgs();
        (bool hasRole, uint32 executionDelay) = accessManager_.hasRole(roleId_, account);
```
with
```
    /// @dev Decodes the access manager and role id packed in the clone's immutable arguments.
    function _fetchArgs() internal view virtual override returns (IAccessManager, uint64) {
        require(_self != address(this), DirectCallNotAllowed());
        bytes memory cloneArgs = Clones.fetchCloneArgs(address(this));
        require(cloneArgs.length >= 28, MissingImmutableArgs());

        bytes28 data = bytes28(cloneArgs);
        return (IAccessManager(address(bytes20(data))), uint64(bytes8(data << 160)));
    }
```

arr00: Note this means we'd need to move the immutables back to the signer from the account. I prefer to have the immutablees and getters implemented at the account level. Curious on your thoughts @gonzaotc

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+        require(_self != address(this), DirectCallNotAllowed());
+        (accessManager_, ) = _fetchArgs();
+    }
+
+    /**
+     * @dev Returns the role id this signer is bound to, decoded from the clone's immutable arguments.
+     *
+     * Reverts with {DirectCallNotAllowed} when called on the implementation directly (i.e. not through a
+     * `Clones.cloneWithImmutableArgs` proxy), and with {MissingImmutableArgs} when the immutable arguments
+     * are malformed.
```
arr00:
```suggestion
     * are not at least 28 bytes.
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+    error MissingImmutableArgs();
+
+    constructor() EIP712("RoleAccount", "1.0.0") {}
+
+    /**
+     * @dev Returns the access manager this account is bound to, decoded from the clone's immutable arguments.
+     *
+     * Reverts with {DirectCallNotAllowed} when called on the implementation directly (i.e. not through a
+     * `Clones.cloneWithImmutableArgs` proxy), and with {MissingImmutableArgs} when the immutable arguments
+     * are malformed.
```
arr00:
```suggestion
     * are not at least 28 bytes.
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,69 @@
...
+ * tracked by an {IAccessManager}.
+ *
+ * Instead of holding its own key material, this signer is bound to a single `roleId` and accepts a
+ * signature only when it was produced by an address that currently holds that role in the associated
+ * {accessManager} with 0 delay. This lets a role behave as a shared signer: membership can be granted
+ * or revoked through the access manager without redeploying or reconfiguring the signer.
+ *
+ * How the {accessManager} and `roleId` are resolved is left to implementations (see {accessManager} and
+ * {roleId}). {RoleAccount} decodes both from the immutable arguments of a `Clones.cloneWithImmutableArgs`
+ * proxy (see {RoleAccountFactory}, which deploys one clone per (access manager, role) pair).
```
arr00:
```suggestion
 * How the {accessManager} and {roleId} are resolved is left to implementations.
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,69 @@
...
+ */
+abstract contract SignerRole is AbstractSigner {
+    /**
+     * @dev Returns the {IAccessManager} whose role membership authorizes signatures for this signer.
+     *
+     * Implementations are responsible for defining how the access manager is resolved (see {RoleAccount},
+     * which decodes it from the clone's immutable arguments).
```
arr00:
```suggestion
     * Implementations are responsible for defining how the access manager is resolved.
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,69 @@
...
+     */
+    function accessManager() public view virtual returns (IAccessManager);
+
+    /**
+     * @dev Returns the role id this signer is bound to. Members of this role in the {accessManager}
+     * are authorized to produce signatures on behalf of this signer.
+     *
+     * Implementations are responsible for defining how the role id is resolved (see {RoleAccount},
+     * which decodes it from the clone's immutable arguments).
```
arr00:
```suggestion
     * Implementations are responsible for defining how the role id is resolved.
```

arr00 (dismissed):

contracts/account/RoleAccount.sol
arr00: Prefer to keep stricter unless we have an explicit usecase requiring more clone args. Given this is community it can change later no problem.

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+
+import {IAccessManager} from "@openzeppelin/contracts/access/manager/IAccessManager.sol";
+import {ERC7821} from "@openzeppelin/contracts/account/extensions/draft-ERC7821.sol";
+import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
+import {ERC7739} from "@openzeppelin/contracts/utils/cryptography/signers/draft-ERC7739.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
+import {SignerRole} from "../utils/cryptography/signers/SignerRole.sol";
+
+/**
+ * @dev On-chain account that represents a role of an {IAccessManager}.
```
gonzaotc: "represent" sounds abstract to me, I would frame "manage" or "own" instead
```suggestion
 * @dev On-chain account managed by the members of a role of an {IAccessManager}.
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+ * role) pair, by {RoleAccountFactory}. The access manager and role id are encoded in the clone's immutable
+ * arguments, so calling the getters on the implementation directly (outside a clone) reverts.
+ */
+contract RoleAccount is ERC7821, ERC7739, SignerRole {
+    address private immutable _self = address(this);
+
+    error DirectCallNotAllowed();
+    error MissingImmutableArgs();
+
+    constructor() EIP712("RoleAccount", "1.0.0") {}
```
gonzaotc:
```suggestion
    constructor() EIP712("RoleAccount", "1") {}
```

nitpick, we use "1" as the EIP712 version everywhere in community and vanilla

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+ * it can produce ERC-1271 signatures and execute batched calls, and its authority follows role
+ * membership as it is granted or revoked through the access manager.
+ *
+ * The account address can be computed off-chain (or on-chain via {getRoleAccount}) before deployment,
+ * so it can be used as an authorization target or funded ahead of time. {deployRoleAccount} materializes
+ * the clone at that address when needed.
+ *
+ * WARNING: A role account grants control to *every current member* of its role. For the special
+ * `PUBLIC_ROLE` (`type(uint64).max`), which every address belongs to, this means the account is
+ * controllable by anyone.
```
gonzaotc: Move this warning into the SignerRole instead?. I think it matters more there

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+            address(_template),
+            abi.encodePacked(accessManager, roleId),
+            bytes32(0)
+        );
+        emit RoleAccountDeployed(accessManager, roleId, roleAccount);
+
+        return roleAccount;
+    }
+
+    /// @dev Called once during construction to get the template used for deploying role accounts.
```
gonzaotc:
```suggestion
    /// @dev Called once during construction to get the template used for deploying role accounts. Can be overrided to use a different implementation template
```

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+ *
+ * NOTE: {deployRoleAccount} is permissionless. Because the deployment is deterministic and behaviorally
+ * fixed, this is harmless (front-running it only produces the same account).
+ */
+contract RoleAccountFactory {
+    /**
+     * @dev The template used for deploying each role account. This is set in the constructor and cannot
+     * be changed.
+     */
+    address private immutable _template;
```
gonzaotc: Add a public getter for the template?

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+
+    constructor() {
+        _template = _deployTemplate();
+    }
+
+    /**
+     * @dev Returns the deterministic address of the {RoleAccount} for `roleId` on `accessManager`, whether
+     * or not it has already been deployed.
+     */
+    function getRoleAccount(address accessManager, uint64 roleId) public view virtual returns (address) {
```
gonzaotc, line 41:
```suggestion
    function getRoleAccount(IAccessManager accessManager, uint64 roleId) public view virtual returns (address) {
```

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+                abi.encodePacked(accessManager, roleId),
+                bytes32(0)
+            );
+    }
+
+    /**
+     * @dev Deploys the {RoleAccount} clone for `roleId` on `accessManager` at its deterministic address and
+     * returns it. Reverts if the account has already been deployed.
+     */
+    function deployRoleAccount(address accessManager, uint64 roleId) public virtual returns (address) {
```
gonzaotc, line 59:
```suggestion
    function deployRoleAccount(IAccessManager accessManager, uint64 roleId) public virtual returns (address) {
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @dev Returns the role id this signer is bound to, decoded from the clone's immutable arguments.
+     *
+     * Reverts with {DirectCallNotAllowed} when called on the implementation directly (i.e. not through a
+     * `Clones.cloneWithImmutableArgs` proxy), and with {MissingImmutableArgs} when the immutable arguments
+     * are malformed.
+     */
+    function roleId() public view virtual override returns (uint64 roleId_) {
+        require(_self != address(this), DirectCallNotAllowed());
```
gonzaotc: > require(_self != address(this), DirectCallNotAllowed()); 

Can be moved into `_fetchArgs` to deduplicate it from `accessManager()`

gonzaotc: > * Reverts with {DirectCallNotAllowed} when called on the implementation directly (i.e. not through a `Clones.cloneWithImmutableArgs` proxy) 

should also be moved along

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,73 @@
...
+        address roleAccount = Clones.cloneDeterministicWithImmutableArgs(
+            address(_template),
```
gonzaotc:
```suggestion
            _template,
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,65 @@
...
+        return hasRole && executionDelay == 0;
+    }
+
+    /**
+     * @dev See {AbstractSigner-_rawSignatureValidation}.
+     *
+     * The `signature` is expected to be the concatenation `[20-byte signer address][inner signature]`.
+     * The leading 20 bytes identify the account that produced the inner signature. Validation succeeds
+     * only when the inner signature is valid for `hash` (verified through {SignatureChecker}, so both
+     * EOAs and ERC-1271 smart contract signers are supported) AND that signer currently holds {roleId}.
```
gonzaotc:
```suggestion
     * EOAs and ERC-1271 smart contract signers are supported) AND that signer is an unrestricted
     * {roleId} member (see {_isUnrestrictedMember}).
```

contracts/account/README.adoc
```diff
@@ -4,6 +4,8 @@ NOTE: This document is better viewed at https://docs.openzeppelin.com/community-
 
 This directory includes contracts to build accounts for ERC-4337. These include:
```
gonzaotc: This document states:

> This directory includes contracts to build accounts for ERC-4337. 

However  the `RoleAccount` being shipped here isn't compatible with ERC-4337 as it is.

In [this comment](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248#discussion_r3691931659) I was suggesting that the RoleAccount should be a construction of our [`Account`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/account/Account.sol) plus the introduced `SignerRole`, ideally through the wizard. 

If we want anyway to include RoleAccount and RoleAccountFactory here, I think inheriting `Account` to bring ERC-4337 compatibility would be a good idea, wdyt @arr00

gonzaotc: Otherwise if we have a strong reason to avoid ERC-4337 (bytecode size?, 4337 not mattering on ZAMA?), we can sligthly update the doc framing about the contracts included here

contracts/account/RoleAccountFactory.sol
arr00: I strongly prefer address here

contracts/account/RoleAccountFactory.sol
arr00: I strongly prefer address here. Integrators don't want to be required to import `IAccessManager` and it really gives no assurances.

contracts/account/RoleAccountFactory.sol
arr00: why not

contracts/account/RoleAccountFactory.sol
arr00: applied slightly modified

contracts/account/README.adoc
arr00: We should update the docs to be inclusive. Not the other was around.

contracts/account/RoleAccountFactory.sol
arr00: done

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,70 @@
...
+ * `PUBLIC_ROLE` (`type(uint64).max`), which every address belongs to, this means the account is
+ * controllable by anyone.
+ */
+abstract contract SignerRole is AbstractSigner {
+    /**
+     * @dev Returns the {IAccessManager} whose role membership authorizes signatures for this signer.
+     *
+     * Implementations are responsible for defining how the access manager is resolved.
+     */
+    function accessManager() public view virtual returns (IAccessManager);
```
ernestognw: I feel something is off with the `SignerRole` naming. In principle, I understand that `SignerRole` is the right generalization for "a signer whose authorization depends on roles", so it came up weird to me that we expose an `accessManager` function, so I thought it'd be better to have an `authority` function instead (see `IAuthority`). However, that doesn't work because we're calling `hasRole`, which is a function specific to the AccessManager; therefore, it sounds like `SignerRole` is not good because it sounds ambiguous and (I believe) attempts to generalize the signer when it's specific to the `AccessManager` design.

Sorry if this sounds confusing, but essentially I'd proposed to call this contract `SignerAccessManaged` instead. That's straight to the point imo

ernestognw: That said, this function shouldn't be virtual imo. It should be:


```suggestion
abstract contract SignerRole is AbstractSigner {
    IAccessManager private immutable _accessManager;
    
    constructor(IAccessManager accessManager) {
      _accessManager = accessManager;
    }

    /**
     * @dev Returns the {IAccessManager} whose role membership authorizes signatures for this signer.
     *
     * Implementations are responsible for defining how the access manager is resolved.
     */
    function accessManager() public view virtual returns (IAccessManager) {
      return _accessManager;
    }
```

I understand that we left it virtual so that the `RoleAccount` receives it through immutable arguments when deployed as clone, and this variant would force a constructor. However, I believe passing `address(0)` to the constructor of `RoleAccount` is better than leaving this function unimplemented.

wdyt?

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @dev Returns the role id this signer is bound to. Members of this role in the {accessManager}
+     * are authorized to produce signatures on behalf of this signer.
+     *
+     * Implementations are responsible for defining how the role id is resolved.
+     */
+    function roleId() public view virtual returns (uint64);
```
ernestognw: Similar to my previous comment. It'd be better if this function is implemented by default.

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @dev Returns whether `account` currently holds {roleId} in the {accessManager} and has no execution delay.
+     * This signer does not allow roles with execution delays to interact with it.
+     */
+    function _isUnrestrictedMember(address account) internal view virtual returns (bool) {
```
ernestognw: It would be preferred to avoid reading in "negative"

```suggestion
    function _isAuthorizedMember(address account) internal view virtual returns (bool) {
```

contracts/account/RoleAccountFactory.sol
ernestognw: Why `template`, btw? We refer to these addresses as `implementation` unuformly.

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+     * be changed.
+     */
+    address private immutable _template;
+
+    /// @dev Emitted when a {RoleAccount} is deployed for a role on an access manager.
+    event RoleAccountDeployed(address indexed accessManager, uint64 indexed roleId, address account);
+
+    constructor() {
+        _template = _deployTemplate();
+    }
```
ernestognw, line 35: Instead of a virtual function, I believe it would be better to receive an argument:

```suggestion
    constructor(address implementation) {
        _template = implementation == address(0) ? address(new RoleAccount()) : implementation;
    }
```

imo, customization should be done by an argument instead of a virtual function, because it's only done at deployment

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+     */
+    function deployRoleAccount(address accessManager, uint64 roleId) public virtual returns (address) {
+        address roleAccount = Clones.cloneDeterministicWithImmutableArgs(
+            _template,
+            abi.encodePacked(accessManager, roleId),
+            bytes32(0)
+        );
+        emit RoleAccountDeployed(accessManager, roleId, roleAccount);
+
+        return roleAccount;
```
ernestognw, line 67: I'd suggest moving this to an internal function that receives the `salt` as an argument, then, having 2 public functions; one that doesn't receive the `salt` and passes `bytes32(0)` to the internal variant by default and another that let's passing a custom `salt`

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+     * @dev Returns the deterministic address of the {RoleAccount} for `roleId` on `accessManager`, whether
+     * or not it has already been deployed.
+     */
+    function getRoleAccount(address accessManager, uint64 roleId) public view virtual returns (address) {
+        return
+            Clones.predictDeterministicAddressWithImmutableArgs(
+                _template,
+                abi.encodePacked(accessManager, roleId),
+                bytes32(0)
+            );
```
ernestognw, line 47: Same about having an internal function and then 2 public getters

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * @dev Deploys the {RoleAccount} clone for `roleId` on `accessManager` at its deterministic address and
+     * returns it. Reverts if the account has already been deployed.
+     */
+    function deployRoleAccount(address accessManager, uint64 roleId) public virtual returns (address) {
```
ernestognw, line 59: The `PUBLIC_ROLE` should be restricted.

```suggestion
    function deployRoleAccount(address accessManager, uint64 roleId) public virtual returns (address) {
      require(roleId != type(uint64).max, ...);
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+     * Reverts with {DirectCallNotAllowed} when called on the implementation directly (i.e. not through a
+     * `Clones.cloneWithImmutableArgs` proxy), and with {InvalidImmutableArgs} when the immutable arguments
+     * are not 28 bytes long.
+     */
+    function roleId() public view virtual override returns (uint64 roleId_) {
+        (, roleId_) = _fetchArgs();
+    }
+
+    /// @dev Decodes the access manager and role id packed in the clone's immutable arguments.
+    function _fetchArgs() private view returns (IAccessManager, uint64) {
```
ernestognw: Any reason to make this function `private`? A derived version wouldn't be able to extend the arguments


```suggestion
    function _fetchArgs() internal virtual view returns (IAccessManager, uint64) {
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+        require(_self != address(this), DirectCallNotAllowed());
+
+        bytes memory cloneArgs = Clones.fetchCloneArgs(address(this));
+        require(cloneArgs.length == 28, InvalidImmutableArgs());
```
ernestognw: Following the `internal virtual` comment, this might be better as:


```suggestion
        require(cloneArgs.length >= 28, InvalidImmutableArgs());
```

contracts/account/RoleAccount.sol
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * These accounts are intended to be deployed as `Clones.cloneWithImmutableArgs`, once per (access manager,
+ * role) pair, by {RoleAccountFactory}. The access manager and role id are encoded in the clone's immutable
+ * arguments, so calling the getters on the implementation directly (outside a clone) reverts.
+ */
+contract RoleAccount is ERC7821, ERC7739, SignerRole {
+    address private immutable _self = address(this);
+
+    error DirectCallNotAllowed();
+    error InvalidImmutableArgs();
```
ernestognw: Remember to namespace custom errors with the contract name so that we don't get "already declared identifier" errors. These 2 sound pretty generic and easy to repeat accidentally

```suggestion
    error RoleAccountDirectCallNotAllowed();
    error RoleAccountInvalidImmutableArgs();
```

contracts/utils/cryptography/signers/SignerRole.sol
```diff
@@ -0,0 +1,70 @@
...
+     */
+    function _rawSignatureValidation(
+        bytes32 hash,
+        bytes calldata signature
+    ) internal view virtual override returns (bool) {
+        address signer = address(bytes20(signature));
+        return
+            signature.length >= 20 &&
+            SignatureChecker.isValidSignatureNowCalldata(signer, hash, signature[20:]) &&
+            _isUnrestrictedMember(signer);
```
ernestognw: Not sure if it's better, but I feel uncomfortable casting the signer before checking the length matches

```suggestion
        if (signature.length < 20) return false;
        address signer = address(bytes20(signature));
        return
            SignatureChecker.isValidSignatureNowCalldata(signer, hash, signature[20:]) &&
            _isUnrestrictedMember(signer);
```

contracts/account/RoleAccountFactory.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    constructor() {
+        _template = _deployTemplate();
+    }
+
+    /**
+     * @dev Returns the deterministic address of the {RoleAccount} for `roleId` on `accessManager`, whether
+     * or not it has already been deployed.
+     */
+    function getRoleAccount(address accessManager, uint64 roleId) public view virtual returns (address) {
```
ernestognw, line 41: Note this function doesn't tell whether the account is deployed or not. Integrators should check `address.code.length > 0` later on. Either we document or add a `isRoleAccountDeployed(address, uint64) view returns (bool)` function

contracts/account/RoleAccountFactory.sol
arr00: The internal function will be stripped from the bytecode anyways since its only used at deployment. I think overriding the internal function provides more transparency and is less confusing for someone simply using the default version.

contracts/account/RoleAccountFactory.sol
arr00: We have a warning for this https://github.com/OpenZeppelin/openzeppelin-community-contracts/blob/f1b22916241af9bf33e34563f2bd8a28e854936c/contracts/utils/cryptography/signers/SignerRole.sol#L20-L22

Don't see why we need to restrict creation.

contracts/account/RoleAccount.sol
arr00: This being private doesn't restrict any args usage elsewhere, just forces our 28 byte encoding. What does restrict derivation here is https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/248#discussion_r3767487941

contracts/account/RoleAccountFactory.sol
arr00: I don't see this as a feature that is desired in the usecase that inspired this contract. If anything it would be a bug. We'd want a canonical account per role and that's all.

contracts/account/RoleAccountFactory.sol
arr00: It does have the clause
> whether or not it has already been deployed.

contracts/utils/cryptography/signers/SignerRole.sol
arr00: will add

contracts/account/RoleAccount.sol
arr00: name vars

ernestognw (approved):
Thanks @arr00 and everyone who reviewed

gonzaotc (approved):

arr00:
Reapplied via #267 (commit 07fb6f252e2b25a40825d03d9dc1289799bbf1db) to fix commit metadata.

PR #249, opened 2026-07-16
Fix ERC20uRWA canTransfer/forcedTransfer restriction checks

### LINKED ISSUE #244: ERC20uRWA: potential ERC-7943 compliance issues and self-forced-transfer edge case

mdimran29:
## Summary

Fixes #244 (points 3 & 4).

- Add virtual `canSend(address)` and `canReceive(address)`, both defaulting to `canTransact()`, so send-side and receive-side restrictions can be overridden independently.
- `canTransfer()`: remove the `amount <= available(from)` balance check (a `canTransfer` availability query shouldn't also encode balance sufficiency) and check `canSend(from) && canReceive(to)` instead of `canTransact(from) && canTransact(to)`.
- `forcedTransfer()`: gate the recipient with `canReceive(to)` instead of `canTransact(to)`, consistent with `canTransfer`.

No changes to `setFrozenTokens`, the `forcedTransfer` `from == to` handling, or any other files.

## Test plan

- [x] `forge build` passes
- [x] `forge test` passes (41/41, including 11 new tests in `test/token/ERC20/extensions/ERC20uRWA.t.sol`)
- New tests cover:
  - `canTransfer` returns `true` even when `amount` exceeds available/frozen balance
  - `canTransfer` returns `false` when `canSend` or `canReceive` is false
  - `forcedTransfer` reverts when `canReceive` is false
  - Accounts that can receive but not send, and vice versa

### DIFF
--- a/contracts/token/ERC20/extensions/ERC20uRWA.sol
+++ b/contracts/token/ERC20/extensions/ERC20uRWA.sol
@@ -30,13 +30,31 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
     }
 
     /**
+     * @dev Returns whether `account` is allowed to send tokens. Defaults to {canTransact}.
+     *
+     * Override to implement sender-specific restrictions distinct from {canReceive}.
+     */
+    function canSend(address account) public view virtual returns (bool) {
+        return canTransact(account);
+    }
+
+    /**
+     * @dev Returns whether `account` is allowed to receive tokens. Defaults to {canTransact}.
+     *
+     * Override to implement recipient-specific restrictions distinct from {canSend}.
+     */
+    function canReceive(address account) public view virtual returns (bool) {
+        return canTransact(account);
+    }
+
+    /**
      * @dev See {IERC7943Fungible-canTransfer}.
      *
      * CAUTION: This function is only meant for external use. Overriding it will not apply the new checks to
      * the internal {_update} function. Consider overriding {_update} accordingly to keep both functions in sync.
      */
-    function canTransfer(address from, address to, uint256 amount) external view virtual returns (bool) {
-        return (amount <= available(from) && canTransact(from) && canTransact(to));
+    function canTransfer(address from, address to, uint256 /* amount */) external view virtual returns (bool) {
+        return (canSend(from) && canReceive(to));
     }
 
     /// @inheritdoc IERC7943Fungible
@@ -70,7 +88,7 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
      */
     function forcedTransfer(address from, address to, uint256 amount) public virtual returns (bool result) {
         _checkEnforcer(from, to, amount);
-        require(canTransact(to), ERC7943CannotTransact(to));
+        require(canReceive(to), ERC7943CannotTransact(to));
 
         // Update frozen balance if needed. ERC-7943 requires that balance is unfrozen first and then send the tokens.
         uint256 currentFrozen = frozen(from);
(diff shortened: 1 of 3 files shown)


xaler5:
Thanks for this, there are three deviations from the final EIP-7943 spec that should be addressed:

### 1. `canTransfer` must validate the unfrozen amount

The EIP explicitly states:

> \`canTransfer\` **MUST validate that the amount being transferred doesn't exceed the unfrozen amount** (which is the difference between the current balance and the frozen balance).

The current PR removes the `amount <= available(from)` check. While the intent (separating availability from balance-sufficiency) is reasonable, the EIP specifically carves out the **unfrozen-amount** check as a permissioned compliance rule that belongs in `canTransfer`, it's not a base-standard balance check, it's a compliance gate on how much of your own tokens you're allowed to move. Restoring it keeps `canTransfer` spec-compliant while `canSend`/`canReceive` handle the account-level eligibility.

### 2. Wrong error type for `canReceive` in `forcedTransfer`

The require was changed to:
```solidity
require(canReceive(to), ERC7943CannotTransact(to));
```

But `ERC7943CannotTransact` is not defined in the EIP interface anymore. The spec defines `ERC7943CannotReceive(address account)`. The error should be:
```solidity
require(canReceive(to), ERC7943CannotReceive(to));
```

### 3. `_update` doesn't enforce `canSend`/`canReceive` at transfer time

The EIP states:

> Public transfers (transfer, transferFrom, safeTransferFrom, etc.) **MUST NOT succeed** in cases where canTransfer would return false, or where canSend would return false for the from address, or canReceive would return false for the to address.

`canSend` and `canReceive` are introduced as virtual functions but never enforced during `_update`. The `ERC20Restricted._update()` only checks `canTransact`, not the new split checks. This means a blocked sender could still push tokens through (and the receiver check is also missed). Either override `_update` in `ERC20uRWA` to add `canSend(from) && canReceive(to)`, or have `canTransact` delegate to both.

---

xaler5:
Thanks @mdimran29 — the canSend/canReceive split is now part of #254, with your commit cherry-picked as-is (authorship preserved). On top of it, #254 addresses the three points from my review: it restores the unfrozen-balance validation in canTransfer (returning false only when the amount is within the total balance, so plain balance insufficiency stays a base-standard concern), aligns IERC7943Fungible with the final spec (ERC7943CannotSend/ERC7943CannotReceive, final 0x3edbb4c4 interface id), and enforces canSend/canReceive in _update with a transient-flag bypass for forcedTransfer. Your Foundry tests are included with two expectations updated to the final-spec semantics.

mdimran29:
Thanks for the thorough review, @xaler5 — appreciate you catching the spec deviations and doing the follow-up work in #254. Agreed on all three points:

Restoring the unfrozen-balance check in canTransfer makes sense — that's a compliance-specific rule the EIP calls out explicitly, not a generic balance check.
Good catch on ERC7943CannotReceive vs ERC7943CannotTransact — I was working off an earlier version of the interface.
You're right that _update needs to enforce canSend/canReceive directly — otherwise the view function and the actual transfer path can disagree.
Happy to have #254 supersede this PR. I'll close this one out — thanks for carrying it forward.

PR #251, opened 2026-07-17
Make ERC7786OpenBridge tolerant to failing gateways

Amxx:
Fixes https://audits.openzeppelin.com/openzeppelin-solidity/project/openzeppelin-community-contracts/e2bb51d8-df7a-4d26-934a-be65c637d4e5/issue/any-single-gateway-revert-blocks-sendmessage-for-all-gateways

## Summary

Makes `ERC7786OpenBridge` resilient to individual gateway failures on the source side, so a single misbehaving gateway can no longer block a message that can still satisfy the N-of-M threshold.

## Changes

- **`sendMessage`**: each `gateway.sendMessage` is now wrapped in `try/catch`. A reverting gateway no longer aborts the whole send; instead the loop continues with the others. After the loop, the call requires at least `getThreshold()` successful sends (`ERC7786OpenBridgeInsufficientGateways` otherwise).
- **`ERC7786OpenBridgeSendMessageFailed(address indexed gateway)`**: new event emitted for each gateway whose `sendMessage` reverts, for observability.
- **`_addGateway`**: rejects non-contract addresses (`ERC7786OpenBridgeGatewayNotAContract`), covering both the constructor and `addGateway`.
- The gateway-posting locals are scoped in a block to avoid a stack-too-deep (the project compiles without `viaIR`).

## Tests

- New `gateway resilience` test: two synchronous gateways, threshold 1, one gateway set to revert on send — asserts the message is still delivered through the healthy gateway and that `ERC7786OpenBridgeSendMessageFailed` is emitted for the failing one.
- Updated the `bad return value` / `EOA` recipient cases: under the synchronous mock, the destination revert is now caught on the send side, so they revert with `ERC7786OpenBridgeInsufficientGateways`.
- `ERC7786GatewayMock` gains a `setRevertOnSent` toggle to drive the resilience test.

All tests pass; prettier / eslint / solhint clean.

## Summary by CodeRabbit

* **New Features**
  * Cross-chain messages can now succeed when some gateways fail, provided the configured delivery threshold is met.
  * Failed gateway attempts are reported through a dedicated event.
* **Bug Fixes**
  * Gateway registration now rejects addresses that are not deployed contracts.
  * A single unresponsive or reverting gateway no longer automatically prevents message delivery through other available gateways.


contracts/crosschain/ERC7786OpenBridge.sol
```diff
@@ -107,22 +110,33 @@ contract ERC7786OpenBridge is IERC7786GatewaySource, IERC7786Recipient, Ownable,
...
+                    // if ID, track it
+                    if (id != bytes32(0)) {
+                        outbox[i] = Outbox(gateway, id);
+                        needsId = true;
+                    }
+                    ++sent;
+                } catch {
+                    // if one gateway fails, we still want to send the message through the others
+                    emit ERC7786OpenBridgeSendMessageFailed(gateway);
+                }
```
ernestognw, line 133: Shall we include the revert reason in the event?

```suggestion
                } catch (bytes memory reason) {
                    // if one gateway fails, we still want to send the message through the others
                    emit ERC7786OpenBridgeSendMessageFailed(gateway, reason);
                }
```

Otherwise I suspect it may be better to use `LowLevelCall` to avoid loading the revert reason to memory in case the call fails.

Non-blocking

contracts/crosschain/ERC7786OpenBridge.sol
```diff
@@ -107,22 +110,33 @@ contract ERC7786OpenBridge is IERC7786GatewaySource, IERC7786Recipient, Ownable,
...
             }
-        }
+            require(sent >= getThreshold(), ERC7786OpenBridgeInsufficientGateways());
```
ernestognw, line 135: Do we strictly need this require and `sent` count? I believe calls to `sendMessage` could be cumulative, meaning that for a threshold of 3, one could call `sendMessage` 3 times, each time adding and removing a different gateway

contracts/crosschain/ERC7786OpenBridge.sol
```diff
@@ -288,6 +302,7 @@ contract ERC7786OpenBridge is IERC7786GatewaySource, IERC7786Recipient, Ownable,
     // ================================================== Internal ===================================================
 
     function _addGateway(address gateway) internal virtual {
+        require(gateway.code.length > 0, ERC7786OpenBridgeGatewayNotAContract(gateway));
```
ernestognw, line 306: Shall we be worried about deterministic addresses? Let's say a gateway is not deployed yet but will be. I believe in a deployment script, the order may matter.

Non-blocking either

contracts/crosschain/ERC7786OpenBridge.sol
Amxx: I'm not sure how low level call would be better here. I don't expect try catch allocated the memory of the call doesn't fail.

I don't have a strong opinion about weither the event should include the revert reason or not. It would greatly increase the gas cost, I'm not sure it that is worth it (AFAIK, revert reason really aren't used that much in production and are mostly a testing thing)

contracts/crosschain/ERC7786OpenBridge.sol
Amxx: I would not worry about that here

People should get the factory deployed before they add it. Note that it can be added after construction

contracts/crosschain/ERC7786OpenBridge.sol
Amxx: If you call send message 3 times, you'll get 3 "packets" with different nonces (see line 110). This means the payload that goes trough the underlying gateways will be different everytime. Aggregation on the receiving chain will see them as independent entities, and you will not get consensus.

There is however an argument that could be made about this check: we have no guarantee that the threshold seen here (sending chain) matches the one on the receiving chain. This check assumes some consistency in the configurations

contracts/crosschain/ERC7786OpenBridge.sol
ernestognw: Right, I agree the revert reason is not needed for the event. I was thinking that any Solidity call copies the revert reason to memory afaiu, so I was proposing LowLevelCall to avoid such case.

I believe it's fine for now

ernestognw (approved):
CodeCov seems to be buggy. I'm merging

PR #254, opened 2026-07-20
fix(ERC20uRWA): align with the final EIP-7943 specification

### LINKED ISSUE #244: ERC20uRWA: potential ERC-7943 compliance issues and self-forced-transfer edge case

xaler5:
## Align ERC20uRWA with the final EIP-7943 specification

Closes #244. Supersedes #246 and #249 and the job is completed here so the extension is fully compliant with the [final EIP-7943 spec](https://eips.ethereum.org/EIPS/eip-7943) in a single reviewable unit.

### Interface (`IERC7943Fungible`)

The interface in this repo predated finalization. This PR aligns it with the final spec:

- `canTransact(address)` → `canSend(address)` + `canReceive(address)`.
- `ERC7943CannotTransact` → `ERC7943CannotSend` + `ERC7943CannotReceive`.
- `type(IERC7943Fungible).interfaceId` now equals the final `0x3edbb4c4`, satisfying the ERC-165 MUST (asserted in tests).

### `ERC20uRWA`

- **`setFrozenTokens` no longer caps to balance**: the spec requires *"MUST allow freezing more assets than those held"* (future balances withholding). (No `Frozen`-event change was needed: `ERC20Freezable._setFrozen` already emits it before the base `Transfer` event.)
- **`forcedTransfer(from, from, amount)` reverts**: a self-directed forced transfer moves no tokens but lowered the frozen balance, acting as an unauthorized unfreeze that bypassed the freezer role (from #246).
- **`canSend` / `canReceive` split** (from #249), defaulting to `canTransact`, so send-side and receive-side eligibility can be overridden independently.
- **`canTransfer` returns false only for permissioned rules**: an `amount` exceeding the unfrozen balance *while within the total balance*, or failing `canSend(from)` / `canReceive(to)`. Plain balance insufficiency returns true, per *"MUST NOT return false based on token-specific, non-permissioned validations"*, while still honoring *"MUST validate that the amount being transferred doesn't exceed the unfrozen amount"*.
- **`_update` enforces `canSend` / `canReceive`**, so public transfers, minting, and burning cannot succeed where `canTransfer` would return false, even when those checks are overridden beyond `canTransact`.
- **`forcedTransfer` gates the recipient with `canReceive`** (reverting with `ERC7943CannotReceive`) and bypasses sender-side checks through a transient-storage flag instead of temporarily rewriting the sender's restriction, the previous mechanism didn't bypass overridden `canSend` and emitted spurious `UserRestrictionsUpdated` events.

### Breaking changes

- `IERC7943Fungible` ABI: `canTransact` and `ERC7943CannotTransact` are gone; the interface id changed to the final `0x3edbb4c4`.
- Transfers involving restricted accounts now revert with `ERC7943CannotSend` / `ERC7943CannotReceive` instead of `ERC20UserRestricted` (the spec prioritizes error specificity).

### Test results

- Hardhat: 51 passing in the `ERC20uRWA` suite (full repo suite: 1163 passing), including the final interface id, freeze-above-balance withholding, single `Frozen` event ordered before `Transfer` on forced unfreeze, self-forced-transfer rejection, one-way send/receive restrictions, and `_update` enforcement of overridden checks.
- Foundry: 11 passing in `test/token/ERC20/extensions/ERC20uRWA.t.sol` (from #249, expectations updated to final-spec `canTransfer` semantics).

### DIFF
--- a/contracts/interfaces/IERC7943.sol
+++ b/contracts/interfaces/IERC7943.sol
@@ -16,9 +16,13 @@ interface IERC7943Fungible is IERC165 {
     /// @param amount The amount of tokens frozen after the change.
     event Frozen(address indexed account, uint256 amount);
 
-    /// @notice Error reverted when an account is not allowed to transact.
-    /// @param account The address of the account which is not allowed for transfers.
-    error ERC7943CannotTransact(address account);
+    /// @notice Error reverted when an account is not allowed to send tokens.
+    /// @param account The address of the account which is not allowed to send.
+    error ERC7943CannotSend(address account);
+
+    /// @notice Error reverted when an account is not allowed to receive tokens.
+    /// @param account The address of the account which is not allowed to receive.
+    error ERC7943CannotReceive(address account);
 
     /// @notice Error reverted when a transfer is not allowed according to internal rules.
     /// @param from The address from which tokens are being sent.
@@ -48,11 +52,17 @@ interface IERC7943Fungible is IERC165 {
     /// @return result True if the freezing executed correctly, false otherwise.
     function setFrozenTokens(address account, uint256 amount) external returns (bool result);
 
-    /// @notice Checks if a specific account is allowed to transact according to token rules.
+    /// @notice Checks if a specific account is allowed to send tokens according to token rules.
     /// @dev This is often used for allowlist/KYC/KYB/AML checks.
     /// @param account The address to check.
-    /// @return allowed True if the account is allowed, false otherwise.
-    function canTransact(address account) external view returns (bool allowed);
+    /// @return allowed True if the account is allowed to send, false otherwise.
+    function canSend(address account) external view returns (bool allowed);
+
+    /// @notice Checks if a specific account is allowed to receive tokens according to token rules.
+    /// @dev This is often used for allowlist/KYC/KYB/AML checks.
+    /// @param account The address to check.
+    /// @return allowed True if the account is allowed to receive, false otherwise.
+    function canReceive(address account) external view returns (bool allowed);
 
     /// @notice Checks the frozen status/amount.
     /// @param account The address of the account.
--- a/contracts/interfaces/IERC7943.sol
+++ b/contracts/interfaces/IERC7943.sol
@@ -93,9 +93,13 @@ interface IERC7943NonFungible is IERC165 {
     /// @param frozenStatus Whether `tokenId` has been frozen or unfrozen.
     event Frozen(address indexed account, uint256 indexed tokenId, bool indexed frozenStatus);
 
-    /// @notice Error reverted when an account is not allowed to transact.
-    /// @param account The address of the account which is not allowed for transfers.
-    error ERC7943CannotTransact(address account);
+    /// @notice Error reverted when an account is not allowed to send tokens.
+    /// @param account The address of the account which is not allowed to send.
+    error ERC7943CannotSend(address account);
+
+    /// @notice Error reverted when an account is not allowed to receive tokens.
+    /// @param account The address of the account which is not allowed to receive.
+    error ERC7943CannotReceive(address account);
 
     /// @notice Error reverted when a transfer is not allowed according to internal rules.
     /// @param from The address from which tokens are being sent.
@@ -125,11 +129,17 @@ interface IERC7943NonFungible is IERC165 {
     /// @return result True if the freezing executed correctly, false otherwise.
     function setFrozenTokens(address account, uint256 tokenId, bool frozenStatus) external returns (bool result);
 
-    /// @notice Checks if a specific account is allowed to transact according to token rules.
+    /// @notice Checks if a specific account is allowed to send tokens according to token rules.
     /// @dev This is often used for allowlist/KYC/KYB/AML checks.
     /// @param account The address to check.
-    /// @return allowed True if the account is allowed, false otherwise.
-    function canTransact(address account) external view returns (bool allowed);
+    /// @return allowed True if the account is allowed to send, false otherwise.
+    function canSend(address account) external view returns (bool allowed);
+
+    /// @notice Checks if a specific account is allowed to receive tokens according to token rules.
+    /// @dev This is often used for allowlist/KYC/KYB/AML checks.
+    /// @param account The address to check.
+    /// @return allowed True if the account is allowed to receive, false otherwise.
+    function canReceive(address account) external view returns (bool allowed);
 
     /// @notice Checks the frozen status of a specific `tokenId`.
     /// @dev It could return true even if account does not hold the token.
@@ -162,9 +172,13 @@ interface IERC7943MultiToken is IERC165 {
     /// @param amount The amount of tokens frozen after the change.
     event Frozen(address indexed account, uint256 indexed tokenId, uint256 amount);
 
-    /// @notice Error reverted when an account is not allowed to transact.
-    /// @param account The address of the account which is not allowed for transfers.
-    error ERC7943CannotTransact(address account);
+    /// @notice Error reverted when an account is not allowed to send tokens.
+    /// @param account The address of the account which is not allowed to send.
+    error ERC7943CannotSend(address account);
+
+    /// @notice Error reverted when an account is not allowed to receive tokens.
+    /// @param account The address of the account which is not allowed to receive.
+    error ERC7943CannotReceive(address account);
 
     /// @notice Error reverted when a transfer is not allowed according to internal rules.
     /// @param from The address from which tokens are being sent.
@@ -198,11 +212,17 @@ interface IERC7943MultiToken is IERC165 {
     /// @return result True if the freezing executed correctly, false otherwise.
     function setFrozenTokens(address account, uint256 tokenId, uint256 amount) external returns (bool result);
 
-    /// @notice Checks if a specific account is allowed to transact according to token rules.
+    /// @notice Checks if a specific account is allowed to send tokens according to token rules.
     /// @dev This is often used for allowlist/KYC/KYB/AML checks.
     /// @param account The address to check.
-    /// @return allowed True if the account is allowed, false otherwise.
-    function canTransact(address account) external view returns (bool allowed);
+    /// @return allowed True if the account is allowed to send, false otherwise.
+    function canSend(address account) external view returns (bool allowed);
+
+    /// @notice Checks if a specific account is allowed to receive tokens according to token rules.
+    /// @dev This is often used for allowlist/KYC/KYB/AML checks.
+    /// @param account The address to check.
+    /// @return allowed True if the account is allowed to receive, false otherwise.
+    function canReceive(address account) external view returns (bool allowed);
 
     /// @notice Checks the frozen status/amount of a specific `tokenId`.
     /// @dev It could return an amount higher than the account's balance.
--- a/contracts/interfaces/IERC7943.sol
+++ b/contracts/interfaces/IERC7943.sol
@@ -41,15 +41,15 @@ interface IERC7943Fungible is IERC165 {
     /// @param from The address from which `amount` is taken.
     /// @param to The address that receives `amount`.
     /// @param amount The amount to force transfer.
-    /// @return result True if the transfer executed correctly, false otherwise.
+    /// @return result True if the transfer executed correctly. Reverts on failure.
     function forcedTransfer(address from, address to, uint256 amount) external returns (bool result);
 
     /// @notice Changes the frozen status of `amount` tokens belonging to `account`.
-    /// This overwrites the current value, similar to an `approve` function.
-    /// @dev Requires specific authorization. Frozen tokens cannot be transferred by the account.
+    /// @dev Overwrites the current value, similar to an `approve` function.
+    /// Requires specific authorization. Frozen tokens cannot be transferred by the account.
     /// @param account The address of the account whose tokens are to be frozen.
-    /// @param amount The amount of tokens to freeze. It can be greater than account balance.
-    /// @return result True if the freezing executed correctly, false otherwise.
+    /// @param amount The amount of tokens to freeze. It can be greater than the account balance.
+    /// @return result True if the freezing executed correctly. Reverts on failure.
     function setFrozenTokens(address account, uint256 amount) external returns (bool result);
 
     /// @notice Checks if a specific account is allowed to send tokens according to token rules.
@@ -71,7 +71,7 @@ interface IERC7943Fungible is IERC165 {
     function getFrozenTokens(address account) external view returns (uint256 amount);
 
     /// @notice Checks if a transfer is currently possible according to token rules. It enforces validations on the frozen tokens.
-    /// @dev This can involve checks like allowlists, blocklists, transfer limits and other policy-defined restrictions.
+    /// @dev This can involve checks like allowlists, blocklists, transfer limits, and other policy-defined restrictions.
     /// @param from The address sending tokens.
     /// @param to The address receiving tokens.
     /// @param amount The amount being transferred.
@@ -104,7 +104,7 @@ interface IERC7943NonFungible is IERC165 {
     /// @notice Error reverted when a transfer is not allowed according to internal rules.
     /// @param from The address from which tokens are being sent.
     /// @param to The address to which tokens are being sent.
-    /// @param tokenId The id of the token being sent.
+    /// @param tokenId The ID of the token being sent.
     error ERC7943CannotTransfer(address from, address to, uint256 tokenId);
 
     /// @notice Error reverted when a transfer is attempted from `account` with a `tokenId` which has been previously frozen.
@@ -117,16 +117,16 @@ interface IERC7943NonFungible is IERC165 {
     /// @param from The address from which `tokenId` is taken.
     /// @param to The address that receives `tokenId`.
     /// @param tokenId The ID of the token being transferred.
-    /// @return result True if the transfer executed correctly, false otherwise.
+    /// @return result True if the transfer executed correctly. Reverts on failure.
     function forcedTransfer(address from, address to, uint256 tokenId) external returns (bool result);
 
     /// @notice Changes the frozen status of `tokenId` belonging to an `account`.
-    /// This overwrites the current value, similar to an `approve` function.
-    /// @dev Requires specific authorization. Frozen tokens cannot be transferred by the account.
+    /// @dev Overwrites the current value, similar to an `approve` function.
+    /// Requires specific authorization. Frozen tokens cannot be transferred by the account.
     /// @param account The address of the account whose tokens are to be frozen.
     /// @param tokenId The ID of the token to freeze.
     /// @param frozenStatus Whether `tokenId` is being frozen or not.
-    /// @return result True if the freezing executed correctly, false otherwise.
+    /// @return result True if the freezing executed correctly. Reverts on failure.
     function setFrozenTokens(address account, uint256 tokenId, bool frozenStatus) external returns (bool result);
 
     /// @notice Checks if a specific account is allowed to send tokens according to token rules.
@@ -149,7 +149,7 @@ interface IERC7943NonFungible is IERC165 {
     function getFrozenTokens(address account, uint256 tokenId) external view returns (bool frozenStatus);
 
     /// @notice Checks if a transfer is currently possible according to token rules. It enforces validations on the frozen tokens.
-    /// @dev This can involve checks like allowlists, blocklists, transfer limits and other policy-defined restrictions.
+    /// @dev This can involve checks like allowlists, blocklists, transfer limits, and other policy-defined restrictions.
     /// @param from The address sending tokens.
     /// @param to The address receiving tokens.
     /// @param tokenId The ID of the token being transferred.
@@ -183,7 +183,7 @@ interface IERC7943MultiToken is IERC165 {
     /// @notice Error reverted when a transfer is not allowed according to internal rules.
     /// @param from The address from which tokens are being sent.
     /// @param to The address to which tokens are being sent.
-    /// @param tokenId The id of the token being sent.
+    /// @param tokenId The ID of the token being sent.
     /// @param amount The amount sent.
     error ERC7943CannotTransfer(address from, address to, uint256 tokenId, uint256 amount);
 
@@ -200,16 +200,16 @@ interface IERC7943MultiToken is IERC165 {
     /// @param to The address that receives `amount`.
     /// @param tokenId The ID of the token being transferred.
     /// @param amount The amount to force transfer.
-    /// @return result True if the transfer executed correctly, false otherwise.
+    /// @return result True if the transfer executed correctly. Reverts on failure.
     function forcedTransfer(address from, address to, uint256 tokenId, uint256 amount) external returns (bool result);
 
     /// @notice Changes the frozen status of `amount` of `tokenId` tokens belonging to an `account`.
-    /// This overwrites the current value, similar to an `approve` function.
-    /// @dev Requires specific authorization. Frozen tokens cannot be transferred by the account.
+    /// @dev Overwrites the current value, similar to an `approve` function.
+    /// Requires specific authorization. Frozen tokens cannot be transferred by the account.
     /// @param account The address of the account whose tokens are to be frozen.
     /// @param tokenId The ID of the token to freeze.
-    /// @param amount The amount of tokens to freeze. It can be greater than account balance.
-    /// @return result True if the freezing executed correctly, false otherwise.
+    /// @param amount The amount of tokens to freeze. It can be greater than the account balance.
+    /// @return result True if the freezing executed correctly. Reverts on failure.
     function setFrozenTokens(address account, uint256 tokenId, uint256 amount) external returns (bool result);
 
     /// @notice Checks if a specific account is allowed to send tokens according to token rules.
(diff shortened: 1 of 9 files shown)


xaler5:
Open design question for the maintainers: should `ERC20Restricted` itself become directional, replacing `canTransact` with `canSend`/`canReceive`?

As merged in this PR, directionality lives only in `ERC20uRWA`, which has two consequences:

1. **Redundant checks by default**: a public transfer evaluates `canSend(from)`/`canReceive(to)` in `ERC20uRWA._update` and then `canTransact` on both sides again in `ERC20Restricted._update`.
2. **A loosening trap** (documented in the contract NOTE): overriding `canSend`/`canReceive` to be *more* permissive than `canTransact` doesn't work, because the symmetric check in `ERC20Restricted._update` still blocks the transfer.

A possible shape that avoids both without duplicating storage: keep the single `Restriction` enum and a single default predicate, but expose `canSend(account)`/`canReceive(account)` as the virtual surface in `ERC20Restricted` (both defaulting to `getRestriction(account) != BLOCKED`), and have its `_update` check `canSend(from)` (when not minting) and `canReceive(to)` (when not burning). The simple symmetric blocklist/allowlist case stays a single-override affair, asymmetry becomes opt-in, and `ERC20uRWA` shrinks to the frozen/forced logic plus the ERC-7943 error surface (probably via virtual check hooks, so `ERC20Restricted` stays EIP-agnostic).

It is a breaking change to a standalone extension, so I deliberately left it out of this PR. Happy to open a follow-up if there's interest.

contracts/interfaces/IERC7943.sol
ernestognw, line 1: I believe we should update IERC7943NonFungible and IERC7943MultiToken to remove `canTransact` and `ERC7943CannotTransact` as well. Might not be strictly on scope of this PR but seems reasonable to update all 7943 interfaces now it's final.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -83,31 +117,47 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
         address from,
         address to,
         uint256 amount
     ) internal virtual override(ERC20, ERC20Freezable, ERC20Restricted) {
-        // Note: We rely on the inherited _update chain (ERC20Freezable + ERC20Restricted) to enforce
-        // the same restrictions that canTransfer would check. This avoids duplicate validation
-        // while maintaining consistency between external queries and internal transfer logic.
+        if (!_isForcedTransfer()) {
+            if (from != address(0)) require(canSend(from), ERC7943CannotSend(from)); // Not minting
+            if (to != address(0)) require(canReceive(to), ERC7943CannotReceive(to)); // Not burning
```
ernestognw: Note both `canSend` and `canReceive` will call `canTransact`, and then, the `super._update` will run ERC20Restricted's `_checkRestriction` again.

This is not strictly a bug because it's enforcing all restrictions correctly, and each one can be overridden individually so the checks are justified. Though, if ERC7943 is not enforcing `canTransact` anymore I believe we can update ERC20Restricted to have `canSend` and `canReceive` instead. It's a breaking change but it's justifiable for community contracts

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -83,31 +117,47 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
-        Restriction restriction = getRestriction(from);
-        bool wasAccountAllowed = canTransact(from);
-        if (!wasAccountAllowed) _setRestriction(from, Restriction.ALLOWED);
-        _update(from, to, amount); // Explicit raw update to bypass all restrictions
-        if (!wasAccountAllowed) _setRestriction(from, restriction);
+        // Temporarily flag the transfer as forced rather than calling ERC20._update directly.
+        // This preserves any side effects from future overrides to _update while letting
+        // {_update} and {_checkRestriction} skip the sender-side and recipient-side checks.
+        ERC20_URWA_FORCED_TRANSFER_SLOT.asBoolean().tstore(true);
+        _update(from, to, amount);
```
ernestognw, line 130: Any subclass that overrides `_update` with an external call enables reentrancy while the `ERC20_URWA_FORCED_TRANSFER_SLOT` is set to true. We should warn about it in the NatSpec of the `forcedTransfer` function.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -70,9 +98,15 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
-        require(canTransact(to), ERC7943CannotTransact(to));
+        require(canReceive(to), ERC7943CannotReceive(to));
+        // A forced transfer to self moves no tokens, but the frozen balance adjustment below would still
+        // lower the frozen amount, effectively acting as an unauthorized unfreeze. Rejecting it preserves
+        // the separation between the enforcer and freezer roles.
+        require(from != to, ERC7943CannotTransfer(from, to, amount));
 
-        // Update frozen balance if needed. ERC-7943 requires that balance is unfrozen first and then send the tokens.
+        // Update frozen balance if needed. ERC-7943 requires that balance is unfrozen first (emitting
+        // the corresponding {IERC7943Fungible-Frozen} event via {ERC20Freezable-_setFrozen}) and then
```
ernestognw: AsciiDoc references don't work here

```suggestion
        // the corresponding Frozen event via _setFrozen) and then
```

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -70,9 +98,15 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
      */
     function forcedTransfer(address from, address to, uint256 amount) public virtual returns (bool result) {
         _checkEnforcer(from, to, amount);
-        require(canTransact(to), ERC7943CannotTransact(to));
+        require(canReceive(to), ERC7943CannotReceive(to));
+        // A forced transfer to self moves no tokens, but the frozen balance adjustment below would still
+        // lower the frozen amount, effectively acting as an unauthorized unfreeze. Rejecting it preserves
+        // the separation between the enforcer and freezer roles.
+        require(from != to, ERC7943CannotTransfer(from, to, amount));
```
ernestognw: Is this revert mandated by 7943 now? In my opinion it may be wiser to short circuit and return false or no-op.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -15,28 +15,56 @@ import {ERC20Restricted} from "./ERC20Restricted.sol";
...
+    bytes32 private constant ERC20_URWA_FORCED_TRANSFER_SLOT =
+        0x2f15ed3f796dcc08760ea3f28d5fdea0e800838283141b1cb98a3bdd04ef5400;
 
     /// @inheritdoc ERC165
     function supportsInterface(bytes4 interfaceId) public view virtual override(ERC165, IERC165) returns (bool) {
         return interfaceId == type(IERC7943Fungible).interfaceId || super.supportsInterface(interfaceId);
     }
 
+    /**
+     * @dev Returns whether `account` is allowed to send tokens. Defaults to {ERC20Restricted-canTransact}.
```
ernestognw:
```suggestion
     * @dev See {IERC7943Fungible-canSend}. Returns whether `account` is allowed to send tokens. Defaults to {ERC20Restricted-canTransact}.
```

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -15,28 +15,56 @@ import {ERC20Restricted} from "./ERC20Restricted.sol";
...
+     *
+     * Override to implement sender-specific restrictions distinct from {canReceive}.
+     */
+    function canSend(address account) public view virtual returns (bool) {
+        return canTransact(account);
+    }
+
+    /**
+     * @dev Returns whether `account` is allowed to receive tokens. Defaults to {ERC20Restricted-canTransact}.
```
ernestognw:
```suggestion
     * @dev See {IERC7943Fungible-canReceive}. Returns whether `account` is allowed to receive tokens. Defaults to {ERC20Restricted-canTransact}.
```

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -83,31 +117,47 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
-        // Assuming `forcedTransfer` will be used occasionally, the added costs of temporary
-        // restrictions would be justifiable under this path.
-        Restriction restriction = getRestriction(from);
-        bool wasAccountAllowed = canTransact(from);
-        if (!wasAccountAllowed) _setRestriction(from, Restriction.ALLOWED);
-        _update(from, to, amount); // Explicit raw update to bypass all restrictions
-        if (!wasAccountAllowed) _setRestriction(from, restriction);
+        // Temporarily flag the transfer as forced rather than calling ERC20._update directly.
+        // This preserves any side effects from future overrides to _update while letting
+        // {_update} and {_checkRestriction} skip the sender-side and recipient-side checks.
```
ernestognw:
```suggestion
        // _update and _checkRestriction skip the sender-side and recipient-side checks.
```

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -83,31 +117,47 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
+    /**
+     * @dev See {ERC20Restricted-_checkRestriction}. Skipped during a {forcedTransfer} so that
+     * enforcement transfers can move tokens out of restricted accounts.
+     */
+    function _checkRestriction(address account) internal view virtual override {
+        if (!_isForcedTransfer()) super._checkRestriction(account);
+    }
+
+    /// @dev Whether the current transfer is being performed by {forcedTransfer}.
+    function _isForcedTransfer() internal view returns (bool) {
```
ernestognw:
```suggestion
    function _isForcedTransfer() internal view virtual returns (bool) {
```

test/token/ERC20/extensions/ERC20uRWA.test.js
```diff
@@ -30,9 +30,14 @@ describe('ERC20uRWA', function () {
...
         'forcedTransfer(address,address,uint256)',
-        'canTransact(address)',
+        'canSend(address)',
+        'canReceive(address)',
       ],
     });
+
+    it('supports the final EIP-7943 fungible interface id', async function () {
+      await expect(this.token.supportsInterface('0x3edbb4c4')).to.eventually.equal(true);
+    });
```
ernestognw: I don't believe we need this test since we have the `shouldSupportInterfaces({ ERC7943Fungible: [...] })` up there

ernestognw:
> Open design question for the maintainers: should ERC20Restricted itself become directional, replacing canTransact with canSend/canReceive?

Just realized this is essentially answered by [this review comment](https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/254#discussion_r3625286082). I agree with splitting `canTransact` into `canSend`/`canReceive`. Breaking changes are not too strict for community contracts

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -15,28 +15,56 @@ import {ERC20Restricted} from "./ERC20Restricted.sol";
...
+     * here, as those validations belong to the base ERC-20 standard per EIP-7943.
+     *
+     * CAUTION: This function is only meant for external use. Overriding it will not apply the new
+     * checks to the internal {_update} function, which enforces {canSend}, {canReceive} and the
+     * unfrozen balance directly. Consider overriding {_update} accordingly to keep both in sync.
      */
     function canTransfer(address from, address to, uint256 amount) external view virtual returns (bool) {
-        return (amount <= available(from) && canTransact(from) && canTransact(to));
+        if (amount <= balanceOf(from) && amount > available(from)) return false;
+        return canSend(from) && canReceive(to);
```
ernestognw:
```suggestion
        return canSend(from) && canReceive(to) && (amount > balanceOf(from) || amount <= available(from));
```

xaler5:
Thanks for the thorough review Ernesto :saluting_face: ! Everything should be applied in `11d0905` and `2c94aab`:

- All the NatSpec and style suggestions are in: 
    - The `canSend`/`canReceive` docs
    - Plain text instead of AsciiDoc references in inline comments
    - Virtual `_isForcedTransfer`
    - The one-liner `canTransfer`
    - The redundant interface id test is gone
    
- `IERC7943NonFungible` and `IERC7943MultiToken` are now aligned with the final spec as well

- On the `from == to` revert, it is not mandated by the EIP, so I went with the no-op flavor. Returning `false` would clash with the final spec's `forcedTransfer` contract ("True if the transfer executed correctly. Reverts on failure"), so a self forced transfer now simply skips the frozen adjustment (which was the original bug, an implicit unfreeze) and behaves like a regular ERC-20 self transfer. It reverts with `ERC20InsufficientUnfrozenBalance` when the amount exceeds the unfrozen balance, exactly as a public self transfer would.

- Added the reentrancy CAUTION to `forcedTransfer` NatSpec

- And I took you up on the `ERC20Restricted` split: `canSend`/`canReceive` now live there, keeping the single `Restriction` enum, with `_checkSend`/`_checkReceive` hooks replacing `_checkRestriction`. `ERC20uRWA` no longer duplicates the eligibility checks in `_update`, maps the hooks to `ERC7943CannotSend`/`ERC7943CannotReceive`, and skips them during `forcedTransfer` via the transient flag.

All tests pass :)

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: Added

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: done :)

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -33,19 +33,20 @@ abstract contract ERC20Restricted is ERC20 {
...
+    /**
+     * @dev Returns whether a user account is allowed to receive tokens.
      *
-     * To convert into an allowlist, override as:
-     *
-     * ```solidity
-     * function canTransact(address account) public view virtual override returns (bool) {
-     *     return getRestriction(account) == Restriction.ALLOWED;
-     * }
-     * ```
```
ernestognw, line 46: I believe this example is valuable in both `canReceive` and `canSend`. I'd suggest keeping it

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -33,19 +33,20 @@ abstract contract ERC20Restricted is ERC20 {
     }
```
ernestognw, line 33: Now this function reads weird because the interface is directional but this function only returns a "single restriction". I'm feeling this function should go away in favor of `canSend` and `canReceive`

contracts/token/ERC20/extensions/ERC20Restricted.sol
```diff
@@ -33,19 +33,20 @@ abstract contract ERC20Restricted is ERC20 {
...
 
     /**
-     * @dev Returns whether a user account is allowed to interact with the token.
+     * @dev Returns whether a user account is allowed to send tokens.
      *
      * Default implementation only disallows explicitly BLOCKED accounts (i.e. a blocklist).
+     */
+    function canSend(address account) public view virtual returns (bool) {
+        return getRestriction(account) != Restriction.BLOCKED; // i.e. DEFAULT && ALLOWED
```
ernestognw: Hmm, so I'm realizing that ERC20Restricted interface is directional, but the storage is not.
In my opinion using a single `_restrictions` mapping for both `canSend` and `canReceive` seems fine because you don't want to force another slot's SLOAD on every transfer, but it feels weird that the interface allows 2 directions while the storage can only represent 1

ernestognw (commented):
This review is exclusive to ERC20Restricted; something feels off about its current state. On one hand, the interface supports directionality but, on the other, the storage is designed for a single restriction.

I'm thinking about the possibility of splitting `ERC20Restricted` into `ERC20RestrictedSend`  and `ERC20RestrictedReceive`, so that the original `ERC20Restricted` becomes a combination of both these contracts (there wouldn't be an `ERC20Restricted`, only the 2 composable versions. That means that by default, any combination of restricting both send and receive will use 2 storage slots (though, you can do `function canSend(...) ... override returns (bool) { return canReceive(...) }` or viceversa but I feel it doesn't compose well).

I'm still puzzled; I do want to support directionality at the ERC20Restricted level, but users looking for the equivalent functionality will be punished one way or another (either by the 2 slots, or by a weird `canSend`/`canReceive` override).

ernestognw:
> On the from == to revert, it is not mandated by the EIP, so I went with the no-op flavor. Returning false would clash with the final spec's forcedTransfer contract ("True if the transfer executed correctly. Reverts on failure")

Oh I see. It's weird that the ERC finalized with a return value that's always `true` for `forceSend`, isn't it?
In any case, the current no-op variant seems fine. Note that it keeps emitting the `ForcedTransfer` event, which imo seems correct

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -47,67 +72,99 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
+        // the corresponding Frozen event via _setFrozen) and then send the tokens. Skipped for
+        // self-transfers, where the balance does not change and no unfreeze is warranted.
         uint256 currentFrozen = frozen(from);
         uint256 newBalance;
         unchecked {
             // Safe because ERC20._update will check that balanceOf(from) >= amount
             newBalance = balanceOf(from) - amount;
         }
-        if (currentFrozen > newBalance) {
+        if (from != to && currentFrozen > newBalance) {
```
ernestognw: Maybe the `from != to` should be place differently. See that currently, `currentFrozen` is gotten from an SLOAD and `newBalance` is calculated with a `balanceOf` regardless of `from != to`. Another instance is the `_update` function, which still runs (arguably correct, Transfer event emits regardless).

Bottom line is that `frozen(...)` and `balanceOf(...)` are legitimately removable; `_update` is fine but wanted to flag it.

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -47,67 +72,99 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
         super._update(from, to, amount);
     }
 
+    /**
+     * @dev See {ERC20Restricted-_checkSend}. Reverts with {IERC7943Fungible-ERC7943CannotSend}.
+     * Skipped during a {forcedTransfer} so that enforcement transfers can move tokens out of
+     * restricted accounts.
+     */
+    function _checkSend(address account) internal view virtual override {
+        if (!_isForcedTransfer()) require(canSend(account), ERC7943CannotSend(account));
```
ernestognw: According to ERC-7943, the `ERC7943CannotSend` was introduced for completeness but it's not mandated by the spec. I believe it's better if we:

```suggestion
        if (!_isForcedTransfer()) super._checkReceive(account);
```

contracts/token/ERC20/extensions/ERC20uRWA.sol
```diff
@@ -47,67 +72,99 @@ abstract contract ERC20uRWA is ERC20, ERC165, ERC20Freezable, ERC20Restricted, I
...
+    }
+
+    /**
+     * @dev See {ERC20Restricted-_checkReceive}. Reverts with {IERC7943Fungible-ERC7943CannotReceive}.
+     * Skipped during a {forcedTransfer}, where the recipient has already been checked upfront.
+     */
+    function _checkReceive(address account) internal view virtual override {
+        if (!_isForcedTransfer()) require(canReceive(account), ERC7943CannotReceive(account));
```
ernestognw: Similar to above:


```suggestion
        if (!_isForcedTransfer()) super._checkReceive(account);
```

ernestognw (commented):
Almost there! After reviewing, I realized the ERC20Restricted refactor might be more complicated than I thought. Not because of backwards compatibility but because a more fundamental decision about what to ship. As of now, it seems to me that any user of ERC20Restricted picked it assuming non-directionality; adding it on top, although achieving the same functionality, feels like the contract drifted away completely from its original purpose without a clean workaround.

Sorry for asking this, but because I don't have an answer on the new shape of ERC20Restricted, I believe we should revert back to when ERC20Restricted didn't add directionality. It's the least controversial position imo, then we can file an issue to explore directionality in ERC20Restricted.

xaler5:
Thanks Ernesto, all good points. This round is applied:

- **ERC20Restricted reverted** in 740b81b: it is back to the exact non-directional shape from master (`canTransact`, single `Restriction`, allowlist example included). Directionality stays layered in `ERC20uRWA`, where `canSend`/`canReceive` default to `canTransact` and are enforced in `_update`. I agree it is the least controversial position while the ERC20Restricted shape is undecided; I filed #257 to explore that properly (including your composable Send/Receive idea and a packed-storage variant that keeps a single SLOAD while making storage match the directional interface).
- **`from != to` placement** in 33d16b3: `frozen(...)` and `balanceOf(...)` are now only read when `from != to`, and `_update` still runs so the Transfer event is preserved, as you noted.
- **Error surface** in af0392b: followed your suggestion, adapted to the reverted shape (the `_checkSend`/`_checkReceive` hooks no longer exist). Regular transfer, mint, and burn eligibility failures revert with `ERC20UserRestricted`; `forcedTransfer` keeps `ERC7943CannotReceive` on its explicit recipient pre-check since that one is the 7943-facing path.
- Your comments on `getRestriction` and the storage/interface mismatch are resolved by the revert; they become input for #257.

Full suite green as before. Ready for another look whenever you have time!

contracts/token/ERC20/extensions/ERC20Restricted.sol
xaler5: Kept! The revert restored ERC20Restricted to the exact version from master, allowlist example included.

contracts/token/ERC20/extensions/ERC20Restricted.sol
xaler5: Resolved by the revert: ERC20Restricted is back to the symmetric canTransact shape, so getRestriction matches the interface again. Whether it survives a directional redesign is now part of #257.

contracts/token/ERC20/extensions/ERC20Restricted.sol
xaler5: Agreed, that mismatch was the core of the problem. Reverted here and moved the design discussion to #257, including a packed single-slot option that keeps one SLOAD per transfer while making the storage genuinely directional.

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: Applied in 33d16b3: the from != to guard now wraps the whole adjustment block, so frozen(...) and balanceOf(...) are only read when from != to. _update still runs, keeping the Transfer event as you noted.

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: Followed, adapted to the reverted shape (the _checkSend/_checkReceive hooks are gone with the revert): the eligibility checks in ERC20uRWA._update now revert with ERC20UserRestricted (af0392b).

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: Same as above: ERC20UserRestricted on the regular transfer path. forcedTransfer keeps ERC7943CannotReceive on its explicit recipient pre-check, since that one is the 7943-facing entry point.

contracts/token/ERC20/extensions/ERC20uRWA.sol
xaler5: Agreed on both counts: the always-true bool is indeed a bit odd in the final spec (it stayed for ERC-20 style composability, with reverts carrying the failure information), and yes, the no-op variant intentionally keeps emitting ForcedTransfer plus the base Transfer event.

xaler5:
> This review is exclusive to ERC20Restricted; something feels off about its current state. On one hand, the interface supports directionality but, on the other, the storage is designed for a single restriction.

Totally fair, and I share the feeling. That interface/storage tension (and your composable Send/Receive idea, plus a packed-storage variant that avoids the second SLOAD) is now captured in #257 so we can settle the ERC20Restricted shape without blocking this PR.

xaler5:
> Sorry for asking this, but because I don't have an answer on the new shape of ERC20Restricted, I believe we should revert back to when ERC20Restricted didn't add directionality. It's the least controversial position imo, then we can file an issue to explore directionality in ERC20Restricted.

No need to apologize, it was the right call. Done in 740b81b: ERC20Restricted is byte for byte the version from master again, directionality lives only in ERC20uRWA, and #257 tracks the exploration. All your inline comments from this round are addressed (see the thread replies).

ernestognw:
Amazing, thanks for the updates @xaler5. I'm merging even though #257 might require more discussion/exploration.

ernestognw (approved):

PR #262, opened 2026-07-31
Enable setting template via inheritance

arr00:


contracts/account/RoleAccountFactory.sol
```diff
@@ -58,4 +65,9 @@ contract RoleAccountFactory {
 
         return roleAccount;
     }
+
+    /// @dev Called once during construction to get the template used for deploying role accounts.
+    function _deployTemplate() internal virtual returns (address) {
+        return address(new RoleAccount());
+    }
```
gonzaotc, line 72:
```suggestion
    /// @dev Called once during construction to get the template used for deploying role accounts.
    function template() public virtual returns (address) {
        return address(new RoleAccount());
    }
```

So that the name doesn't imply the template is being deployed at construction (I can override with the address of my custom account that is already deployed)

gonzaotc (approved):

contracts/account/RoleAccountFactory.sol
arr00: The reason I don't want it to be simply `template` is that leads developers to believe this is consulted on every deploy.

PR #266, opened 2026-08-10
Add ERC20UniversalWrapper, a clone-based universal ERC-20 wrapper

Amxx:
Following [this twitter discussion](https://x.com/Walodja1987/status/2086429936055669119), I wanted to build it using OZ contracts.

Adds `ERC20UniversalWrapper` and `ERC20UniversalWrapperFactory` under `contracts/token/ERC20/examples/`.

## What it does

`ERC20UniversalWrapper` wraps any ERC-20 token, exposing ERC-20 permit, ERC-1363, ERC-3009 and ERC-4626 on top of it.

It is deployed as a **clone with immutable args**: the underlying asset and the feature switches live in the clone's bytecode rather than in storage. The consequences are that it needs no constructor and no initializer, deploying a wrapper costs no storage writes, and each `(asset, settings)` pair maps to a single deterministic address. `name`, `symbol` and `decimals` are derived from the underlying token's metadata, so they aren't stored either (`decimals` degrades to 18 when the asset doesn't expose it).

Two behaviors are configurable per clone, through a settings byte:

| Flag | Effect |
| --- | --- |
| `_VARIABLE_RATE_FLAG` | The wrapper acts as a regular ERC-4626 vault. When unset, the rate is fixed 1:1, as `ERC20Wrapper` would. |
| `_FLASH_MINT_FLAG` | The wrapper token can be flash-minted. When unset, `maxFlashLoan` returns 0. |

The two are mutually exclusive: flash minting shares would interfere with the vault's exchange rate, so requesting both reverts with `IncompatibleSettings`.

`ERC20UniversalWrapperFactory` deploys the shared implementation in its constructor and mints the clones. Deployment is permissionless and idempotent — anyone can deploy the wrapper for a given configuration, but only once — and `predict` computes a wrapper's address ahead of (or without) deployment.

## Tests

42 tests in `test/token/ERC20/examples/ERC20UniversalWrapperFactory.test.js`, table-driven over the three valid settings combinations, covering the implementation, `predict`, `deploy` (address, event, permissionlessness, redeploy revert, coexisting configurations), and the resulting wrapper's asset, metadata, flash-mint switch and exchange rate.

## Notes for reviewers

- `_parseArgs` requires the immutable args to be **exactly** 21 bytes. `Clones.fetchCloneArgs` sizes its result from `code.length - 0x2d`, so a `>=` check would silently pass on the implementation contract itself and decode its own bytecode as an asset address. The strict check makes `MissingImmutableArgs` reachable and the implementation inert; it also means a hand-rolled clone with padded args is rejected rather than tolerated.
- Three imports carry `// prettier-ignore` to keep them on one line; they exceed the configured `printWidth` of 120.

## Follow-ups, deliberately not in this PR

- [ ] `CHANGELOG.md` entry.
- [ ] Listing in `contracts/token/README.adoc` (both the bullet list and the `{{...}}` docs section).

## Summary by CodeRabbit

* **New Features**
  * Added universal ERC-20 wrappers that can represent underlying tokens with fixed or variable conversion rates.
  * Added optional flash minting and support for permit, flash loan, ERC-1363, ERC-3009, and ERC-4626 integrations.
  * Added a factory for deploying wrappers and predicting their deterministic addresses.
  * Added dynamic token metadata and deployment events.
* **Bug Fixes**
  * Invalid configurations and malformed setup data now fail safely.
  * Flash minting is unavailable unless explicitly enabled.

### DIFF
--- a/contracts/token/ERC20/examples/ERC20UniversalWrapperFactory.sol
+++ b/contracts/token/ERC20/examples/ERC20UniversalWrapperFactory.sol
@@ -8,13 +8,19 @@ import {Clones} from "@openzeppelin/contracts/proxy/Clones.sol";
 import {SafeERC20} from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import {Math} from "@openzeppelin/contracts/utils/math/Math.sol";
 import {ERC20Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/ERC20Upgradeable.sol";
-// prettier-ignore
-import {ERC20PermitUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20PermitUpgradeable.sol";
-// prettier-ignore
-import {ERC20FlashMintUpgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20FlashMintUpgradeable.sol";
+import {
+    ERC20PermitUpgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20PermitUpgradeable.sol";
+import {
+    ERC20FlashMintUpgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC20FlashMintUpgradeable.sol";
+import {
+    ERC20TemporaryApprovalUpgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/draft-ERC20TemporaryApprovalUpgradeable.sol";
 import {ERC1363Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC1363Upgradeable.sol";
-// prettier-ignore
-import {ERC3009Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/draft-ERC3009Upgradeable.sol";
+import {
+    ERC3009Upgradeable
+} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/draft-ERC3009Upgradeable.sol";
 import {ERC4626Upgradeable} from "@openzeppelin/contracts-upgradeable/token/ERC20/extensions/ERC4626Upgradeable.sol";
 
 /// @dev Settings flag enabling a variable (vault-like) exchange rate between the underlying asset and the wrapper.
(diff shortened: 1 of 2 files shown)


Walodja1987:
Thanks @Amxx for putting this together!

Maybe some additional context on how I arrived at the wrapper idea and what I was originally trying to solve:

The idea originated from looking for a pragmatic way to improve confidentiality when transacting with ERC-20 tokens on Ethereum. One approach is to receive funds into fresh addresses so that counterparties don't immediately see the balance and transaction history of your main account.

The problem with standard ERC-20 tokens such as USDT or USDC is that once tokens are received into a fresh address, that address needs ETH before it can move the tokens again.

That's where the idea of `transferWithAuthorization` came from. It is conceptually similar to ERC-3009, but I added two fields:

- `executor`: specifies who is allowed to execute the authorization. `address(0)` means anyone can execute it.
- `executorFee`: an amount denominated in the token itself that is paid to the executor as compensation for submitting the transaction and paying the gas in ETH.

This allows an address holding, for example, XUSDC but no ETH to sign an authorization offchain and have a relayer execute the transfer:

`XUSDC holder → signs authorization → relayer pays ETH → recipient receives XUSDC + relayer receives XUSDC fee`

One way to introduce functionality like this without requiring the issuer of the underlying asset to change its token is simply to wrap it. For example, USDT could be wrapped into XUSDT, which remains ERC-20 compatible but exposes additional functionality such as authorization-based transfers.

Once we have such a wrapper, the broader question becomes: **which improvements to ERC-20 should the wrapper include?**

There are several potentially useful additions that have been proposed or explored over the years: permit, authorization-based transfers, expiring approvals, callbacks, batching, etc. Your draft takes this quite far and makes the wrapper broadly useful, including ERC-4626 and optional flash minting.

For my specific use case, I'm currently experimenting with a deliberately narrower, payment-focused version:

- standard ERC-20 + permit
- expiring approvals via `approveUntil`
- `revokeApproval` as a convenience function
- `transferWithAuthorization` with optional `executor` and `executorFee`
- batch authorization transfers
- authorization-based withdrawals, allowing XERC20 to be redeemed back into the underlying token through a relayer without the holder needing ETH

Draft: gist [dot] github [dot] com/Walodja1987/e8ecad9ce8ebaa5ad6cc540c3da2c7f6

I took your clone-with-immutable-args architecture and adapted my XERC20 draft to it. I really like the property that the wrapper address can be deterministically derived from the underlying asset without requiring a registry.

For now, I also removed the settings byte entirely, so there is exactly one deterministic XERC20 wrapper per underlying asset for a given factory/implementation.

I think the main question is how much functionality should a universal wrapper reasonably expose without becoming too complex.

Would be very interested in your thoughts.

PR #283, opened 2026-09-16
Hardhat 3 migration

Amxx:
Migrates this repository to Hardhat 3, following the transition already done in the main repo: OpenZeppelin/openzeppelin-contracts#6542 (`Migrate to Hardhat3`), plus its follow-ups. The goal is to stay aligned with `@openzeppelin/contracts`, whose Hardhat plugins and test conventions this repo reuses directly through the `lib/@openzeppelin-contracts` submodule.

## Tooling and configuration

- `hardhat.config.js` → **`hardhat.config.ts`**, rewritten around `defineConfig` with an explicit `plugins` list instead of implicit `require` side effects.
- The package is now **`"type": "module"`**. Consequently `solhint.config.js` → `solhint.config.cjs`, and the config is passed explicitly via `--config`.
- Plugins are imported straight from the contracts submodule rather than duplicated here: `hardhat-exposed`, `hardhat-transpiler`, `hardhat-solidity-docgen`, `hardhat-oz-contracts-helpers`, `async-test-sanity`.
- `hardhat/remappings.js` and `.solcover.js` are removed — remappings are handled by `remappings.txt` / `foundry.toml`, and coverage is now built into Hardhat.
- `npm run coverage` → `hardhat test --coverage`; `npm run gas-report` → `hardhat test --gas-stats`. Both previously went through custom scripts and `hardhat-gas-reporter`.
- Solidity defaults move to `0.8.35` / `evmVersion: osaka`, overridable through the same `yargs` env options as before.
- Solidity (Foundry-style) tests now run under Hardhat's test runner, with `fuzz.runs`, `maxTestRejects` and `fsPermissions` configured in `test.solidity`.

## Artifact resolution

Hardhat 3 only emits artifacts for compilation roots, so two workarounds from the Hardhat 2 setup had to be replaced:

- `contracts/mocks/import.sol` (importing dependencies purely to force artifact emission) is replaced by an explicit **`solidity.npmFilesToBuild`** list.
- `ECDSAOwnedDKIMRegistry` is needed as-is by the tests but is reached through a `remappings.txt` entry, so no compilation root imports it by name. Its directory is added to **`paths.sources`**, which is what makes the artifact appear.

Both are commented in `hardhat.config.ts` so the reasoning isn't lost.

## Tests

- All JS tests and helpers move from **CommonJS to ESM**.
- The global `hardhat` `ethers` object is gone; each test file creates its own connection with `await network.create()` and destructures `ethers`, `networkHelpers` and `helpers` from it. This is the single largest source of churn in the diff, and it is mechanical.
- Shared helpers (`chains`, `signers`, `enums`, `eip712-types`, the crosschain helpers) are updated to take or expose a connection rather than reaching for the global.

## Submodule bumps and the source changes they require

`lib/@openzeppelin-contracts` and `lib/@openzeppelin-contracts-upgradeable` are bumped, which pulls in a few upstream changes that this repo has to follow:

- **`draft-` prefix removed from the ERC-7786 interface** (openzeppelin-contracts#6745) — every `interfaces/draft-IERC7786.sol` import is updated to `interfaces/IERC7786.sol`.
- **Import ordering is now linted** upstream (openzeppelin-contracts#6746); imports across `contracts/` are reordered to match.
- **`EnumerableSet.removeAt`** (openzeppelin-contracts#6691) — `EnumerableSetExtended` and `EnumerableMapExtended` gain the matching `removeAt` overloads (and `Bytes32x2Set` gains the paginated `values(start, end)`), so the extended types keep parity with the base library. Codegen templates and `scripts/generate/format-lines.js` are updated accordingly.

## Unrelated changes riding along

Flagging these explicitly since they are not part of the migration:

- **`WormholeGatewayAdapter`: the `wormholeChainId` constructor argument is removed.** The immutable it fed was never read. This is an **ABI-breaking constructor change** for anyone deploying this adapter.
- `ERC7786OpenBridge` return-buffer hardening was committed here before noticing it had already landed on `master` as #278. The merge of `master` kept both copies of the accompanying test; the duplicate has been removed, keeping the copy that pins an explicit `gasLimit` (without it, gas estimation settles on the path where the recipient runs out of gas allocating its return buffer, and the test fails). The contract change in this branch is now redundant with `master`.
- The js lint scripts regained their `lib/` exclusions, which had been dropped during the dependency update. `eslint` lost `--ignore-pattern lib/`, and routing prettier through a shared script that passes `--ignore-path .gitignore` silently disabled `.prettierignore` (in prettier 3, `--ignore-path` replaces the default rather than adding to it). Both linters were walking the submodules under `lib/`, leaving `npm run lint` permanently red.

## Known gaps

- **Docgen is disabled**: the `docgen` key in `hardhat.config.ts` is commented out, so `npm run docs` / the `docs.yml` workflow are not exercised by this PR and need a follow-up.
- `scripts/checks/coverage.sh` is now orphaned — `npm run coverage` no longer calls it. Left in place rather than deleted as part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added indexed `removeAt` operations for enumerable maps and sets, returning removed entries.
  * Expanded internal access to selected token operations through public visibility.

* **Compatibility**
  * Updated ERC-7786 integrations to use finalized interface definitions.

* **Tooling**
  * Migrated project configuration and test workflows to the latest Hardhat setup and ES modules.
  * Improved formatting, linting, coverage, and gas-reporting commands.

* **Maintenance**
  * Updated OpenZeppelin components and refreshed generated collection utilities.


test/utils/structs/EnumerableSetExtended.test.js
```diff
@@ -17,12 +23,13 @@ async function fixture() {
...
         values: Array.from(
           { length: 3 },
-          value.size ? () => Array.from({ length: value.size }, generators[value.base]) : generators[value.type],
+          value.size ? () => Array.from({ length: value.size }, randomOf(value.base)) : randomOf(value.type),
         ),
         methods: mapValues(
           {
             add: `$add(uint256,${value.type})`,
             remove: `$remove(uint256,${value.type})`,
+            removeAt: `$removeAt(uint256,uint256)`,
```
Amxx:
```suggestion
            removeAt: `$removeAt(uint256,uint256)`, // `$removeAt_EnumerableSetExtended_${name}(uint256,uint256)` when more types are added
```

remappings.txt
```diff
@@ -1,7 +1,8 @@
-@openzeppelin/contracts/=lib/@openzeppelin-contracts/contracts/
+@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/contracts/
```
Amxx, line 1: This change is necessary for the backward lookup to work. I don't think its controversial. If we could avoid it it may be preferable, but not at the cost of greatly increassing the complexity of the exposed plugin.

PR #286, opened 2026-09-21
Fix inheritance ordering

james-toussaint:
Target: https://github.com/OpenZeppelin/openzeppelin-community-contracts/pull/283.


scripts/checks/inheritance-ordering.js
Amxx, line 1: this file was updated in vanilla. It uses the "files" from `package.json` to filter which contract are in scope. 

Would it work to "just" use that script (from vanilla here) ? Ideally we would not want to maintain 2 different versions

james-toussaint: imo it would be great to parametrize vanilla script and symlink it here.
Also this PR fixes the order issue: https://github.com/OpenZeppelin/openzeppelin-community-contracts/issues/287

Amxx: done

Amxx (approved):

scripts/checks/inheritance-ordering.js
james-toussaint: Nice!

